PR #33, opened 2013-04-19
Stop using deprecated API tokens for Enterprise

watsonian:
Authentication by API token is deprecated and doesn't work anymore. Instead, authentication should be done via OAuth token now. This is causing the GitHub plugin to fail when trying to setup Jenkins to auto-manage web hooks and will just be breaking in general for anyone trying to use this with Enterprise.

I'm not much of a Java guy, so I'd appreciate a look over this to make sure it works.

### DIFF
--- a/src/main/java/org/kohsuke/github/GitHub.java
+++ b/src/main/java/org/kohsuke/github/GitHub.java
@@ -61,17 +61,16 @@ public class GitHub {
 
 
     /*package*/ final String encodedAuthorization;
-    /*package*/ final String apiToken;
 
     private final Map<String,GHUser> users = new HashMap<String, GHUser>();
     private final Map<String,GHOrganization> orgs = new HashMap<String, GHOrganization>();
-	/*package*/ String oauthAccessToken;
+    /*package*/ String oauthAccessToken;
 
-	private final String apiUrl;
+    private final String apiUrl;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -132,24 +130,24 @@ public static GitHub connect() throws IOException {
      *      "http://ghe.acme.com/api/v3". Note that GitHub Enterprise has <tt>/api/v3</tt> in the URL.
      *      For historical reasons, this parameter still accepts the bare domain name, but that's considered deprecated.
      */
-    public static GitHub connectToEnterprise(String apiUrl, String login, String apiToken) {
-        return new GitHub(apiUrl,login,apiToken,null);
+    public static GitHub connectToEnterprise(String apiUrl, String oauthAccessToken) throws IOException {
```
iraleigh: I know that this is late, because this has already been merged. There is a problem with removing the login parameter makes this connection fail. From my understanding the login and OAuth are associated with each other, an OAuth cannot be used to login by itself.

I see the call to getMyself(), but I am confused about the source of truth for "myself". If you have not explicitly set "myself" with the login, how can you call getMyself() for the login. 

@watsonian @kohsuke

PR #155, opened 2015-02-18
Implementing github trees

ddtxra:
Implementing GitHub Trees as described in https://developer.github.com/v3/git/trees/
Supports getTree and getTreeRecursive method on the GHRepository class


suryagaddipati:
:+1:

PR #156, opened 2015-02-22
Picking endpoint from the properties file and environment variables

ashwanthkumar:
Helps seemless switching between public github and enterprise without any code changes.

### DIFF
--- a/src/main/java/org/kohsuke/github/GitHubBuilder.java
+++ b/src/main/java/org/kohsuke/github/GitHubBuilder.java
@@ -69,6 +69,10 @@ public class GitHubBuilder {
 		}
     
     }
+
+    public static GitHubBuilder fromEnvironment(String loginVariableName, String passwordVariableName, String oauthVariableName) throws IOException {
+        return fromEnvironment(loginVariableName, passwordVariableName, oauthVariableName, "");
+    }
     
     public static GitHubBuilder fromEnvironment(String loginVariableName, String passwordVariableName, String oauthVariableName, String endpointVariableName) throws IOException {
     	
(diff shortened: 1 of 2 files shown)


src/main/java/org/kohsuke/github/GitHubBuilder.java
```diff
@@ -70,7 +70,7 @@ public static GitHubBuilder fromCredentials() throws IOException {
     
     }
     
-    public static GitHubBuilder fromEnvironment(String loginVariableName, String passwordVariableName, String oauthVariableName) throws IOException {
+    public static GitHubBuilder fromEnvironment(String loginVariableName, String passwordVariableName, String oauthVariableName, String endpointVariableName) throws IOException {
```
KostyaSha: Will broke backward compatibility

src/main/java/org/kohsuke/github/GitHubBuilder.java
```diff
@@ -89,7 +89,11 @@ public static GitHubBuilder fromEnvironment(String loginVariableName, String pas
     	
     	if (oauthValue != null)
     		env.put("oauth", oauthValue);
-    	
+
+        Object endPoint = System.getenv(endpointVariableName);
```
KostyaSha: Wrong indentation, and why Object?

ashwanthkumar: Sorry about the wrong indentation, will fix it. All `System.getenv` was taken in as Object, that's why. Do I change all to String?

src/main/java/org/kohsuke/github/GitHubBuilder.java
ashwanthkumar: Ah, Will overload it.

PR #161, opened 2015-03-05
Add method to get the list of languages using in repository

khoa-nd:
Hi kohsuke
I used your API to call Github API but I found that there isn't any method lists all languages using in repository. Therefore, I ask you to add it into your master code.
Thanks

### DIFF
--- a/src/main/java/org/kohsuke/github/GHRepository.java
+++ b/src/main/java/org/kohsuke/github/GHRepository.java
@@ -285,6 +285,18 @@ public class GHRepository extends GHObject {
         };
     }
 
+    /**
+     * List languages for the specified repository.
+     * The value on the right of a language is the number of bytes of code written in that language.
+     * {
+         "C": 78769,
+         "Python": 7769
+       }
+     */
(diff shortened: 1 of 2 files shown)


src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -285,6 +285,18 @@ protected void wrapUp(GHTag[] page) {
...
 
+    /**
+     * List languages for the specified repository.
+     * The value on the right of a language is the number of bytes of code written in that language.
+     * {
+         "C": 78769,
+         "Python": 7769
+       }
+     */
+    public Map<String,Integer> listLanguages() throws IOException {
```
KostyaSha: Size in bytes, max Integer will be 2Gb, i think some crazy repo may have such number. Long looks better.

khoa-nd:
Hi KostyaSha , 
Thanks for your feedback. I already changed the code as you mentioned

KostyaSha:
Thanks, looks good for me, let's wait @kohsuke

oleg-nenashev:
Merged, thanks!

Issue #171, opened 2015-03-23
@WithBridgeMethods decorator in GHObject has no value adapterMethod


madhephaestus:
When pulling the sources and building there is a build error in GHObject at each usage of adapterMethod. This is not a runtime problem, it will not even compile. 

Perhaps this is a problem with the decorator definition and the dependency version number problem?

build 1.6.4-SNAPSHOT 
bridge-method-annotation-1.8.jar

madhephaestus:
Deleting the overloading decorator does not seem to break this for now.

KostyaSha:
Logs?

madhephaestus:
re-running the mvn with the stack trace on. I worked around it by going into the source and removing the decorator. No jar artifact was produced. I am working on a projecct in eclipse and copied the sources in to get it working. Ill replace it once the 1.64 release is posted.

madhephaestus:
$ mvn deploy -e
+ Error stacktraces are turned on.
[INFO] Scanning for projects...
[INFO] ------------------------------------------------------------------------
[INFO] Building GitHub API for Java
[INFO]    task-segment: [deploy]
[INFO] ------------------------------------------------------------------------
[INFO] [resources:resources {execution: default-resources}]
[INFO] Using 'UTF-8' encoding to copy filtered resources.
[INFO] skip non existing resourceDirectory /home/hephaestus/git/github-api/src/main/resources
[INFO] [compiler:compile {execution: default-compile}]
[INFO] Nothing to compile - all classes are up to date
[INFO] [bridge-method-injector:process {execution: default}]
[INFO] [resources:testResources {execution: default-testResources}]
[INFO] Using 'UTF-8' encoding to copy filtered resources.
[INFO] skip non existing resourceDirectory /home/hephaestus/git/github-api/src/test/resources
[INFO] [compiler:testCompile {execution: default-testCompile}]
[INFO] Nothing to compile - all classes are up to date
[INFO] [groovy:testCompile {execution: default}]
[INFO] No sources found to compile
[INFO] [surefire:test {execution: default-test}]
[INFO] Surefire report directory: /home/hephaestus/git/github-api/target/surefire-reports

-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running org.kohsuke.github.CommitTest
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.44 sec <<< FAILURE!
Running org.kohsuke.github.LifecycleTest
[...]

PR #173, opened 2015-03-23
Removing mis-match of decorator and method defintions

madhephaestus:
This addresses the issue #171


kohsuke:
I can't take this change as this breaks backward compatibility with earlier versions. We need to clarify what #171 really means.

kohsuke:
Closing this until #171 is determined to be a problem in this library.

PR #185, opened 2015-04-30
Fixes #183: added a method listForks() to GHRepository

### LINKED ISSUE #183: no way to list forks of a repository
Seems there is no way to list all the forks of a GHRepository. I only find methods to count them(?), and to check if it is a fork, and look for parent/source, and to create a fork.

Github API supports to [list forks](https://developer.github.com/v3/repos/forks/)

marc-guenther:
listForks() will list all forks of a repository.

An optional sort argument is also supported.

I did not know if list* or get* is the way to go, but as getForks() is already taken (it returns the number of forks), I decided to use listForks(), with an optional sort argument.

### DIFF
--- a/src/main/java/org/kohsuke/github/GHRepository.java
+++ b/src/main/java/org/kohsuke/github/GHRepository.java
@@ -321,6 +321,10 @@ public class GHRepository extends GHObject {
         return fork;
     }
 
+    /**
+     * Returns the number of all forks of this repository.
+     * This not only counts direct forks, but also forks of forks, and so on.
+     */
     public int getForks() {
         return forks;
     }
@@ -511,14 +515,15 @@ public class GHRepository extends GHObject {
     public static enum Sort { NEWEST, OLDEST, STARGAZERS }
 
     /**
-     * Lists all the forks of this repository.
+     * Lists all the direct forks of this repository, sorted by {@link Sort#NEWEST Sort.NEWEST}
      */
     public PagedIterable<GHRepository> listForks() {
       return listForks(null);
     }
 
     /**
-     * Lists up all the forks of this repository, sorted by the given sort order.
+     * Lists all the direct forks of this repository, sorted by the given sort order.
+     * @param sort the sort order. If null, defaults to {@link Sort#NEWEST Sort.NEWEST}.
      */
     public PagedIterable<GHRepository> listForks(final Sort sort) {
         return new PagedIterable<GHRepository>() {
--- a/src/main/java/org/kohsuke/github/GHRepository.java
+++ b/src/main/java/org/kohsuke/github/GHRepository.java
@@ -506,6 +506,35 @@ public class GHRepository extends GHObject {
     }
 
     /**
+     * Sort orders for listing forks
+     */
+    public static enum Sort { NEWEST, OLDEST, STARGAZERS }
+
+    /**
+     * Lists all the forks of this repository.
+     */
+    public PagedIterable<GHRepository> listForks() {
+      return listForks(null);
+    }
+
+    /**
+     * Lists up all the forks of this repository, sorted by the given sort order.
+     */
+    public PagedIterable<GHRepository> listForks(final Sort sort) {
+        return new PagedIterable<GHRepository>() {
+            public PagedIterator<GHRepository> iterator() {
+                return new PagedIterator<GHRepository>(root.retrieve().asIterator(getApiTailUrl("forks" + ((sort == null)?"":("?sort="+sort.toString().toLowerCase(Locale.ENGLISH)))), GHRepository[].class)) {
+                    @Override
+                    protected void wrapUp(GHRepository[] page) {
+                        for (GHRepository c : page)
+                            c.wrap(root);
+                    }
+                };
+            }
+        };
+    }
+
+    /**
      * Forks this repository as your repository.
      *
      * @return
--- a/src/main/java/org/kohsuke/github/GHRepository.java
+++ b/src/main/java/org/kohsuke/github/GHRepository.java
@@ -37,6 +37,7 @@ import java.net.URL;
 import java.util.*;
 
 import static java.util.Arrays.asList;
+import static org.apache.commons.lang.ObjectUtils.defaultIfNull;
 
 /**
  * A repository on GitHub.
@@ -512,10 +513,11 @@ public class GHRepository extends GHObject {
     /**
      * Sort orders for listing forks
      */
-    public static enum Sort { NEWEST, OLDEST, STARGAZERS }
+    public static enum ForkSort { NEWEST, OLDEST, STARGAZERS }
 
     /**
-     * Lists all the direct forks of this repository, sorted by {@link Sort#NEWEST Sort.NEWEST}
+     * Lists all the direct forks of this repository, sorted by
+     * github api default, currently {@link ForkSort#NEWEST ForkSort.NEWEST}.
      */
     public PagedIterable<GHRepository> listForks() {
       return listForks(null);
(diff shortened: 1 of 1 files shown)


src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -506,6 +506,35 @@ public void delete() throws IOException {
...
+     * Lists all the forks of this repository.
+     */
+    public PagedIterable<GHRepository> listForks() {
+      return listForks(null);
+    }
+
+    /**
+     * Lists up all the forks of this repository, sorted by the given sort order.
+     */
+    public PagedIterable<GHRepository> listForks(final Sort sort) {
```
oleg-nenashev: Add <code>@CheckForNull</code> and describe the behavior in Javadoc

marc-guenther: There is not a single `@CheckForNull` in the entire codebase so far. If you want to start introducing them, go ahead, but that was not the scope of this pull request.
Regarding Javadoc, I was under the impression that both methods I added actually have Javadoc comments, but maybe I miscounted.

oleg-nenashev: My intention was to highlight the method accepts nulls and to describe the behavior in that case.

> There is not a single @CheckForNull in the entire codebase so far. If you want to start introducing them, go ahead, but that was not the scope of this pull request.

I think that new pull requests could follow better practices than the legacy code

marc-guenther: Agreed.

Just added (hopefully) better Javadoc. Unfortunately, I was not able to find out how to add the `@CheckForNull` annotation.

marc-guenther:
Any chance of merging this? We are using this all the time now, and not a single problem so far...

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -506,6 +510,36 @@ public void delete() throws IOException {
...
+    }
+
+    /**
+     * Lists all the direct forks of this repository, sorted by the given sort order.
+     * @param sort the sort order. If null, defaults to {@link Sort#NEWEST Sort.NEWEST}.
+     */
+    public PagedIterable<GHRepository> listForks(final Sort sort) {
+        return new PagedIterable<GHRepository>() {
+            public PagedIterator<GHRepository> iterator() {
+                return new PagedIterator<GHRepository>(root.retrieve().asIterator(getApiTailUrl("forks" + ((sort == null)?"":("?sort="+sort.toString().toLowerCase(Locale.ENGLISH)))), GHRepository[].class)) {
```
lanwen: `(sort == null)?"":("?sort="+sort.toString().toLowerCase(Locale.ENGLISH))))`
this line is too difficult to understand - can you use separate method for this logic?

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -506,6 +510,36 @@ public void delete() throws IOException {
...
+                    @Override
+                    protected void wrapUp(GHRepository[] page) {
+                        for (GHRepository c : page)
```
lanwen: please use brackets if there only one line inside

src/main/java/org/kohsuke/github/GHRepository.java
marc-guenther: Actually it is quite easy, if a sort is given, use it, otherwise don't.

Sorry, but I don't know how to write this more clearly...

src/main/java/org/kohsuke/github/GHRepository.java
marc-guenther: This has been copied verbatim from other list* method, and it is like this all over the place in this file.

Is this really a reason to not merge this pull request?

KostyaSha: > This has been copied verbatim from other list* method, and it is like this all over the place in this file.

Don't follow bad practises :)

marc-guenther: Well, different people have quite different opinions on these kind of things, and as I don't know this particular project's conventions, I decided to stick with what I found in the existing source...
I also didn't think this is such a big deal...

lanwen: oracle guide is main document on how to format code. If no rules in project, oracle guide should be used

src/main/java/org/kohsuke/github/GHRepository.java
lanwen: 1. Spaces. Use them.
2. 
```
public PagedIterator<GHRepository> iterator() {
               String sortParam = ("?sort=" + );
               return new PagedIterator<GHRepository>(root.retrieve()
// statically imported org.apache.commons.lang.ObjectUtils
.with("sort", defaultIfNull(sort, Sort.NEWEST).name().toLowerCase())
.asIterator(getApiTailUrl("forks" + sortParam)), GHRepository[].class)) {
```

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -506,6 +510,36 @@ public void delete() throws IOException {
     }
 
     /**
+     * Sort orders for listing forks
+     */
+    public static enum Sort { NEWEST, OLDEST, STARGAZERS }
```
lanwen: Change name to ForksSort, as it already 3 Sort classes

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -506,6 +510,36 @@ public void delete() throws IOException {
...
+
+    /**
+     * Lists all the direct forks of this repository, sorted by {@link Sort#NEWEST Sort.NEWEST}
+     */
+    public PagedIterable<GHRepository> listForks() {
+      return listForks(null);
```
lanwen: why just not to pass default value and not handle null?

marc-guenther: Because I don't know the default. It is defined by github, and they do not expose it, except in their documentation.
So my javadoc is wrong, it should read "sorted by whatever github api defined as the default sort order, currently NEWEST"

src/main/java/org/kohsuke/github/GHRepository.java
marc-guenther: That is your opinion, obviously not that of the original author.

I find it more important to follow existing conventions when making such a small contribution to an existing bigger project, rather than to impose my current opinion about what is good or bad coding style onto it. I consider consistency to be more important than personal taste. (and I completely agree with you that you always should use brackets, but that is besides the point)

I also smell some serious case of bikeshedding here...

src/main/java/org/kohsuke/github/GHRepository.java
marc-guenther: Again, I looked how similar stuff was implemented, so yes,  I know there are 3 other Sort enums. Mine is also the only one with javadoc.

lanwen: You shouldn't copypaste all code and change only request, you should implement new feature in the most actual way. Or this lib became unmaintainable soon.

src/main/java/org/kohsuke/github/GHRepository.java
lanwen: Lot of minor issues make library unstable, untestable and difficult to read and hard to improve

src/main/java/org/kohsuke/github/GHRepository.java
marc-guenther: Everyone adding code in their own particular coding style also makes this library hard to read and understand. Also, this is not some cool new feature that needed to be implemented, I just added one missing api method (out of, I don't know, a hundred?), that for some reason has been overlooked in the past.

Implementing this in a "new way" by someone like me who doesn't know this library at all is way more error prone, as I don't know the particulars of the currently chosen approach.

Also now, if someone wants to clean up this code, at least everything looks the same, which imho makes cleaning up much easier than if everything looks different.

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -37,6 +37,7 @@
 import java.util.*;
 
 import static java.util.Arrays.asList;
+import static org.apache.commons.lang.ObjectUtils.defaultIfNull;
```
lanwen: you dont need it if you dont use it :)

lanwen:
@marc-guenther, thanks for update, :+1: for me now

src/main/java/org/kohsuke/github/GHRepository.java
marc-guenther: Oh, lol, I put it in and then forgot to remove it again ;)

PR #187, opened 2015-05-01
Recognize previous_file field in GHCommit.File

yegorius:
This field appears in API response in case file has moved.


oleg-nenashev:
:+1:

PR #189, opened 2015-05-09
fixed regression that caused POST operations to be sent as GET

if6was9:
This fixes a major regression where operations that are supposed to POST data to the server were actually just issuing GETs.


oleg-nenashev:
:+1:

PR #190, opened 2015-05-09
allow default branch to be set

if6was9:
GHRepository did not provide a method to set the default branch. 

Note that https://github.com/kohsuke/github-api/pull/189 is required for this to work.

### DIFF
--- a/src/main/java/org/kohsuke/github/GHRepository.java
+++ b/src/main/java/org/kohsuke/github/GHRepository.java
@@ -494,6 +494,12 @@ public class GHRepository extends GHObject {
         edit("homepage",value);
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -494,6 +494,12 @@ public void setHomepage(String value) throws IOException {
         edit("homepage",value);
     }
 
+    public void setMasterBranch(String value) throws IOException {
```
oleg-nenashev: I suppose the method should be called <code>setDefaultBranch</code>

KostyaSha: +1

PR #192, opened 2015-05-15
Enable creation and retrieval of org webhooks

chrisrhut:
made GHHook abstract and created two concrete subclasses for org
and repo hooks. Created utility class GHHooks to manage creation
and retrieval of org/repo hooks with minimal code duplication. These
are invoked by GHOrganization and GHRepository respectively.

Additional info on org webhooks here: https://developer.github.com/v3/orgs/hooks/

Note: requires #189 to test creation of org/repo hooks

### DIFF
--- a/src/main/java/org/kohsuke/github/GHHook.java
+++ b/src/main/java/org/kohsuke/github/GHHook.java
@@ -11,21 +11,11 @@ import java.util.Map;
 /**
  * @author Kohsuke Kawaguchi
  */
-public class GHHook extends GHObject {
-    /**
-     * Repository that the hook belongs to.
-     */
-    /*package*/ transient GHRepository repository;
-    
+public abstract class GHHook extends GHObject {
     String name;
     List<String> events;
     boolean active;
     Map<String,String> config;
-
-    /*package*/ GHHook wrap(GHRepository owner) {
-        this.repository = owner;
-        return this;
-    }
 
     public String getName() {
         return name;
@@ -50,7 +40,7 @@ public class GHHook extends GHObject {
      * Deletes this hook.
      */
     public void delete() throws IOException {
-        new Requester(repository.root).method("DELETE").to(String.format("/repos/%s/%s/hooks/%d", repository.getOwnerName(), repository.getName(), id));
+        new Requester(root()).method("DELETE").to(path());
     }
 
     /**
@@ -60,4 +50,8 @@ public class GHHook extends GHObject {
     public URL getHtmlUrl() {
         return null;
     }
+
+    abstract GitHub root();
+
+    abstract String path();
 }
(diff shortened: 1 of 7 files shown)


src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -57,14 +57,14 @@
     private int watchers,forks,open_issues,size,network_count,subscribers_count;
     private String pushed_at;
     private Map<Integer,GHMilestone> milestones = new HashMap<Integer, GHMilestone>();
-    
+
```
oleg-nenashev: please avoid such empty-line changes

src/main/java/org/kohsuke/github/GHHook.java
```diff
@@ -11,22 +11,12 @@
 /**
  * @author Kohsuke Kawaguchi
  */
-public class GHHook extends GHObject {
-    /**
-     * Repository that the hook belongs to.
-     */
-    /*package*/ transient GHRepository repository;
-    
+public abstract class GHHook extends GHObject {
```
oleg-nenashev: No way. It breaks the binary compatibility of the code

src/main/java/org/kohsuke/github/GHHook.java
```diff
@@ -60,4 +50,8 @@ public void delete() throws IOException {
     public URL getHtmlUrl() {
         return null;
     }
+
+    abstract GitHub root();
```
oleg-nenashev: * Implement methods with a default behavior instead.
* Oracle Java codestyle requires getRoot() and getPath()
* Javadocs are required as well. Also add <code>@since TODO</code> to document the API changes properly

src/main/java/org/kohsuke/github/GHHooks.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+
+/**
+ * Utility class for creating and retrieving webhooks; removes duplication between GHOrganization and GHRepository
+ * functionality
+ */
+class GHHooks {
```
oleg-nenashev: GHHooksHelper?

src/main/java/org/kohsuke/github/GHHooks.java
```diff
@@ -0,0 +1,130 @@
...
+            return GHOrgHook.class;
+        }
+
+        @Override
+        GHHook wrap(GHHook hook) {
+            return ((GHOrgHook)hook).wrap(organization);
+        }
+    }
+
+      static Context repoContext(GHRepository repository, GHUser owner) {
```
oleg-nenashev: Does it make sense to create new objects every operation?
I would also rename the method

oleg-nenashev:
I like the idea with the helper class BTW

PR #193, opened 2015-05-27
Fix bug in Requester. The body would never get assembled.

yegorius:
Eliminate dead code branch in Requester::buildRequest().


kohsuke:
It looks like I end up picking up a fix for this problem in PR #200.

PR #198, opened 2015-06-11
fix for GH Enterprise which does not have rate limit reset field

lanwen:
[Review on Reviewable](https://reviewable.io/reviews/kohsuke/github-api/198)


KostyaSha:
:+1: fixes regression caused my commit https://github.com/kohsuke/github-api/commit/a4c1c8de24f7f1ab38b28e5e6f19cc46a5993d5d

oleg-nenashev:
:+1:

PR #200, opened 2015-06-15
fix for unused json map when method with body, but body is null

lanwen:
fixes regression from b976e0ef4ef70f26b8d75a1a847b251f8c895e62

[Review on Reviewable](https://reviewable.io/reviews/kohsuke/github-api/200)


KostyaSha:
:+1:

PR #203, opened 2015-06-26
GitHub API have changed the semantics of /user/repos API

lucamilanesio:
It seems that since a couple of days (or weeks?) the /user/repos is returning
    ALL the repositories that the user has access or collaborates to whilst previously
    were only the ones that belong to him.
    
    JavaDoc updated in order to avoid getting unwanted results and introduced as well
    the possibility to filter a specific type of repository to be returned:
    - All (the GitHub's default)
    - Owner (the user's repos)
    - Public / Private (public or private repos)
    - Member (the user collaborates to)

### DIFF
--- a/src/main/java/org/kohsuke/github/GHMyself.java
+++ b/src/main/java/org/kohsuke/github/GHMyself.java
@@ -17,6 +17,18 @@ import java.util.TreeMap;
  * @author Kohsuke Kawaguchi
  */
 public class GHMyself extends GHUser {
+
+    /**
+     * Type of repositories returned during listing.
+     */
+    public enum RepositoryType {
+        ALL,  // All public and private repositories that current user has access or collaborates to
+        OWNER, // Public and private repositories owned by current user
+        PUBLIC, // Public repositories that current user has access or collaborates to
+        PRIVATE, // Private repositories that current user has access or collaborates to
+        MEMBER; // Public and private repositories that current user is a member
+    }
+
     /**
      * @deprecated
      *      Use {@link #getEmails2()}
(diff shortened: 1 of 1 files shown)


lucamilanesio:
@kohsuke this patch is *very urgent* since GitHub API have been rolled out two days ago with the new semantics :-( Can you have a look today if makes sense?

src/main/java/org/kohsuke/github/GHMyself.java
```diff
@@ -17,6 +17,18 @@
  * @author Kohsuke Kawaguchi
  */
 public class GHMyself extends GHUser {
+
+    /**
+     * Type of repositories returned during listing.
+     */
+    public enum RepositoryType {
+        All,  // All public and private repositories that current user has access or collaborates to
```
lanwen: Enum should be CAPITAL_CASED

you can rewrite toString() method if want to change string representation

src/main/java/org/kohsuke/github/GHMyself.java
```diff
@@ -109,16 +121,31 @@
...
+     *
+     * @param pageSize size for each page of items returned by GitHub. Maximum page size is 100.
+     * @param repoType type of repository returned in the listing
+     */
+    public PagedIterable<GHRepository> listRepositories(final int pageSize, final RepositoryType repoType) {
         return new PagedIterable<GHRepository>() {
             public PagedIterator<GHRepository> iterator() {
-                return new PagedIterator<GHRepository>(root.retrieve().asIterator("/user/repos?per_page=" + pageSize, GHRepository[].class)) {
+                return new PagedIterator<GHRepository>(root.retrieve().asIterator("/user/repos?per_page=" + pageSize +
+                    "&type=" + repoType.toString().toLowerCase(), GHRepository[].class)) {
```
lanwen: better to use name().toLowerCase() in that case

lanwen: also, can use `String.format("/user/repos?per_page=%s&type=%s", pageSize,  repoType)` if you override toString in enum

src/main/java/org/kohsuke/github/GHMyself.java
lucamilanesio: Done. I've decided to use name().toLowerCase() instead of relying on the toString() override.

src/main/java/org/kohsuke/github/GHMyself.java
lucamilanesio: decided to use name().toLowerCase() which is more explicit on what it does :-)

lucamilanesio:
@lanwen all changes done, thanks for your feedback.

PR #210, opened 2015-07-19
Cleanup issues discovered by FindBugs

oleg-nenashev:
The change closes about 100 issues (mostly API ones), but there're also several changes for encodings and file streams

@reviewbybees @lanwen @KostyaSha

### DIFF
--- a/src/main/java/org/kohsuke/github/GHNotificationStream.java
+++ b/src/main/java/org/kohsuke/github/GHNotificationStream.java
@@ -152,7 +152,7 @@ public class GHNotificationStream implements Iterable<GHThread> {
                         while (true) {
                             long now = System.currentTimeMillis();
                             if (nextCheckTime < now) break;
-                            long waitTime = Math.max(Math.min(nextCheckTime - now, 1000), 60 * 1000);
+                            long waitTime = Math.min(Math.max(nextCheckTime - now, 1000), 60 * 1000);
                             Thread.sleep(waitTime);
                         }
 
@@ -180,7 +180,8 @@ public class GHNotificationStream implements Iterable<GHThread> {
             private long calcNextCheckTime() {
                 String v = req.getResponseHeader("X-Poll-Interval");
                 if (v==null)    v="60";
-                return System.currentTimeMillis()+Integer.parseInt(v)*1000;
+                long seconds = Integer.parseInt(v);
+                return System.currentTimeMillis() + seconds*1000;
             }
 
             public void remove() {
--- a/src/main/java/org/kohsuke/github/GHObject.java
+++ b/src/main/java/org/kohsuke/github/GHObject.java
@@ -1,6 +1,7 @@
 package org.kohsuke.github;
 
 import com.infradna.tool.bridge_method_injector.WithBridgeMethods;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 import java.io.IOException;
 import java.net.URL;
@@ -9,6 +10,8 @@ import java.util.Date;
 /**
  * Most (all?) domain objects in GitHub seems to have these 4 properties.
  */
+@SuppressFBWarnings(value = {"UWF_UNWRITTEN_PUBLIC_OR_PROTECTED_FIELD", "UWF_UNWRITTEN_FIELD", 
+    "NP_UNWRITTEN_FIELD"}, justification = "JSON API")
 public abstract class GHObject {
     protected String url;
     protected int id;
@@ -24,10 +27,6 @@ public abstract class GHObject {
     @WithBridgeMethods(value=String.class, adapterMethod="createdAtStr")
     public Date getCreatedAt() throws IOException {
         return GitHub.parseDate(created_at);
-    }
-
-    private Object createdAtStr(Date id, Class type) {
-        return created_at;
     }
 
     /**
@@ -57,13 +56,5 @@ public abstract class GHObject {
     @WithBridgeMethods(value=String.class, adapterMethod="intToString")
     public int getId() {
         return id;
-    }
-
-    private Object intToString(int id, Class type) {
-        return String.valueOf(id);
-    }
-
-    private Object urlToString(URL url, Class type) {
-        return url==null ? null : url.toString();
     }
 }
--- a/src/main/java/org/kohsuke/github/GHPersonSet.java
+++ b/src/main/java/org/kohsuke/github/GHPersonSet.java
@@ -10,6 +10,8 @@ import java.util.HashSet;
  * @author Kohsuke Kawaguchi
  */
 public class GHPersonSet<T extends GHPerson> extends HashSet<T> {
+    private static final long serialVersionUID = 1L;
+ 
     public GHPersonSet() {
     }
 
--- a/src/main/java/org/kohsuke/github/GHRelease.java
+++ b/src/main/java/org/kohsuke/github/GHRelease.java
@@ -70,7 +70,7 @@ public class GHRelease extends GHObject {
     }
 
     public Date getPublished_at() {
-        return published_at;
+        return new Date(published_at.getTime());
     }
 
     public GitHub getRoot() {
@@ -123,9 +123,14 @@ public class GHRelease extends GHObject {
 
         String url = format("https://uploads.github.com%s/releases/%d/assets?name=%s",
                 owner.getApiTailUrl(""), getId(), file.getName());
-        return builder.contentType(contentType)
-                .with(new FileInputStream(file))
+        FileInputStream istream = new FileInputStream(file);
+        try {
+            return builder.contentType(contentType)
+                .with(istream)
                 .to(url, GHAsset.class).wrap(this);
+        } finally {
+            istream.close();
+        }
     }
 
     public List<GHAsset> getAssets() throws IOException {
--- a/src/main/java/org/kohsuke/github/GHRepository.java
+++ b/src/main/java/org/kohsuke/github/GHRepository.java
@@ -25,6 +25,7 @@ package org.kohsuke.github;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.infradna.tool.bridge_method_injector.WithBridgeMethods;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.apache.commons.lang.StringUtils;
 
 import javax.xml.bind.DatatypeConverter;
@@ -33,7 +34,9 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.InterruptedIOException;
 import java.io.Reader;
+import java.io.UnsupportedEncodingException;
 import java.net.URL;
+import java.nio.charset.Charset;
 import java.util.*;
 
 import static java.util.Arrays.asList;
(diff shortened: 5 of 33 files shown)


src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -1101,11 +1111,13 @@ public GHContent getReadme() throws IOException {
     }
 
     public GHContentUpdateResponse createContent(String content, String commitMessage, String path) throws IOException {
-        return createContent(content.getBytes(), commitMessage, path, null);
+        return createContent(content, commitMessage, path, null);
     }
 
     public GHContentUpdateResponse createContent(String content, String commitMessage, String path, String branch) throws IOException {
-        return createContent(content.getBytes(), commitMessage, path, branch);
+        //TODO: UTF-8?
```
lanwen: yep, plz change to utf8

src/main/java/org/kohsuke/github/Requester.java
```diff
@@ -323,16 +323,19 @@ private boolean isMethodWithBody() {
...
+        final StringBuilder strBuilder = new StringBuilder(_tailApiUrl);
         if (!args.isEmpty()) {
             boolean first=true;
             try {
                 for (Entry a : args) {
-                    _tailApiUrl += first ? '?' : '&';
+                    strBuilder.append(first ? '?' : '&');
                     first = false;
-                    _tailApiUrl += URLEncoder.encode(a.key,"UTF-8")+'='+URLEncoder.encode(a.value.toString(),"UTF-8");
+                    strBuilder.append(URLEncoder.encode(a.key, "UTF-8"));
```
lanwen: just use .append.append.append

src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -128,7 +129,9 @@
         } else {
             if (password!=null) {
                 String authorization = (login + ':' + password);
-                encodedAuthorization = "Basic "+new String(Base64.encodeBase64(authorization.getBytes()));
+                //TODO: UTF-8?
+                final Charset charset = Charset.defaultCharset();
```
KostyaSha: @lanwen the same?

KostyaSha:
:+1:

src/main/java/org/kohsuke/github/GHRepository.java
oleg-nenashev: Is it a standard for GitHub?

src/main/java/org/kohsuke/github/GitHub.java
lanwen: yep

src/main/java/org/kohsuke/github/GHRepository.java
lanwen: yes. Content-type includes utf-8 as charset

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -942,7 +949,7 @@ private void verifyMine() throws IOException {
         if (!root.login.equals(owner.login))
             throw new IOException("Operation not applicable to a repository owned by someone else: "+owner.login);
     }
-
+
```
KostyaSha: typo?

src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -128,7 +129,13 @@
         } else {
             if (password!=null) {
                 String authorization = (login + ':' + password);
-                encodedAuthorization = "Basic "+new String(Base64.encodeBase64(authorization.getBytes()));
+                final Charset charset;
+                try {
+                    charset = Charset.forName("UTF-8");
+                } catch(Exception ex) {
```
KostyaSha: space

KostyaSha:
:+1:

src/main/java/org/kohsuke/github/Requester.java
oleg-nenashev: I would keep the code as is. Just to have a better readability

oleg-nenashev:
@KostyaSha 
I've squashed the stuff

lanwen:
:+1:

src/main/java/org/kohsuke/github/GHNotificationStream.java
```diff
@@ -180,7 +180,8 @@ GHThread fetch() {
             private long calcNextCheckTime() {
                 String v = req.getResponseHeader("X-Poll-Interval");
                 if (v==null)    v="60";
-                return System.currentTimeMillis()+Integer.parseInt(v)*1000;
+                long seconds = Integer.parseInt(v);
```
Shredder121: Wouldn't `Long.parseLong()` suit better here?

src/main/java/org/kohsuke/github/GHObject.java
```diff
@@ -26,10 +29,6 @@ public Date getCreatedAt() throws IOException {
         return GitHub.parseDate(created_at);
     }
 
-    private Object createdAtStr(Date id, Class type) {
-        return created_at;
-    }
```
Shredder121: This was the bridge method am I right?

src/main/java/org/kohsuke/github/GHPersonSet.java
```diff
@@ -10,6 +10,8 @@
  * @author Kohsuke Kawaguchi
  */
 public class GHPersonSet<T extends GHPerson> extends HashSet<T> {
+    private static final long serialVersionUID = 1L;
+
```
Shredder121: Next time when adding `serialVersionUID`s, I recommend running `serialver` and take the version UID that would have been generated anyway.

KostyaSha: For already existed classes yes.

src/main/java/org/kohsuke/github/GHNotificationStream.java
KostyaSha: Depends on what is in `X-Poll-Interval` but you may be right

src/main/java/org/kohsuke/github/GHPersonSet.java
Shredder121: Just noticed it, don't worry.
It's just a pro-tip for later.

src/main/java/org/kohsuke/github/GHNotificationStream.java
Shredder121: This is the same, I just noticed it.

oleg-nenashev: In any case this code does not handle parsing exceptions, which may strike back at some point

src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -128,7 +129,13 @@
         } else {
             if (password!=null) {
                 String authorization = (login + ':' + password);
-                encodedAuthorization = "Basic "+new String(Base64.encodeBase64(authorization.getBytes()));
+                final Charset charset;
+                try {
+                    charset = Charset.forName("UTF-8");
```
lanwen: can use Charsets.UTF_8?

src/main/java/org/kohsuke/github/Requester.java
lanwen: it not makes readability better

src/main/java/org/kohsuke/github/GitHub.java
Shredder121: The Guava `Charsets` yes.

Do you know what the lowest supported JRE version must be?

src/main/java/org/kohsuke/github/GHPersonSet.java
stephenc: For newly created classes the preference is to force the value to `1L` so that we know it was always set from the start

Shredder121: Ah okay, that's fine.
Then disregard my comment.

src/main/java/org/kohsuke/github/GitHub.java
stephenc: Guava is a dangerous dependency to rely on as they do not have the same backwards compat rules as Jenkins follows and you can end up breaking your plugins. Exposing core's Guava to plugins was a major mistake...

oleg-nenashev: It's not a Jenkins plugin. Theoretically we could add Guava to github-api and then shade it. But it may be an overkill

Shredder121: No, what I mean is, there is a `java.nio.charset.StandardCharsets`, but if the minimum supported version is 5 (like it says in the `pom.xml` then don't use it.
Then outlining it to a static method sounds better.

KostyaSha: imho time to bump to 1.7

Shredder121: That's a little radical statement.

src/main/java/org/kohsuke/github/GHObject.java
amuniz: Yes. I think it must not be removed (or remove both the method and the annotation).

Shredder121: If it is to be removed, then yes, also the annotation.

src/main/java/org/kohsuke/github/GitHub.java
stephenc: Jenkins 1.612+ is Java 1.7+ only... and Oracle Java 1.7 is End of Support

oleg-nenashev: GitHub API is being widely used outside Jenkins. It definitely makes sense to consider the migration, but it needs a wider discussion or a decision from the library owner (@kohsuke)

lanwen: `org.apache.commons.codec.Charsets.UTF_8` is available

src/main/java/org/kohsuke/github/GHRelease.java
```diff
@@ -70,7 +70,7 @@ public boolean isPrerelease() {
     }
 
     public Date getPublished_at() {
-        return published_at;
+        return new Date(published_at.getTime());
```
lanwen: can you explain this change?

Shredder121: It's a defensive copy.

src/main/java/org/kohsuke/github/GitHub.java
Shredder121: Commons Codec `Charsets` then has my preference.

src/main/java/org/kohsuke/github/GHRelease.java
```diff
@@ -123,9 +123,14 @@ public GHAsset uploadAsset(File file, String contentType) throws IOException {
 
         String url = format("https://uploads.github.com%s/releases/%d/assets?name=%s",
                 owner.getApiTailUrl(""), getId(), file.getName());
-        return builder.contentType(contentType)
-                .with(new FileInputStream(file))
+        FileInputStream istream = new FileInputStream(file);
+        try {
+            return builder.contentType(contentType)
+                .with(istream)
```
Shredder121: I would annotate `Requester.with()` like this:
```java
    public Requester with(@WillClose InputStream body) { /*etc*/}
```
The closing is already handled in `Requester`. :(

PR #214, opened 2015-08-24
Making Javadoc clarify a perhaps surprising behavior

jglick:
Was the cause of hours of confusion. Changing the behavior now might break existing clients, so just noting the better alternative.

@reviewbybees

### DIFF
--- a/src/main/java/org/kohsuke/github/GHOrganization.java
+++ b/src/main/java/org/kohsuke/github/GHOrganization.java
@@ -235,22 +235,20 @@ public class GHOrganization extends GHPerson {
     }
 
     /**
-     * Beware that this lists only <em>public</em> repositories.
-     * Probably you wanted to use {@link #listRepositories(int)} instead.
-     * <p>
      * {@inheritDoc}
+     * @deprecated Beware that this lists only <em>public</em> repositories. Probably you wanted to use {@link #listRepositories(int)} instead.
      */
+    @Deprecated
     @Override
     public synchronized Map<String, GHRepository> getRepositories() throws IOException {
         return super.getRepositories();
     }
 
     /**
-     * Beware that this lists only <em>public</em> repositories.
-     * Probably you wanted to use {@link #listRepositories(int)} instead.
-     * <p>
      * {@inheritDoc}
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
Would it make sense to mark methods as deprecated ones and to provide more explicit ones? E.g. <code>getPublicRepositories()</code> and <code>getAllRepositories</code> (didn't check Github API docs)

jglick:
Well `listRepositories(int)` already works fine, but probably deprecating these overrides would be wise.

oleg-nenashev:
:bee:

tfennelly:
:bee:

src/main/java/org/kohsuke/github/GHOrganization.java
```diff
@@ -235,11 +235,29 @@ protected void wrapUp(GHEventInfo[] page) {
...
+        return super.getRepositories();
+    }
+
+    /**
+     * {@inheritDoc}
+     * @deprecated Beware that this lists only <em>public</em> repositories. Probably you wanted to use {@link #listRepositories(int)} instead.
+     */
+    @Deprecated
+    @Override
+    public PagedIterable<GHRepository> listRepositories() {
```
kohsuke: :bug: This version is a short-cut for `listRepositories(int)`. How come this version is deprecated but the other one is not? What am I missing?

jglick: > This version is a short-cut for `listRepositories(int)`.

Well that is what you would expect, of course! But that is not at all what it does. The `super` implementation is in `GHPerson` and calls a totally different GitHub API with different behavior. Hence this PR.

kohsuke:
I talked with Jesse and this still makes no sense.

The call sequence should be `GHOrganization.listRepositories(void)` → `GHPerson.listRepositories(void)` → `GHOrganization.listRepositories(30)` and so `listRepositories(void)` should behave the same way as `listRepositories(int)`.

The same goes for `getRepositories()`, which is implemented in terms of `listRepositories(void)`

Jesse is going to double check what actually happened.

jglick:
Indeed I can no longer reproduce the problem that originally led me to file this. Not sure what happened here. I must have inadvertently fixed some other bug and misdiagnosed the root cause. Sorry for noise.

PR #216, opened 2015-09-01
#215 fix read() failure with private repos

### LINKED ISSUE #215: GHContent.read() is broken due to incorrect HTTP Method
```GHContent.read()``` issues the request as POST rather than GET.  This is fine for URLs that do not
require authentication.  However, when ```download_url``` links are returned that have tokens in the query string, the corresponding POST fails.

Presumably it fails because on the remote end, the query string parameters of the POST request are not honored.

This issue would never be detected if testing happens against a public GitHub URL.

if6was9:
This fixes #215

### DIFF
--- a/src/main/java/org/kohsuke/github/GHHooks.java
+++ b/src/main/java/org/kohsuke/github/GHHooks.java
@@ -21,8 +21,9 @@ class GHHooks {
         }
 
         public List<GHHook> getHooks() throws IOException {
-            List<GHHook> list = new ArrayList<GHHook>(Arrays.asList(
-                root.retrieve().to(collection(), collectionClass())));
+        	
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
Should the PR also modify <code> asHttpStatusCode</code>? Otherwise it seems to be quite inconsistent

oleg-nenashev:
There are many such request patterns in the class...

if6was9:
Possibly.  read() is 100% broken on private repos, though.

getContent() works but it is deprecated. 

This does fix the issue and it seems unlikely to cause regression.  
 
I'll add a MockWebServer test if you support it.  OkHttp is used so I suspect it is ok?

oleg-nenashev:
@if6was9 
Any new tests will be appreciated :)

oleg-nenashev:
@if6was9 
Are you waiting for an additional review?

PR #217, opened 2015-09-14
Support Milestone closed_at date

dblevins:


src/main/java/org/kohsuke/github/GHMilestone.java
```diff
@@ -17,6 +17,7 @@
     GHUser creator;
     private String state, due_on, title, description, html_url;
     private int closed_issues, open_issues, number;
+    protected String closed_at;
```
lanwen: why it protected?

dblevins:
Private is fine.  Just copied from the parent class.

Issue #218, opened 2015-09-21
GHRepository.getCompare(GHBranch, GHBranch) does not allow for cross-fork compares


if6was9:
GHRepository.getCompare(String,String) supports cross-fork comparisons, but this does not:

https://github.com/kohsuke/github-api/blob/master/src/main/java/org/kohsuke/github/GHRepository.java#L684-L686

This is a simple fix.

kohsuke:
Solved by merging PR.

PR #219, opened 2015-09-21
#218 enable cross fork compare

if6was9:
this addresses #218

### DIFF
--- a/src/main/java/org/kohsuke/github/GHRepository.java
+++ b/src/main/java/org/kohsuke/github/GHRepository.java
@@ -682,7 +682,23 @@ public class GHRepository extends GHObject {
     }
 
     public GHCompare getCompare(GHBranch id1, GHBranch id2) throws IOException {
-        return getCompare(id1.getName(),id2.getName());
+
+        GHRepository owner1 = id1.getOwner();
+        GHRepository owner2 = id2.getOwner();
+
+        // If the owner of the branches is different, we have a cross-fork compare.
+        if (owner1!=null && owner2!=null) {
+            String ownerName1 = owner1.getOwnerName();
+            String ownerName2 = owner2.getOwnerName();
+            if (!StringUtils.equals(ownerName1, ownerName2)) {
+                String qualifiedName1 = String.format("%s:%s", ownerName1, id1.getName());
+                String qualifiedName2 = String.format("%s:%s", ownerName2, id2.getName());
+                return getCompare(qualifiedName1, qualifiedName2);
+            }
+        }
+
+        return getCompare(id1.getName(), id2.getName());
+
     }
 
     /**


src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -682,7 +682,25 @@ public GHCompare getCompare(GHCommit id1, GHCommit id2) throws IOException {
...
+
+        try {
+            // try to grab the org name and qualify the branch name
+            String owner1 = id1.getOwner().getOwnerName();
+            String owner2 = id2.getOwner().getOwnerName();
+            if (!owner1.equals(owner2)) {
+                fqid1 = String.format("%s:%s",owner1,id1.getName());
+                fqid2 = String.format("%s:%s",owner2,id2.getName());
+            }
+        }
```
lanwen: catch should be after } -> `} catch {`

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -682,7 +682,25 @@ public GHCompare getCompare(GHCommit id1, GHCommit id2) throws IOException {
...
+        catch (RuntimeException e) {
```
lanwen: bad practice to catch everything and ignore. Should catch only when we request owner.

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -682,7 +682,25 @@ public GHCompare getCompare(GHCommit id1, GHCommit id2) throws IOException {
...
-        return getCompare(id1.getName(),id2.getName());
+
+        String fqid1 = id1.getName();
+        String fqid2 = id2.getName();
+
+        try {
+            // try to grab the org name and qualify the branch name
+            String owner1 = id1.getOwner().getOwnerName();
+            String owner2 = id2.getOwner().getOwnerName();
+            if (!owner1.equals(owner2)) {
```
lanwen: use StringUtils.equals(s1, s2) as NPE-safe version

lanwen:
:-1: For now.

I think it should be:
1. Get the owners. (Can it be exception here? If not, skip 2 step)
2. Use default owner names if exc.
3. If owners equals, use `getCompare(id1.getName(), id2.getName());`
4. Else format owner:branch and use `getCompare` with that

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -682,7 +682,24 @@ public GHCompare getCompare(GHCommit id1, GHCommit id2) throws IOException {
     }
 
     public GHCompare getCompare(GHBranch id1, GHBranch id2) throws IOException {
+
```
lanwen: please avoid empty lines in count more than one

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -682,7 +682,23 @@ public GHCompare getCompare(GHCommit id1, GHCommit id2) throws IOException {
...
+        GHRepository owner1 = id1.getOwner();
+        GHRepository owner2 = id2.getOwner();
+
+        // If the owner of the branches is different, we have a cross-fork compare.
+        if (owner1!=null && owner2!=null) {
+            String ownerName1 = owner1.getOwnerName();
+            String ownerName2 = owner2.getOwnerName();
+            if (!StringUtils.equals(ownerName1, ownerName2)) {
+                String qualifiedName1 = String.format("%s:%s",ownerName1,id1.getName());
```
lanwen: please add space after `,`

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -682,7 +682,23 @@ public GHCompare getCompare(GHCommit id1, GHCommit id2) throws IOException {
...
+                String qualifiedName2 = String.format("%s:%s",ownerName2,id2.getName());
+                return getCompare(qualifiedName1,qualifiedName2);
```
lanwen: and here

lanwen:
:+1: after some code style fixes

oleg-nenashev:
As I see from the code, null owners are not supposed to happen. BTW I'm not sure I've investigated all potential cases.

:+1: in any case, because the original behavior smells like a bug

PR #224, opened 2015-10-06
Remove trailing slash when requesting directory content

Shredder121:
I noticed something strange on GitHub's end.

When requesting content from a directory (with a specific ref), you get [ref-specific content][1].

Now look closely what happens when you [add a trailing slash][2].

What is your view on this?

[1]: https://api.github.com/repos/github/developer.github.com/contents/lib/webhooks?ref=21295b477e6727ae09c6691e78fca7a33d28b54d
[2]: https://api.github.com/repos/github/developer.github.com/contents/lib/webhooks/?ref=21295b477e6727ae09c6691e78fca7a33d28b54d

### DIFF
--- a/src/main/java/org/kohsuke/github/GHRepository.java
+++ b/src/main/java/org/kohsuke/github/GHRepository.java
@@ -1094,6 +1094,9 @@ public class GHRepository extends GHObject {
 
     public List<GHContent> getDirectoryContent(String path, String ref) throws IOException {
         Requester requester = root.retrieve();
+        while (path.endsWith("/")) {
+            path = path.substring(0, path.length() - 1);
+        }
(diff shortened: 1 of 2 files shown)


Shredder121:
Could someone restart the build :smile: 
I first saw a mistake, so pushed a new solution.

Shredder121:
Only merge if it's an issue that we _should_ fix.
And only then I still want to add in a test case.

oleg-nenashev:
I have no strong opinion about the fix, but it looks reasonable to me.
CC @KostyaSha

KostyaSha:
Sorry, have no opinion. Will it be possible to make some test case?

Shredder121:
I did so just now.

PR #225, opened 2015-10-08
Overzealous FindBugs changes.

Shredder121:
Charsets that are standard on the JRE are try-lookuped,
bridge methods were removed and a stream that would be closed later is closed explicitly

### DIFF
--- a/src/main/java/org/kohsuke/github/GitHub.java
+++ b/src/main/java/org/kohsuke/github/GitHub.java
@@ -47,6 +47,7 @@ import java.util.Set;
 import java.util.TimeZone;
 import java.util.concurrent.TimeUnit;
 
+import org.apache.commons.codec.Charsets;
 import org.apache.commons.codec.binary.Base64;
 
 import com.fasterxml.jackson.databind.DeserializationFeature;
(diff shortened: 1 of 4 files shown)


src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -129,12 +130,7 @@
         } else {
             if (password!=null) {
                 String authorization = (login + ':' + password);
-                final Charset charset;
-                try {
-                    charset = Charset.forName("UTF-8");
-                } catch (Exception ex) {
-                    throw new IOException("UTF-8 encoding is not supported", ex);
-                }
+                Charset charset = Charsets.UTF_8;
```
lanwen: should be inlined with static import i think

Shredder121: If more people agree I'll do that then.
Shall I remove the variable declaration and inline it to the method call or just convert to static import?

oleg-nenashev:
IIRC the problem with charsets was related to JVM 6. I'm not sure how Apache codecs behave there, needs to be checked

PR #226, opened 2015-10-09
Check builder result to either be a token or a user

Shredder121:
Currently, a `user` property is always required (it not having content is also fine).

This adds support for only having the `oauth` key in the property file/environment.


Shredder121:
@kohsuke sorry to be bothering you, have you had time to look at my PRs yet?

oleg-nenashev:
:+1:  from me

PR #231, opened 2015-11-15
Support for merge_commit_sha

recena:
`GHPullRequest` object now includes `merge_commit_sha`

@reviewbybees specially, @oleg-nenashev

### DIFF
--- a/src/main/java/org/kohsuke/github/GHPullRequest.java
+++ b/src/main/java/org/kohsuke/github/GHPullRequest.java
@@ -189,7 +189,7 @@ public class GHPullRequest extends GHIssue {
     }
 
     /**
-     * FYI: https://developer.github.com/changes/2013-04-25-deprecating-merge-commit-sha
+     * See <a href="https://developer.github.com/changes/2013-04-25-deprecating-merge-commit-sha">GitHub blog post</a>
      */
     @Deprecated
     public String getMergeCommitSha() throws IOException {
--- a/src/main/java/org/kohsuke/github/GHPullRequest.java
+++ b/src/main/java/org/kohsuke/github/GHPullRequest.java
@@ -50,8 +50,6 @@ public class GHPullRequest extends GHIssue {
     private int deletions;
     private String mergeable_state;
     private int changed_files;
-
-    @Deprecated
     private String merge_commit_sha;
 
     /**
@@ -191,7 +189,6 @@ public class GHPullRequest extends GHIssue {
     /**
      * See <a href="https://developer.github.com/changes/2013-04-25-deprecating-merge-commit-sha">GitHub blog post</a>
      */
-    @Deprecated
     public String getMergeCommitSha() throws IOException {
         populate();
         return merge_commit_sha;
--- a/src/main/java/org/kohsuke/github/GHPullRequest.java
+++ b/src/main/java/org/kohsuke/github/GHPullRequest.java
@@ -50,6 +50,8 @@ public class GHPullRequest extends GHIssue {
     private int deletions;
     private String mergeable_state;
     private int changed_files;
+
+    @Deprecated
     private String merge_commit_sha;
 
     /**
@@ -186,6 +188,10 @@ public class GHPullRequest extends GHIssue {
         return changed_files;
     }
 
+    /**
+     * FYI: https://developer.github.com/changes/2013-04-25-deprecating-merge-commit-sha
+     */
+    @Deprecated
     public String getMergeCommitSha() throws IOException {
         populate();
         return merge_commit_sha;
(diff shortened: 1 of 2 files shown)


src/main/java/org/kohsuke/github/GHPullRequest.java
```diff
@@ -186,6 +189,15 @@ public int getChangedFiles() throws IOException {
     }
 
     /**
+     * FYI: https://developer.github.com/changes/2013-04-25-deprecating-merge-commit-sha
+     */
+    @Deprecated
+    public String getMergeCommitSha() throws IOException {
```
oleg-nenashev: What happens if the field disappears? Does the method return null in such case?

recena: When this happens, GitHub API v4 will be released and we should review the current status (compatibility).

src/main/java/org/kohsuke/github/GHPullRequest.java
```diff
@@ -142,9 +145,9 @@ public PullRequest getPullRequest() {
     }
 
     //
-// details that are only available via get with ID
-//
-//
```
recena: Unrelated changes but...

src/main/java/org/kohsuke/github/GHPullRequest.java
recena: @oleg-nenashev Any recommendation here to improve it?

recena: @oleg-nenashev I've checked that if your model has an attribute and is not mapped in the JSON, its `get()` method returns `null`. I think that is ok.

src/main/java/org/kohsuke/github/GHPullRequest.java
oleg-nenashev: Makes sense to convert the comment to Javadoc then :)

lanwen:
why you want to use deprecated field?

src/main/java/org/kohsuke/github/GHPullRequest.java
amuniz: Right. [This line](https://github.com/recena/github-api/blob/b0789a7ce72b6e1773b47d4e778535f839863480/src/main/java/org/kohsuke/github/GitHub.java#L564) is preventing to fail on missing properties.

KostyaSha:
@lanwen because it can be valid for some unknown time or on some ancient enterprise installations?

src/main/java/org/kohsuke/github/GHPullRequest.java
```diff
@@ -186,6 +189,15 @@ public int getChangedFiles() throws IOException {
     }
 
     /**
+     * FYI: https://developer.github.com/changes/2013-04-25-deprecating-merge-commit-sha
```
KostyaSha: javadoc `@link`?

amuniz:
:bee:

src/main/java/org/kohsuke/github/GHPullRequest.java
lanwen: javadoc `@deprecated`

recena: Sure.

src/main/java/org/kohsuke/github/GHPullRequest.java
recena: Really, it is not a method comment, only a note in the middle of the file.

src/main/java/org/kohsuke/github/GHPullRequest.java
KostyaSha: @oleg-nenashev `@CheckForNull`?

recena:
@reviewbybees done

oleg-nenashev:
:+1:

jglick:
Does not make sense to introduce a deprecated API, especially as this will probably still working in v4.

KostyaSha:
probably makes sense call variable in camel case according to java convention and use Json annotation.

recena:
@jglick if you don't need to use it, simply ignore it. GH API V4 is not planned, there is not a date. While this happens this attribute is the only way to resolve some problems.

recena:
@jglick make sense to include not documented behaviors? For example, retrieve `/merge` spec???

recena:
@KostyaSha I agree but I tried to follow the current conventions in this project.

recena:
@jglick I apologize. I understood wrong your comment.I should read twice before to reply. Specially, when I read english text.

recena:
@jglick I'm going to remove the tag `@decrecated`.

recena:
@jglick `@deprecated` annotations were removed.

ndeloof:
LGTM

recena:
@reviewbybees done

jglick:
Belated :bee:.

I do not think we really want to be using this API. But I see no harm in exposing it via the library. It is up to the caller whether to rely on it or not. I presume that if and when the field is deleted from the JSON by GH, `getMergeCommitSha` will simply return null and that is that.

recena:
@jglick If a field is removed from GitHub API, its getter asociated will return `null`.

PR #232, opened 2015-11-15
Added a new method to validate the GitHub API URL

recena:
@reviewbybees specially, @oleg-nenashev

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <findbugs-maven-plugin.version>3.0.1</findbugs-maven-plugin.version>
+    <findbugs-maven-plugin.version>3.0.2</findbugs-maven-plugin.version>
     <findbugs-maven-plugin.failOnError>true</findbugs-maven-plugin.failOnError>
   </properties>
 
@@ -52,7 +52,6 @@
         <version>${findbugs-maven-plugin.version}</version>
         <configuration>
           <xmlOutput>true</xmlOutput>
-          <findbugsXmlWithMessages>true</findbugsXmlWithMessages>
           <failOnError>${findbugs-maven-plugin.failOnError}</failOnError>
         </configuration>
         <executions>
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <findbugs-maven-plugin.version>3.0.2</findbugs-maven-plugin.version>
+    <findbugs-maven-plugin.version>3.0.1</findbugs-maven-plugin.version>
     <findbugs-maven-plugin.failOnError>true</findbugs-maven-plugin.failOnError>
   </properties>
 
@@ -52,6 +52,7 @@
         <version>${findbugs-maven-plugin.version}</version>
         <configuration>
           <xmlOutput>true</xmlOutput>
+          <findbugsXmlWithMessages>true</findbugsXmlWithMessages>
           <failOnError>${findbugs-maven-plugin.failOnError}</failOnError>
         </configuration>
         <executions>
--- a/src/main/java/org/kohsuke/github/GitHub.java
+++ b/src/main/java/org/kohsuke/github/GitHub.java
@@ -205,6 +205,15 @@ public class GitHub {
     }
 
     /**
+     * Connects to GitHub Enterprise anonymously.
+     *
+     * All operations that requires authentication will fail.
+     */
+    public static GitHub connectToEnterpriseAnonymously(String apiUrl) throws IOException {
+        return new GitHubBuilder().withEndpoint(apiUrl).build();
+    }
+
+    /**
      * Is this an anonymous connection
      * @return {@code true} if operations that require authentication will fail.
      */
@@ -441,6 +450,23 @@ public class GitHub {
         try {
             retrieve().to("/user", GHUser.class);
             return true;
+        } catch (IOException e) {
+            return false;
+        }
+    }
+
+    /**
+     * Ensures that the API URL is valid.
+     */
+    public boolean isApiUrlValid() throws IOException {
+        try {
+            GHApiInfo apiInfo = retrieve().to("/", GHApiInfo.class);
+            try {
+                new URL(apiInfo.getRateLimitUrl());
+                return true;
+            } catch (MalformedURLException mue) {
+                return false;
+            }
         } catch (IOException e) {
             return false;
         }
--- a/src/main/java/org/kohsuke/github/GHApiInfo.java
+++ b/src/main/java/org/kohsuke/github/GHApiInfo.java
@@ -1,0 +1,14 @@
+package org.kohsuke.github;
+
+/**
+ * Represents information about Github API.
+ */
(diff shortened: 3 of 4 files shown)


oleg-nenashev:
Hm... PR builder is dead

recena:
@oleg-nenashev AFAIK, the test suite require local configuration. Was there a PR Builder configured in this repo?

oleg-nenashev:
Yep. It has been configured on Kohsuke's beescloud account. I'm looking for it

pom.xml
```diff
@@ -28,7 +28,7 @@
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <findbugs-maven-plugin.version>3.0.1</findbugs-maven-plugin.version>
+    <findbugs-maven-plugin.version>3.0.2</findbugs-maven-plugin.version>
```
KostyaSha: Unrelated to commit feature change.

src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -447,6 +456,23 @@ public boolean isCredentialValid() throws IOException {
     }
 
     /**
+     * Ensures that the API URL is valid.
+     */
+    public boolean isApiUrlValid() throws IOException {
+        try {
+            GHApiInfo apiInfo = retrieve().to("/", GHApiInfo.class);
+            try {
+                new URL(apiInfo.getRateLimitUrl());
```
KostyaSha: Why api url is mixed to ratelimit?

src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -205,6 +205,15 @@ public static GitHub connectAnonymously() throws IOException {
     }
 
     /**
+     * Connects to GitHub Enterprise anonymously.
```
KostyaSha: What is the difference between anonymous and not? Why library should care about specific logins?

src/main/java/org/kohsuke/github/GitHub.java
recena: it is an easy way to verify that the endpoint (API URL) replied a right response. [Here](https://api.github.com/) an example.

src/main/java/org/kohsuke/github/GitHub.java
recena: One wants to connect to GH Enterprise as anonymous and this was not possible until now.

recena: Analogous to `GitHub.connectAnonymously()`. [Reference](https://github.com/kohsuke/github-api/blob/master/src/main/java/org/kohsuke/github/GitHub.java#L203)

src/main/java/org/kohsuke/github/GitHub.java
KostyaSha: if rate_limit disappear this logic will be wrong.

pom.xml
recena: Indeed, I'll wait to know the maintainer's opinion.

src/main/java/org/kohsuke/github/GitHub.java
KostyaSha: AFAIR Many of connect methods are useless and exist only for backward compatibility. Right logic should be use exist connects and builders. I see no reason for making separate methods or using anonymous in any env at all.

KostyaSha: > Analogous to GitHub.connectAnonymously(). Reference

So you can't call `new GitHubBuilder().build()` from your code??

pom.xml
KostyaSha: Never mix 2 unrelated changes in one when it possible. FB absolutely not related to `Added a new method to validate the GitHub API URL`

src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -205,6 +205,15 @@ public static GitHub connectAnonymously() throws IOException {
...
+     *
+     * All operations that requires authentication will fail.
+     */
+    public static GitHub connectToEnterpriseAnonymously(String apiUrl) throws IOException {
```
lanwen: -1 for new method. Builder exists for such cases

pom.xml
oleg-nenashev: I agree with @KostyaSha

src/main/java/org/kohsuke/github/GitHub.java
recena: The same way than if disappear other attributes. This method will return `null` is this case. By the way, when it happens GitHub API v4 will be published and a review of this client will be required.

KostyaSha: method called `isApiUrlValid()` if it fetches and deserialized -> then true. 
If rate_limit valid -> other logic that can be checked even now.

src/main/java/org/kohsuke/github/GitHub.java
recena: Well, in this case we use the endpoint of a GH Enterprise server. I don't know where is the problem to add a new method.

src/main/java/org/kohsuke/github/GitHub.java
lanwen: why not just check status code and headers like
```
X-GitHub-Media-Type: github.v3
X-GitHub-Request-Id: 64c9ffd2-29dc-4a81-a85c-57fdba8369a4
```

src/main/java/org/kohsuke/github/GitHub.java
recena: But, it needs to provide the `apiUrl`.

src/main/java/org/kohsuke/github/GitHub.java
KostyaSha: This methods existed before normal builder appeared, use builder from your code as valid way of connection creation.

src/main/java/org/kohsuke/github/GitHub.java
recena: Because these header are returned in any service of GitHub API. And the goal of this method is to verify if `apiUrl` is really a valid endpoint.

src/main/java/org/kohsuke/github/GitHub.java
recena: I'm sorry but I don't understand your last comment. An example?

pom.xml
recena: @oleg-nenashev @KostyaSha I agree. I'll change it.

src/main/java/org/kohsuke/github/GitHub.java
lanwen: so why not to add a new method for every combination of builder methods? :)

recena: It could be other option. I prefer this.

kohsuke:
I have a mixed feeling about adding more `connectXyz` methods, but it is indeed convenient to have some of them. Given that @recena feels strongly about having this particular convenience method, I think it's OK to have it.

I've massaged this PR somewhat by the following commit.

recena:
@kohsuke Thanks for your time. I left `GHApiInfo` in a public class because we can need adding more fields in the future. I used `isApiUrlValid()` because we had `isCredentialsValid()`, I tried to follow the rules.

PR #233, opened 2015-11-19
Add information about mirror url if it exist.

vparfonov:
Like https://github.com/apache/tomee


recena:
:+1:

oleg-nenashev:
maybe needs some polishing, but :+1:

recena:
@oleg-nenashev For example? A test?

oleg-nenashev:
@recena 
No, I mostly think about the API documentation and annotations. BTW it's not important in the current project state. It needs a bulk update.

PR #245, opened 2016-01-20
Fix error when creating email service hook

daniel-beck:
The Jenkins IRC bot started throwing 

`{"message":"Invalid request.\n\nFor 'properties/active', \"true\" is not a boolean.","documentation_url":"https://developer.github.com/v3/repos/hooks/#create-a-hook"}`

Untested.


jglick:
I just got this now.

PR #247, opened 2016-02-11
Use GET method to fetch the GHontent's content

Shredder121:
This is a little more cache-friendly.

Let me know what you think.

### DIFF
--- a/src/main/java/org/kohsuke/github/GHContent.java
+++ b/src/main/java/org/kohsuke/github/GHContent.java
@@ -115,7 +115,7 @@ public class GHContent {
      * Retrieves the actual content stored here.
      */
     public InputStream read() throws IOException {
-        return new Requester(root).asStream(getDownloadUrl());
+        return root.retrieve().asStream(getDownloadUrl());
     }
 
     /**


kohsuke:
I believe this is already addressed in e0b109cba6faa84813a528880dbf3923c1b86058

Shredder121:
It unfortunately isn't.
Only the `HttpURLConnection.method` was set by that change. Not the `Requester.method`.

This means that `Requester.method` is still set to `POST`, `isMethodWithBody` will return true, and `uc.setDoOutput(true)` will be called.

I use Okhttp, and their [HttpURLConnectionImpl's method changes to `POST` if you tell it to open a stream to write to](http://grepcode.com/file/repo1.maven.org/maven2/com.squareup.okhttp/okhttp-urlconnection/2.0.0/com/squareup/okhttp/internal/huc/HttpURLConnectionImpl.java#292).

kohsuke:
Thanks, that explains. Now I claim dbc79f8c423749311495cb548c1a9cda1d8638cf fixes it!

Shredder121:
I indeed verified that it fixed it, thanks!
So I take it my PR is no longer necessary?

When can I expect to see this released?

Shredder121:
@kohsuke when can I expect to see this released?

PR #248, opened 2016-02-21
Populate commit with data for getCommitShortInfo

daniel-beck:


oleg-nenashev:
:-1: due to the binary compatibility issue.

kohsuke:
This is binary compatible.

PR #254, opened 2016-03-06
Better error messages

cyrille-leclerc:
Better error message: 
* Introduce HttpException, subclass of IOException with url, http responseCode and http responseMessage to help exception handling.
* Add a `FINEST` log message in `GitHub.isCredentialsValid()` in case of exception as we silently swallow this exception.

### DIFF
--- a/src/main/java/org/kohsuke/github/GitHub.java
+++ b/src/main/java/org/kohsuke/github/GitHub.java
@@ -31,6 +31,7 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.Reader;
+import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.text.ParseException;
@@ -55,6 +56,8 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.introspect.VisibilityChecker.Std;
 import com.infradna.tool.bridge_method_injector.WithBridgeMethods;
 import java.nio.charset.Charset;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /**
  * Root of the GitHub API.
@@ -68,6 +71,8 @@ import java.nio.charset.Charset;
  * @author Kohsuke Kawaguchi
  */
 public class GitHub {
+    protected final transient Logger logger = Logger.getLogger(getClass().getName());
+
     /*package*/ final String login;
 
     /**
@@ -458,6 +463,7 @@ public class GitHub {
             retrieve().to("/user", GHUser.class);
             return true;
         } catch (IOException e) {
+            logger.log(Level.FINEST, "Exception validating credentials on {0} with login {1}: {2}", new Object[]{this.apiUrl, this.login, e, e});
             return false;
         }
     }
--- a/src/main/java/org/kohsuke/github/GitHub.java
+++ b/src/main/java/org/kohsuke/github/GitHub.java
@@ -463,7 +463,8 @@ public class GitHub {
             retrieve().to("/user", GHUser.class);
             return true;
         } catch (IOException e) {
-            logger.log(Level.FINEST, "Exception validating credentials on {0} with login {1}: {2}", new Object[]{this.apiUrl, this.login, e, e});
+            if (logger.isLoggable(Level.FINE))
+                logger.log(Level.FINE, "Exception validating credentials on " + this.apiUrl + " with login '" + this.login + "' " + e, e);
             return false;
         }
     }
--- a/src/main/java/org/kohsuke/github/HttpException.java
+++ b/src/main/java/org/kohsuke/github/HttpException.java
@@ -1,0 +1,119 @@
+package org.kohsuke.github;
+
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+/**
+ * Http exception
+ *
+ * @author <a href="mailto:cleclerc@cloudbees.com">Cyrille Le Clerc</a>
+ */
+public class HttpException extends IOException {
+    static final long serialVersionUID = 1L;
+
+    private final int responseCode;
+    private final String responseMessage;
+    private final String url;
+
+    /**
+     * @param message         The detail message (which is saved for later retrieval
+     *                        by the {@link #getMessage()} method)
+     * @param responseCode    Http response code. {@code -1} if no code can be discerned.
+     * @param responseMessage Http response message
+     * @param url             The url that was invoked
+     * @see HttpURLConnection#getResponseCode()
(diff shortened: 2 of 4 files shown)


cyrille-leclerc:
@reviewbybees esp @recena @oleg-nenashev

src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -458,6 +463,7 @@ public boolean isCredentialValid() throws IOException {
             retrieve().to("/user", GHUser.class);
             return true;
         } catch (IOException e) {
+            logger.log(Level.FINEST, "Exception validating credentials on {0} with login {1}: {2}", new Object[]{this.apiUrl, this.login, e, e});
```
jglick: Probably `FINE` would be appropriate.

jglick: :bug: you are trying to pass four parameters to a three-argument message parameter; probably you meant to put the last `e` outside the braces, but this does not work because the preformatted overloads do not also accept `Throwable`. Also including `Throwable.toString` in the message is pointless when you are including the full stack trace anyway. So what you meant was probably something like

```java
logger.log(Level.FINE, "Exception validating credentials on " + apiUrl + " with login " + login, e);
```

src/main/java/org/kohsuke/github/HttpException.java
```diff
@@ -0,0 +1,119 @@
+package org.kohsuke.github;
+
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+/**
+ * Http exception
```
jglick: Useless Javadoc.

src/main/java/org/kohsuke/github/HttpException.java
```diff
@@ -0,0 +1,119 @@
...
+        this.responseMessage = responseMessage;
+        this.url = url;
+    }
+
+    /**
+     * @param message         The detail message (which is saved for later retrieval
+     *                        by the {@link #getMessage()} method)
+     * @param responseCode    Http response code. {@code -1} if no code can be discerned.
+     * @param responseMessage Http response message
+     * @param url             The url that was invoked
```
jglick: Is this intended to be `@CheckForNull`? The next overload seems to imply that it is.

src/main/java/org/kohsuke/github/HttpException.java
```diff
@@ -0,0 +1,119 @@
...
+     *
+     * @return url
+     */
+    public String getUrl() {
+        return url;
+    }
+
+    @Override
+    public String getMessage() {
+        return super.getMessage();
```
jglick: Useless, delete.

src/main/java/org/kohsuke/github/Requester.java
```diff
@@ -508,7 +520,15 @@ private InputStream wrapStream(InputStream in) throws IOException {
      * Handle API error by either throwing it or by returning normally to retry.
      */
     /*package*/ void handleApiError(IOException e) throws IOException {
-        if (uc.getResponseCode() == 401) // Unauthorized == bad creds
+        int responseCode;
+        try {
+            responseCode = uc.getResponseCode();
+        } catch (IOException e2) {
+            // likely to be a network exception (e.g. SSLHandshakeException),
```
jglick: Best to log at `FINE` just in case.

src/main/java/org/kohsuke/github/GitHub.java
cyrille-leclerc: Thanks, I'll never get on well with java.util.logging :-)

src/main/java/org/kohsuke/github/HttpException.java
cyrille-leclerc: Indeed deserves a `@CheckForNull` but as  `@CheckForNull` and `@Nonnull` are not used in this project and as we only have `javax.annotation.CheckForNull` that @stephenc don't like, I did not dare to add it.

I am happy to add `@javax.annotation.CheckForNull`.

cyrille-leclerc:
@jglick I have updated the code according to your recommendations!

src/main/java/org/kohsuke/github/HttpException.java
stephenc: I am not a maintainer of this plugin, so if the plugin maintainers want to have you use an annotation that is - in its current source - illegal under the terms of the JCP and may put you in breech of contract if you ever signed a JCP agreement, that is your and their business

cyrille-leclerc: @stephenc can you please remind me which annotation library is "the right one" and I'll commit the change.

stephenc: You need to ask the owner of this codebase what one they prefer.

Now as to which is correct:

* you are only allowed to use the `javax` package for authoring classes that you redistribute where the spec of those classes has been published via the JCP process.

* the JSR that was working on annotations to use for eg null status has not published anything at all. The only public statement is that they might introduce a `javax.annotations.NonNull` (which is not the one everyone is using you will note)

* as such we are breaking the good faith of the EULA or JCP etc agreements that maybe could - at an extreme stretch - be used to initiate proceedings... I doubt such proceedings would win... But then look at the Google - Oracle API appeal

So *where I have a choice* I will use the `edu.` version... under the non GPL license... Because in those cases I can choose to "do the right thing"... If there is ever even an official published draft of annotations for this in 'javax` I will switch to them

jglick: @stephenc since you seem to have a strong opinion would you mind proposing a patch to `plugin-pom` to forbid usage of the one(s) you do not like and/or enforce usage of the one(s) you do like, bearing in mind that `edu.umd.cs.findbugs.annotations.NonNull` was `@Deprecated` in `com.google.code.findbugs:annotations:3.0.0` and that `com.github.stephenc.findbugs:findbugs-annotations:1.3.9-1` is clearly outdated?

jglick: By the way I can guess that the owner of _this_ codebase is not really interested in the question and you can use whatever you find convenient.

cyrille-leclerc: @stephenc thanks, I am a bit stuck, net.sourceforge.findbugs:annotations is LGPL and thus incompatible with the MIT license of this project. com.github.stephenc.findbugs:findbugs-annotations is ASL2 and thus could be OK.

I'll wait for the owner of the code base to make a suggestion :-)

stephenc: I discussed this subject with the owner of the code base, and the owner has agreed that they personally will start making an effort to use the `edu` one.

I will merge the PR on my ASL clean room version of find bugs annotations tomorrow and release into central (without `@Deprecacted`)

What you choose to do is up to you

Issue #266, opened 2016-04-02
GHRelease.uploadAsset() does not work for Github Enterprise


lwiechec:
looking at [source code](https://github.com/kohsuke/github-api/blob/master/src/main/java/org/kohsuke/github/GHRelease.java#L130), the URL for uploading releases is hardcoded to Gighub.com's.

I will attempt to extend it for GitHub Enteprise, using `GitHub.getApiURL()`...

oleg-nenashev:
Yes, it's definitely a bug

lwiechec:
@oleg-nenashev  Thanks for reply :) I was wondering what would be the good course of action; after reading [the release upload API docs](https://developer.github.com/enterprise/2.5/v3/repos/releases/#upload-a-release-asset), I think we need first to 'ask' for the upload URL (via calling GitHub's API `/repos/{user}/{repo}/releases/{id}` for `upload_url` and then use it. Would you think it's correct?

lwiechec:
Hi, I have created a very simple fix, adding another version of `uploadAsset()` that accepts the upload URL.

kohsuke:
Please see my comment to PR #267

mpern:
I just stumbled across this issue, you can not upload assets to Github Enterprise

bitwiseman:
I believe this is fixed.  

@mpern @lwiechec  if you can reproduce, please comment with how steps to reproduce.

PR #267, opened 2016-04-04
'uploadAsset()' method call for Github Enterprise

lwiechec:
I am not sure if it is possible to 'ask' the GitHub API for the default
upload URL; that would be great but if not, we need to pass it somehow.

refs #266

### DIFF
--- a/src/main/java/org/kohsuke/github/GHRelease.java
+++ b/src/main/java/org/kohsuke/github/GHRelease.java
@@ -124,7 +124,7 @@ public class GHRelease extends GHObject {
      *
      * @param file file to upload
      * @param contentType content type
-     * @param uploadUrlPrefix ULR prefix for uploading; for example, for github.com it's https://uploads.github.com
+     * @param uploadUrlPrefix URL prefix for uploading; for example, for github.com it's https://uploads.github.com
      * @return instance of GHAsset uploaded
      * @since 1.75
      * @throws IOException
--- a/src/main/java/org/kohsuke/github/GHRelease.java
+++ b/src/main/java/org/kohsuke/github/GHRelease.java
@@ -9,6 +9,7 @@ import java.util.Date;
 import java.util.List;
 
 import static java.lang.String.format;
+import static org.apache.commons.lang.ObjectUtils.defaultIfNull;
 
 /**
  * Release in a github repository.
@@ -128,7 +129,7 @@ public class GHRelease extends GHObject {
      *
      * @param file file to upload
      * @param contentType content type
-     * @param uploadUrlPrefix URL prefix for uploading; for example, for github.com it's https://uploads.github.com
+     * @param uploadUrlPrefix URL prefix for uploading; if null {@link #DEFAULT_UPLOAD_URL} will be used
      * @return instance of GHAsset uploaded
      * @since 1.75
      * @throws IOException
@@ -137,7 +138,7 @@ public class GHRelease extends GHObject {
         Requester builder = new Requester(owner.root);
 
         String url = format("%s%s/releases/%d/assets?name=%s",
-                            uploadUrlPrefix, owner.getApiTailUrl(""), getId(), file.getName());
+                            defaultIfNull(uploadUrlPrefix, DEFAULT_UPLOAD_URL), owner.getApiTailUrl(""), getId(), file.getName());
         return builder.contentType(contentType)
                 .with(new FileInputStream(file))
                 .to(url, GHAsset.class).wrap(this);
--- a/src/main/java/org/kohsuke/github/GHRelease.java
+++ b/src/main/java/org/kohsuke/github/GHRelease.java
@@ -19,9 +19,6 @@ import static org.apache.commons.lang.ObjectUtils.defaultIfNull;
  */
 public class GHRelease extends GHObject {
 
-    /** github.com's default upload URL */
-    public final static String DEFAULT_UPLOAD_URL = "https://uploads.github.com";
-
     GitHub root;
     GHRepository owner;
 
--- a/src/main/java/org/kohsuke/github/GHRelease.java
+++ b/src/main/java/org/kohsuke/github/GHRelease.java
@@ -122,6 +122,8 @@ public class GHRelease extends GHObject {
     }
 
     /**
+     * upload the release using upload URL in {@link upload_url} field.
+     *
      * Because github relies on SNI (http://en.wikipedia.org/wiki/Server_Name_Indication) this method will only work on
      * Java 7 or greater.  Options for fixing this for earlier JVMs can be found here
      * http://stackoverflow.com/questions/12361090/server-name-indication-sni-on-java but involve more complicated
@@ -129,33 +131,18 @@ public class GHRelease extends GHObject {
      *
      * @param file file to upload
      * @param contentType content type
-     * @param uploadUrlPrefix URL prefix for uploading; if null {@link #DEFAULT_UPLOAD_URL} will be used
-     * @return instance of GHAsset uploaded
-     * @since 1.75
-     * @throws IOException
-     */
-    public GHAsset uploadAsset(File file, String contentType, String uploadUrlPrefix) throws IOException {
-        Requester builder = new Requester(owner.root);
-
-        String url = format("%s%s/releases/%d/assets?name=%s",
-                            defaultIfNull(uploadUrlPrefix, DEFAULT_UPLOAD_URL), owner.getApiTailUrl(""), getId(), file.getName());
-        return builder.contentType(contentType)
-                .with(new FileInputStream(file))
-                .to(url, GHAsset.class).wrap(this);
-    }
-
-    /**
-     * upload the release to defaul upload URL on github.com
-     *
-     * @param file file to upload
-     * @param contentType content type
      * @return instance of GHAsset uploaded
      * @throws IOException
      */
     public GHAsset uploadAsset(File file, String contentType) throws IOException {
-        return this.uploadAsset(file, contentType, DEFAULT_UPLOAD_URL);
-    }
+        Requester builder = new Requester(owner.root);
 
+        String url = format(upload_url.replace("{?name}", "?name=%s"), file.getName());
+
+        return builder.contentType(contentType)
+                .with(new FileInputStream(file))
+                .to(url, GHAsset.class).wrap(this);
+    }
 
     public List<GHAsset> getAssets() throws IOException {
         Requester builder = new Requester(owner.root);
--- a/src/main/java/org/kohsuke/github/GHRelease.java
+++ b/src/main/java/org/kohsuke/github/GHRelease.java
@@ -125,6 +125,8 @@ public class GHRelease extends GHObject {
      * @param file file to upload
      * @param contentType content type
      * @param uploadUrlPrefix ULR prefix for uploading; for example, for github.com it's https://uploads.github.com
+     * @return instance of GHAsset uploaded
+     * @since 1.75
      * @throws IOException
      */
     public GHAsset uploadAsset(File file, String contentType, String uploadUrlPrefix) throws IOException {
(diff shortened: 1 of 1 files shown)


src/main/java/org/kohsuke/github/GHRelease.java
```diff
@@ -122,18 +122,33 @@ GHRelease wrap(GHRepository owner) {
...
                 .with(new FileInputStream(file))
                 .to(url, GHAsset.class).wrap(this);
     }
 
+    /**
+     * upload the release to defaul upload URL on github.com
+     *
+     * @param file file to upload
+     * @param contentType content type
+     * @throws IOException
```
oleg-nenashev: NIT: better to avoid such Javadoc entries w/o comments due to the Java8 default linter

oleg-nenashev:
:+1: in general

src/main/java/org/kohsuke/github/GHRelease.java
```diff
@@ -122,18 +122,33 @@ GHRelease wrap(GHRepository owner) {
      * http://stackoverflow.com/questions/12361090/server-name-indication-sni-on-java but involve more complicated
      * handling of the HTTP requests to github's API.
      *
+     * @param file file to upload
+     * @param contentType content type
+     * @param uploadUrlPrefix ULR prefix for uploading; for example, for github.com it's https://uploads.github.com
      * @throws IOException
      */
```
oleg-nenashev: Makes sense to add <code>@since</code> in order to let others know which minimal API version provides this method

lwiechec:
Hi @oleg-nenashev , thanks for the input - I have added missing JavaDoc tags and run JavaDoc generation (using Java 8); it does not complain on `GHRelease`.

src/main/java/org/kohsuke/github/GHRelease.java
```diff
@@ -122,18 +122,36 @@ GHRelease wrap(GHRepository owner) {
      * http://stackoverflow.com/questions/12361090/server-name-indication-sni-on-java but involve more complicated
      * handling of the HTTP requests to github's API.
      *
+     * @param file file to upload
+     * @param contentType content type
+     * @param uploadUrlPrefix ULR prefix for uploading; for example, for github.com it's https://uploads.github.com
```
lanwen: ULR -> URL ?

lwiechec:
@lanwen thanks for pointing out the typo - corrected

src/main/java/org/kohsuke/github/GHRelease.java
```diff
@@ -122,18 +122,36 @@ GHRelease wrap(GHRepository owner) {
...
      * @throws IOException
      */
-    public GHAsset uploadAsset(File file, String contentType) throws IOException {
+    public GHAsset uploadAsset(File file, String contentType, String uploadUrlPrefix) throws IOException {
         Requester builder = new Requester(owner.root);
 
-        String url = format("https://uploads.github.com%s/releases/%d/assets?name=%s",
-                owner.getApiTailUrl(""), getId(), file.getName());
+        String url = format("%s%s/releases/%d/assets?name=%s",
+                            uploadUrlPrefix, owner.getApiTailUrl(""), getId(), file.getName());
```
lanwen: can u move default prefix to const and make last arg optional (in case of null arg, use default)?

`defaultIfNull(uploadUrlPrefix, DEFAULT_UPLOAD_URL)`
It can be useful instead of manually define in each place both enterprise and not enterprise urls (for example in github-jenkins-plugin)

src/main/java/org/kohsuke/github/GHRelease.java
```diff
@@ -122,18 +122,36 @@ GHRelease wrap(GHRepository owner) {
...
+    /**
+     * upload the release to defaul upload URL on github.com
+     *
+     * @param file file to upload
+     * @param contentType content type
+     * @return instance of GHAsset uploaded
+     * @throws IOException
+     */
+    public GHAsset uploadAsset(File file, String contentType) throws IOException {
+        return this.uploadAsset(file, contentType, "https://uploads.github.com");
```
lwiechec: @lanwen, you mean something like this ?...

```java
 /**
     * Because github relies on SNI (http://en.wikipedia.org/wiki/Server_Name_Indication) this method will only work on
     * Java 7 or greater.  Options for fixing this for earlier JVMs can be found here
     * http://stackoverflow.com/questions/12361090/server-name-indication-sni-on-java but involve more complicated
     * handling of the HTTP requests to github's API.
     *
     * @param file file to upload
     * @param contentType content type
     * @param uploadUrlPrefix URL prefix for uploading; for example, for github.com it's https://uploads.github.com
     * @return instance of GHAsset uploaded
     * @since 1.75
     * @throws IOException
     */
    public GHAsset uploadAsset(File file, String contentType, String uploadUrlPrefix) throws IOException {
        Requester builder = new Requester(owner.root);

        String uploadUrlPrefix_ = uploadUrlPrefix != null ? uploadUrlPrefix : DEFAULT_UPLOAD_URL;

        String url = format("%s%s/releases/%d/assets?name=%s",
                            uploadUrlPrefix_, owner.getApiTailUrl(""), getId(), file.getName());
        return builder.contentType(contentType)
                .with(new FileInputStream(file))
                .to(url, GHAsset.class).wrap(this);
    }

    /**
     * upload the release to defaul upload URL on github.com
     *
     * @param file file to upload
     * @param contentType content type
     * @return instance of GHAsset uploaded
     * @throws IOException
     */
    public GHAsset uploadAsset(File file, String contentType) throws IOException {
        return this.uploadAsset(file, contentType, null);
    }
```

src/main/java/org/kohsuke/github/GHRelease.java
```diff
@@ -122,18 +126,36 @@ GHRelease wrap(GHRepository owner) {
...
 
+    /**
+     * upload the release to defaul upload URL on github.com
+     *
+     * @param file file to upload
+     * @param contentType content type
+     * @return instance of GHAsset uploaded
+     * @throws IOException
+     */
+    public GHAsset uploadAsset(File file, String contentType) throws IOException {
```
lwiechec: @lanwen hi, you mean something like that?...

```java
    /**
     * Because github relies on SNI (http://en.wikipedia.org/wiki/Server_Name_Indication) this method will only work on
     * Java 7 or greater.  Options for fixing this for earlier JVMs can be found here
     * http://stackoverflow.com/questions/12361090/server-name-indication-sni-on-java but involve more complicated
     * handling of the HTTP requests to github's API.
     *
     * @param file file to upload
     * @param contentType content type
     * @param uploadUrlPrefix URL prefix for uploading; for example, for github.com it's https://uploads.github.com
     * @return instance of GHAsset uploaded
     * @since 1.75
     * @throws IOException
     */
    public GHAsset uploadAsset(File file, String contentType, String uploadUrlPrefix) throws IOException {
        Requester builder = new Requester(owner.root);

        String uploadUrlPrefix_ = uploadUrlPrefix != null ? uploadUrlPrefix : DEFAULT_UPLOAD_URL;

        String url = format("%s%s/releases/%d/assets?name=%s",
                            uploadUrlPrefix_, owner.getApiTailUrl(""), getId(), file.getName());
        return builder.contentType(contentType)
                .with(new FileInputStream(file))
                .to(url, GHAsset.class).wrap(this);
    }

    /**
     * upload the release to defaul upload URL on github.com
     *
     * @param file file to upload
     * @param contentType content type
     * @return instance of GHAsset uploaded
     * @throws IOException
     */
    public GHAsset uploadAsset(File file, String contentType) throws IOException {
        return this.uploadAsset(file, contentType, null);
    }
```

lanwen: yep!
You can try `defaultIfEmpty()` from commons-lang (if available) instead of `uploadUrlPrefix != null ? uploadUrlPrefix : DEFAULT_UPLOAD_URL;`

and call `return this.uploadAsset(file, contentType, DEFAULT_UPLOAD_URL);`

also please reflect in javadoc, that third param may be null and default value will be used instead

lwiechec: ok, done and pushed. Thanks!





---- On Ter, 05 Abr 2016 16:25:58 +0200 Merkushev Kirill &lt;notifications@github.com&gt;wrote ---- 




In src/main/java/org/kohsuke/github/GHRelease.java:

&gt; return builder.contentType(contentType) &gt; .with(new FileInputStream(file)) &gt; .to(url, GHAsset.class).wrap(this); &gt; } &gt; &gt; + /** &gt; + * upload the release to defaul upload URL on github.com &gt; + * &gt; + * @param file file to upload &gt; + * @param contentType content type &gt; + * @return instance of GHAsset uploaded &gt; + * @throws IOException &gt; + */ &gt; + public GHAsset uploadAsset(File file, String contentType) throws IOException { 
yep!

 You can try defaultIfEmpty() from commons-lang (if available) instead of uploadUrlPrefix != null ? uploadUrlPrefix : DEFAULT_UPLOAD_URL;


and call return this.uploadAsset(file, contentType, DEFAULT_UPLOAD_URL);

also please reflect in javadoc, that third param may be null and default value will be used instead

lanwen:
LGTM

lwiechec:
@lanwen @oleg-nenashev can we merge this one?...

charleswisniewski:
@lanwen @oleg-nenashev can this get a merge please? Would love to use this.

Thanks!

kohsuke:
I'm afraid I have a problem with this change.

The URL that represents the asset upload endpoint shouldn't be passed as a parameter like this. Ideally, the library should be able to infer that on its own from `GitHub.apiUrl` or through some HTTP header / navigable links provided by GitHub.

Worst case, it should be a parameter to `GitHubBuilder` that we store separately in the `GitHub` class.

If neither of that is possible, then it must be that GitHub Enterprise uses some implicit naming convention to go from API URL to the asset upload endpoint. What would help me bring this change to completion is to see what that implicit convention is. I don't have a ready access to GHE, so I cannot do that on my own. Put shortly, I need to know "if GHE is installed on https://ghe.acme.org/ then the asset upload URL is `https://xxx.yyy.zzz/www/repos/foo/bar/releases/zot/assets?name=abc`.

src/main/java/org/kohsuke/github/GHRelease.java
```diff
@@ -122,41 +122,28 @@ GHRelease wrap(GHRepository owner) {
...
-     * @since 1.75
      * @throws IOException
      */
-    public GHAsset uploadAsset(File file, String contentType, String uploadUrlPrefix) throws IOException {
+    public GHAsset uploadAsset(File file, String contentType) throws IOException {
         Requester builder = new Requester(owner.root);
 
-        String url = format("%s%s/releases/%d/assets?name=%s",
-                            defaultIfNull(uploadUrlPrefix, DEFAULT_UPLOAD_URL), owner.getApiTailUrl(""), getId(), file.getName());
+        String url = format(upload_url.replace("{?name}", "?name=%s"), file.getName());
```
lwiechec: @kohsuke we can add some more tests here, for example to check if it's not null - if needed?

lwiechec:
@kohsuke thanks for looking on my patch :) sorry for delay, I was busy with something else but I wanted to finish that.

You are right, the less needed parameters from the user who wants to upload the release, the better.

could you please check  408832e ?... looks easiest to me...

src/main/java/org/kohsuke/github/GHRelease.java
```diff
@@ -9,6 +9,7 @@
 import java.util.List;
 
 import static java.lang.String.format;
+import static org.apache.commons.lang.ObjectUtils.defaultIfNull;
```
halkeye: It doesn't look like import is actually used anymore

src/main/java/org/kohsuke/github/GHRelease.java
```diff
@@ -17,6 +18,7 @@
  * @see GHRepository#createRelease(String)
  */
 public class GHRelease extends GHObject {
+
```
halkeye: totally nitpicking, but an extra newline was added here

kohsuke:
I lost context of this change, and this has been open for too long. I'm marking this as closed.

josephakroush:
@kohsuke Can we get this PR re-opened and merged? I need this functionality.

atkawa7:
@kohsuke @halkeye PING

PR #270, opened 2016-04-04
[#269] Add reopen method on GHMilestone

### LINKED ISSUE #269: (re)open method on GHMilestone
In some cases (e.g. restoring state for functional tests with sanboxed GitHub project) it could be needed to reopen already closed milestone. Unfortunately currently there is only a `close()` method on `GHMilestone` and as a workaround a private method `edit()` has to be used.

szpak:
Fixes #269.


szpak:
I'm not sure about the method name. `Open` might suggest create a new milestone (like in "open a new issue"). On the other hand `reopen` is less consistent.

PR #272, opened 2016-04-14
Added support for the extended stargazers API in Github V3 API

noctarius:
Github has added starred_at date to stargazers when stargazers are requested with a special (new) version on the REST API. Added support for the version and a type representing the additional information.

### DIFF
--- a/src/main/java/org/kohsuke/github/GHRepository.java
+++ b/src/main/java/org/kohsuke/github/GHRepository.java
@@ -935,10 +935,34 @@ public class GHRepository extends GHObject {
     }
 
     /**
-     * Lists all the users who have starred this repo.
+     * Lists all the users who have starred this repo based on the old version of the API. For
+     * additional information, like date when the repository was starred, see {@link #listExtendedStargazers()}
      */
     public PagedIterable<GHUser> listStargazers() {
         return listUsers("stargazers");
+    }
+
+    /**
+     * Lists all the users who have starred this repo based on new version of the API, having extended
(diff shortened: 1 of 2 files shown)


src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -937,8 +937,22 @@ public GHLabel createLabel(String name, String color) throws IOException {
...
      * Lists all the users who have starred this repo.
      */
-    public PagedIterable<GHUser> listStargazers() {
-        return listUsers("stargazers");
+    public PagedIterable<GHStargazer> listStargazers() {
+        return new PagedIterable<GHStargazer>() {
+            @Override
+            public PagedIterator<GHStargazer> _iterator(int pageSize) {
+                Requester requester = root.retrieve();
+                requester.setHeader("Accept", "application/vnd.github.v3.star+json");
```
oleg-nenashev: Will it continue working with previous APIs? E.g. on old GitHub Enterprise instances

noctarius:
Good question, maybe it should have both versions supported

src/main/java/org/kohsuke/github/GHRepository.java
noctarius: Good question, maybe it should have both versions supported

noctarius: Added back the old version for compatibility and also listExtendedStargazers to return the new version. In addition added Javadoc to link both methods :)

PR #276, opened 2016-04-29
Add support to delete a team

thug-gamer:
Add a method to delete a team.

@kohsuke @oleg-nenashev Could you please review this pull request?

### DIFF
--- a/src/main/java/org/kohsuke/github/GHTeam.java
+++ b/src/main/java/org/kohsuke/github/GHTeam.java
@@ -126,6 +126,13 @@ public class GHTeam {
     public void remove(GHRepository r) throws IOException {
         org.root.retrieve().method("DELETE").to(api("/repos/" + r.getOwnerName() + '/' + r.getName()), null);
     }
+    
+    /**
+     * Deletes this team.
+     */
+    public void delete() throws IOException {
+        org.root.retrieve().method("DELETE").to(api(""));
+    }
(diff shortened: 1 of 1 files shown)


src/main/java/org/kohsuke/github/GHTeam.java
```diff
@@ -126,6 +126,13 @@ public void add(GHRepository r) throws IOException {
     public void remove(GHRepository r) throws IOException {
         org.root.retrieve().method("DELETE").to(api("/repos/" + r.getOwnerName() + '/' + r.getName()), null);
     }
+    
+    /**
+     * Deletes this team.
+     */
+    public void delete() throws IOException {
+        org.root.retrieve().method("DELETE").to(api(""));
```
recena: Empty URL for the API method?

thug-gamer: According to [GitHub API](https://developer.github.com/v3/orgs/teams/#delete-team), the url is `/teams/:id`, which is what `api` method creates

samrocketman: Yep, https://github.com/thug-gamer/github-api/blob/007378c3a6f4d8ea87abd1dfd6a7f8cfc9feac80/src/main/java/org/kohsuke/github/GHTeam.java#L137-L139

PR #281, opened 2016-05-14
Add Slug to GHTeam per v3 API: https://developer.github.com/v3/orgs/t…

apemberton:
https://developer.github.com/v3/orgs/teams/ has a slug property for teams. This property is similar to the login property for users and organizations - a URL path friendly tokenized string (as opposed to free text in the name field). 

Slug is better suited for use in automated / scripting cases when dealing with the Team API.

Admittedly, I wrote a test but could not test as the project's test harness requires super powers which I do not possess.

@reviewbybees

### DIFF
--- a/src/main/java/org/kohsuke/github/GHOrganization.java
+++ b/src/main/java/org/kohsuke/github/GHOrganization.java
@@ -94,6 +94,17 @@ public class GHOrganization extends GHPerson {
     }
 
     /**
+     * Finds a team that has the given slug in its {@link GHTeam#getSlug()}
+     */
+    public GHTeam getTeamBySlug(String slug) throws IOException {
+        for (GHTeam t : listTeams()) {
+            if(t.getSlug().equals(slug))
+                return t;
+        }
+        return null;
+    }
+
+    /**
      * Checks if this organization has the specified user as a member.
      */
     public boolean hasMember(GHUser user) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/kohsuke/github/GHOrganization.java
```diff
@@ -94,6 +94,17 @@ public GHTeam getTeamByName(String name) throws IOException {
     }
 
     /**
+     * Finds a team that has the given slug in its {@link GHTeam#getSlug()}
+     */
+    public GHTeam getTeamBySlug(String slug) throws IOException {
+        for (GHTeam t : listTeams()) {
+            if(t.getSlug().equals(slug))
```
jglick: Irritating that the [single-team API](https://developer.github.com/v3/orgs/teams/#get-team) seems to require a numeric ID rather than the “slug”.

apemberton:
So @jglick was that a +1?

jglick:
Yeah :bee: insofar as I understand it.

src/main/java/org/kohsuke/github/GHOrganization.java
samrocketman: I just opened a feature request with GitHub support to allow selecting teams via `:slug` key in addition to the `:id` key within the API.

aheritier:
:bee:

apemberton:
Nudge. Merge please.

src/main/java/org/kohsuke/github/GHOrganization.java
samrocketman: They accepted the enhancement for selecting teams via `:slug` but with the usually "we make no guarantees" line.  So we may or may not see it as a feature of the API in the future.

apemberton:
Bump.

apemberton:
Nudge.

PR #283, opened 2016-05-16
Add some level of synchronization to the root of the API

mmitche:
This adds some synchronization to the maps at the root of the API to avoid duplicated calls to the actual GH REST API.  Specifically this is targeted around the two maps, orgs and users.  This fix makes the GHPRB jenkins plugin behave much better when there are lots of projects that could build for a specific repo (even if only a few are actually triggered)


kohsuke:
OTOH this change introduces contention when multiple threads try to access different users and orgs. I wonder if it'd be worthwhile to take the fix one step further and address that.

mmitche:
Wouldn't that still cause problems?  We should avoid concurrent modification of the maps.  What are you thinking?

I'm also thinking that this is an incomplete fix.  There are a lot of non-final fields around the API that could be modified by multiple threads if someone passed...say...a pull request object to multiple threads and made multiple concurrent calls to certain APIs.  However, the maps at the root appear to be the primary source of concurrency issues.

src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -276,56 +275,92 @@ public GHRateLimit getRateLimit() throws IOException {
...
         requireCredential();
 
-        GHMyself u = retrieve().to("/user", GHMyself.class);
+        // This entire block is under synchronization to avoid the relatively common case
+        // where a bunch of threads try to enter this code simultaneously.  While we could
+        // scope the synchronization separately around the map retrieval and update (or use a concurrent hash)
+        // map, the point is to avoid making unnecessary GH API calls, which are expensive from
+        // an API rate standpoint
+        synchronized (users) {
+            GHUser cachedUser = users.get(this.login);
```
mmitche: @kohsuke This is invalid right now (we call getMyself in the constructor).  I'm fixing this to check for a valid login, but in the meantime don't merge this.

src/main/java/org/kohsuke/github/GHPullRequest.java
```diff
@@ -200,7 +200,7 @@ public String getMergeCommitSha() throws IOException {
      * Depending on the original API call where this object is created, it may not contain everything.
      */
     private void populate() throws IOException {
-        if (merged_by!=null)    return; // already populated
+        if (mergeable_state != null)    return; // already populated by id
```
mmitche: merged_by may be null if the PR was unmerged, leading to an API call every time we call populate()

src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -77,9 +77,10 @@
      */
     /*package*/ final String encodedAuthorization;
 
-    private final Map<String,GHUser> users = new Hashtable<String, GHUser>();
-    private final Map<String,GHOrganization> orgs = new Hashtable<String, GHOrganization>();
-
+    private final Map<String,GHUser> users;
+    private final Map<String,GHOrganization> orgs;
+    // Cache of myself object.
+    private GHMyself myself;
```
mmitche: We now cache GHMySelf differently than the other users.  This is vs. keeping it in the user's map.  The issue is that if the first call to the user's map was for 'foouser' as a GHUser, then subsequent calls to getMsyelf() (if logged in as foouser) would always call the REST API because the type returned from the map would not be the desired type.  While theoretically we could just update the map with the more specific GHMyself (since it's a subclass of GHUser), I thought it better for clarity to simply cache it separately.

mmitche:
@kohsuke I added a bunch of logging to the API and made a bunch of additional fixes that should improve overall performance (few REST calls).  PTAL.

mmitche:
@kohsuke Have you had a chance to look at the updated version here?

This reduced our GH API calls by around 50% in a large jenkins installation

akostadinov:
In a small install we are hitting the rate_limit issue. Any update on this pull request?

mmitche:
@kohsuke Addressed issues.  Changed over to using a ConcurrentHashMap, which has fast reads (the common case) and only locks on write.

PR #284, opened 2016-06-19
Addition of License API elements

dedickinson:
I've added in functionality to provide license information from https://developer.github.com/v3/licenses/:

- Get list of GitHub licenses using `GitHub#listLicenses()`
- Get the details for a specific license using `GitHub#getLicense(String key)`
- For a `GHRepository`:
  - The basic license details are loaded with the repo information (`GHRepository#getLicense`)
  - `GHRepository#getFullLicense` will get the full license information (ala `GitHub#getLicense(String key)`)
  - `GHRepository#getLicenseContent()` will get the license content information for the repository

As this uses a preview API, access to the information requires the use of `application/vnd.github.drax-preview+json` in the HTTP connection's ACCEPT header. This is provided in `PreviewHttpConnector.java`.

All existing functionality appears to continue working with the normal connector - `PreviewHttpConnector` just opens up the extra data from the Licenses API.


src/main/java/org/kohsuke/github/extras/PreviewHttpConnector.java
```diff
@@ -0,0 +1,67 @@
...
+                return (HttpURLConnection) url.openConnection();
+            }
+        }, ImpatientHttpConnector.CONNECT_TIMEOUT, ImpatientHttpConnector.READ_TIMEOUT);
+    }
+
+    public HttpURLConnection connect(URL url) throws IOException {
+        HttpURLConnection con = base.connect(url);
+        con.setConnectTimeout(connectTimeout);
+        con.setReadTimeout(readTimeout);
+        con.addRequestProperty("Accept", PREVIEW_MEDIA_TYPE);
```
dedickinson: This is the primary difference from the default connector and allows access to the preview API

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -866,26 +849,62 @@ public GHCommitStatus getLastCommitStatus(String sha1) throws IOException {
     }
```
dedickinson: Lines 851 - 889 are the primary additions to the class: the methods `getLicense`, `getFullLicense` and `getLicenseContent`

src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -339,6 +340,33 @@ public GHRepository getRepository(String name) throws IOException {
...
+     * Returns a list of popular open source licenses
+     *
+     * WARNING: This uses a PREVIEW API - you must use {@link org.kohsuke.github.extras.PreviewHttpConnector}
+     *
+     * @see <a href="https://developer.github.com/v3/licenses/">GitHub API - Licenses</a>
+     *
+     * @return a list of popular open source licenses
+     * @throws IOException if the HttpConnector doesn't pass in the preview header or other IO issue
+     */
+    public List<GHLicenseBase> listLicenses() throws IOException {
```
dedickinson: IOException is thrown if the preview API header is not used

kohsuke:
I've incorporated the change into 1.77, but had to squash the change to avoid whitespace changes to `GHRepository`. I hope you won't mind.

PR #289, opened 2016-07-21
Implement an abuse handler

### LINKED ISSUE #285: Excessive concurrent request rate limit not handled
If the hourly rate limit is not hit, but the API abuse mechanism is triggered, the API won't handle this gracefully.  Instead, it bails.  I think the ideal case here would be to wait for a small random amount of time (< 5 seconds perhaps, with some backoff) and retry

mmitche:
If too many requests are made within X amount of time (not the traditional hourly rate limit), github may begin returning 403.  Then we should wait for a bit to attempt to access the API again.  In this case, we parse out the Retry-After field returned and sleep until that (it's usually 60 seconds)

Fixes #285

### DIFF
--- a/src/main/java/org/kohsuke/github/Requester.java
+++ b/src/main/java/org/kohsuke/github/Requester.java
@@ -572,6 +572,14 @@ class Requester {
             root.rateLimitHandler.onError(e,uc);
             return;
         }
+        
+        // Check to see whether we hit a 403, and the Retry-After field is
+        // available.
+        if (responseCode == HttpURLConnection.HTTP_FORBIDDEN && 
+            uc.getHeaderField("Retry-After") != null) {
+            this.root.abuseLimitHandler.onError(e,uc);
+            return;
+        }
 
         InputStream es = wrapStream(uc.getErrorStream());
         try {
--- a/src/main/java/org/kohsuke/github/GitHub.java
+++ b/src/main/java/org/kohsuke/github/GitHub.java
@@ -83,6 +83,7 @@ public class GitHub {
     private final String apiUrl;
 
     /*package*/ final RateLimitHandler rateLimitHandler;
+    /*package*/ final RateLimitHandler abuseLimitHandler;
 
     private HttpConnector connector = HttpConnector.DEFAULT;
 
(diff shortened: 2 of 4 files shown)


src/main/java/org/kohsuke/github/Requester.java
```diff
@@ -576,11 +576,20 @@ private InputStream wrapStream(InputStream in) throws IOException {
         InputStream es = wrapStream(uc.getErrorStream());
         try {
             if (es!=null) {
+                String errorString = IOUtils.toString(es, "UTF-8");
+                // Check to see whether we hit a 403, and the message indicates the
+                // abuse handler (occurs on too many concurrent requests)
+                if (responseCode == HttpURLConnection.HTTP_FORBIDDEN && 
+                    errorString.contains("You have triggered an abuse detection mechanism.")) {
```
mmitche: @kohsuke This is the only bit I'm not totally sure about.  The best bit would be to just be able to check for 403, but I'm not sure whether 403 happens in more circumstances than just API abuse.  There wasn't anything else in the header indicating abuse (though maybe I could just check for Retry-After)

Additionally, I imagine that there isn't a way to get a localized error string here, but if there was the check would be broken.  Thoughts?

Shredder121: My 2 cents:
I'd just check for status code and retry-after, those are the ones that are documented.
The human readable hint/text isn't quite documented.

mmitche: Okay, yeah I think you're right.

mmitche: Fixed

src/main/java/org/kohsuke/github/Requester.java
```diff
@@ -576,6 +576,15 @@ private InputStream wrapStream(InputStream in) throws IOException {
         InputStream es = wrapStream(uc.getErrorStream());
         try {
             if (es!=null) {
+                String errorString = IOUtils.toString(es, "UTF-8");
```
Shredder121:
```
Dead store to errorString in org.kohsuke.github.Requester.handleApiError(IOException) [org.kohsuke.github.Requester] At Requester.java:[line 579]
```

mmitche:
Fixed and moved some code around.

PR #295, opened 2016-09-19
Use maximum permitted page size

jglick:
Pending GraphQL, performance is going to be bad, but do what we can.

@reviewbybees

### DIFF
--- a/src/main/java/org/kohsuke/github/GHPerson.java
+++ b/src/main/java/org/kohsuke/github/GHPerson.java
@@ -52,7 +52,7 @@ public abstract class GHPerson extends GHObject {
      */
     public synchronized Map<String,GHRepository> getRepositories() throws IOException {
         Map<String,GHRepository> repositories = new TreeMap<String, GHRepository>();
-        for (GHRepository r : listRepositories()) {
+        for (GHRepository r : listRepositories(100)) {
             repositories.put(r.getName(),r);
         }
         return Collections.unmodifiableMap(repositories);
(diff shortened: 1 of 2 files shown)


src/main/java/org/kohsuke/github/GHPerson.java
```diff
@@ -52,7 +52,7 @@ protected synchronized void populate() throws IOException {
      */
     public synchronized Map<String,GHRepository> getRepositories() throws IOException {
         Map<String,GHRepository> repositories = new TreeMap<String, GHRepository>();
-        for (GHRepository r : listRepositories()) {
+        for (GHRepository r : listRepositories(100)) {
```
svanoort: 🐜  Probably should amend the comment on line 75

svanoort:
🐝  with ant.

src/main/java/org/kohsuke/github/GHPerson.java
jglick: That would be for `listRepositories(int pageSize)`:

> Unlike `getRepositories()`, this does not wait until all the repositories are returned.

So what do you think needs to be amended? `listRepositories` overloads produce incremental results, whereas `getRepositories` waits for the full list. I am not changing those semantics (that I know of!); I am just asking for larger batches, on the grounds that an organization with 300 repositories is better enumerated using 3 requests returning 100 entries each, rather than 10 requests returning 30 entries each.

svanoort: Nevermind -- I misread something while running through it.  

Bonus of increasing request size: less likely to hit rate limits, probably.

jglick:
@reviewbybees done

jglick:
@recena also suggests making 100 ([the maximum](https://developer.github.com/v3/#pagination)) be the default size to begin with. I am not sure if the lower default is intentional for some reason.

KostyaSha:
Use fibonacci?

abayer:
:bee:

PR #300, opened 2016-10-24
Expose the commit dates

stephenc:
I need this for SCM API stuff

### DIFF
--- a/src/main/java/org/kohsuke/github/GHCommit.java
+++ b/src/main/java/org/kohsuke/github/GHCommit.java
@@ -2,12 +2,12 @@ package org.kohsuke.github;
 
 import com.infradna.tool.bridge_method_injector.WithBridgeMethods;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
-
 import java.io.IOException;
 import java.net.URL;
 import java.util.AbstractList;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Date;
 import java.util.List;
 
 /**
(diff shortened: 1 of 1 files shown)


src/main/java/org/kohsuke/github/GHCommit.java
```diff
@@ -42,11 +42,19 @@ public GitUser getAuthor() {
             return author;
         }
 
+        public Date getAuthoredDate() {
+            return GitHub.parseDate(author.date);
```
KostyaSha: Why `authored`? It's `AuthorDate`

src/main/java/org/kohsuke/github/GHCommit.java
```diff
@@ -42,11 +42,19 @@ public GitUser getAuthor() {
...
+        }
+
         @WithBridgeMethods(value = GHAuthor.class, castRequired = true)
         public GitUser getCommitter() {
             return committer;
         }
 
+        public Date getCommitDate() {
+            return GitHub.parseDate(author.date);
```
KostyaSha: So what is the difference of this two methods? Copy-paste instead `commit.date`?

KostyaSha:
@kohsuke did you review it or blindly merged and released?

KostyaSha:
@stephenc ?

kohsuke:
Good catch! I did look at the change but I missed it.

I'll fix this and push out a new release.

kohsuke:
I'll leave the `getAuthoredDate` as is. I trust Stephen's English than mine and yours.

stephenc:
Good catch on the copy and paste.

So I went with "commitDate" rather than "committerDate" and "authoredDate" vs "authorDate" as these are the dates of the respective actions. In a sense GitHub conflates the action with the person, and it just sounded wrong to my ear to say "getCommitterDate"... what's the committer date... does that method arrange for flowers a movie and a meal with the committer... ok, perhaps the committer deserves that as a reward for their work... that is more the date of the commit... closest analogy for author I could find was authored

KostyaSha:
Question not in english, it not a novel. When you know API you can use library easily, otherwise you should scroll code/methods to get idea what they return. Also why you can't call getter().field yourself and need this proxy method?

PR #301, opened 2016-10-25
Add stargarzer field to GHRepository

goneall:
Signed-off-by: Gary O'Neall gary@sourceauditor.com

Pull request to add the stargazers field to the HGRepository object.

Let me know if you have any questions on the request.

Thanks,
Gary


kohsuke:
I merged #304 that achieved the same result. So this change is effectively in the master.

PR #306, opened 2016-11-08
Add offline support to the API to make parsing events easier

stephenc:
- When we receive events from a webhook, it is non-trivial to determine which GitHub instance the event came from
  or for that matter even if the event actually came from GitHub or GitHub Enterprise.
- In order to ensure that the logic for parsing events does not get replicated in clients, we need to be
  able to call `GitHub.parseEventPayload(Reader,Class)` without knowing which `GitHub` the event originates from
  and without the resulting objects triggering API calls back to a GitHub
- Thus we add `GitHub.offline()` to provide an off-line connection
- Thus we modify some of the object classes to return best-effort objects when off-line
- Add support for more of the event types into `GHEventPayload`
- Add tests of the event payload and accessing critical fields when using `GitHub.offline()`

@reviewbybees


src/main/java/org/kohsuke/github/GHIssueComment.java
```diff
@@ -35,7 +35,7 @@
     GHIssue owner;
 
     private String body, gravatar_id;
-    private GHUser user;
+    private GHUser user; // not fully populated. beware.
```
jglick: Seems like this sort of comment would better be placed in the Javadoc of the getter.

src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -216,13 +216,39 @@ public static GitHub connectToEnterpriseAnonymously(String apiUrl) throws IOExce
     }
 
     /**
+     * An off-line only {@link GitHub} useful for parsing event notification from an unknown source.
```
jglick: `offline-only`

src/test/java/org/kohsuke/github/PayloadRule.java
```diff
@@ -0,0 +1,87 @@
...
+        return stream;
+    }
+
+    public byte[] asBytes() throws IOException {
+        InputStream input = asInputStream();
+        try {
+            return IOUtils.toByteArray(input);
+        } finally {
+            IOUtils.closeQuietly(input);
+        }
```
jglick:
```java
try (InputStream input = asInputStream()) {
    return IOUtils.toByteArray(input);
}
```

src/test/java/org/kohsuke/github/PayloadRule.java
```diff
@@ -0,0 +1,87 @@
...
+    public String asString(String encoding) throws IOException {
+        return new String(asBytes(), encoding);
+    }
+
+    public String asString() throws IOException {
+        return new String(asBytes(), Charset.defaultCharset().name());
+    }
+
+    public Reader asReader() throws FileNotFoundException {
+        return new InputStreamReader(asInputStream(), Charset.defaultCharset());
```
jglick: Since this is being used for JSON, I would think UTF-8 would be more appropriate.

src/test/resources/org/kohsuke/github/GHEventPayloadTest/delete.json
```diff
@@ -0,0 +1,111 @@
...
+    "watchers": 0,
+    "default_branch": "master"
+  },
+  "sender": {
+    "login": "baxterthehacker",
+    "id": 6752317,
+    "avatar_url": "https://avatars.githubusercontent.com/u/6752317?v=3",
+    "gravatar_id": "",
+    "url": "https://api.github.com/users/baxterthehacker",
+    "html_url": "https://github.com/baxterthehacker",
```
jglick: Who are you really @baxterthehacker? :-)

src/test/java/org/kohsuke/github/PayloadRule.java
```diff
@@ -0,0 +1,87 @@
...
+import java.nio.charset.Charset;
+import org.apache.commons.io.IOUtils;
+import org.junit.rules.TestRule;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+/**
+ * @author Stephen Connolly
+ */
+public class PayloadRule implements TestRule {
```
jglick: Not going to use https://github.com/jenkinsci/jenkins-test-harness/pull/39?

src/main/java/org/kohsuke/github/GHIssueComment.java
stephenc: yep... just copying existing code style

src/test/java/org/kohsuke/github/PayloadRule.java
stephenc: This is a kohsuke project not a jenkins project (would be nice if it was something we could release without KK)

src/test/java/org/kohsuke/github/PayloadRule.java
stephenc: Java 5 (yes WAT!)

src/test/resources/org/kohsuke/github/GHEventPayloadTest/delete.json
stephenc: These are the samples from github's api docs

src/main/java/org/kohsuke/github/GHEventPayload.java
```diff
@@ -121,16 +140,339 @@ void wrapUp(GitHub root) {
     }
 
     /**
+     * A comment was added to a commit
+     *
+     * @see <a href="http://developer.github.com/v3/activity/events/types/#commitcommentevent">authoritative source</a>
+     */
+    @SuppressFBWarnings(value = {"UWF_FIELD_NOT_INITIALIZED_IN_CONSTRUCTOR", "NP_UNWRITTEN_FIELD" },
+            justification = "Constructed by JSON deserialization")
+    public static class CommitComment extends GHEventPayload {
```
lanwen: maybe its better to place this classes in separate files?

src/test/java/org/kohsuke/github/GHEventPayloadTest.java
```diff
@@ -0,0 +1,248 @@
...
+        assertThat(event.getRepository().getName(), is("public-repo"));
+        assertThat(event.getRepository().getOwner().getLogin(), is("baxterthehacker"));
+        assertThat(event.getSender().getLogin(), is("baxterthehacker"));
+    }
+
+    @Test
+    public void pull_request() throws Exception {
+        GHEventPayload.PullRequest event =
+                GitHub.offline().parseEventPayload(payload.asReader(), GHEventPayload.PullRequest.class);
+        assertThat(event.getAction(), is("opened"));
```
lanwen: you can create expected object and use `haveSameProperties` matcher

stephenc: That runs the risk of following requests for the partial objects that are not available from the event objects. safer to just test the paths

src/main/java/org/kohsuke/github/GHEventPayload.java
stephenc: it would have been better if the design was originally that way... but needs to be in same package due to current API design (FTR I would not have designed this API this way as it is very tricky to unpick the stuff that requires a connection from the stuff that doesn't... but we are where we are)

src/main/java/org/kohsuke/github/GHEventPayload.java
```diff
@@ -147,6 +489,11 @@ public String getBefore() {
             return before;
         }
 
+        @JsonSetter // alias
+        private void setAfter(String after) {
```
oleg-nenashev: Maybe it worth to add getter as well. Just to avoid confusing people

stephenc: It's `private` just using this as an alias field name... the documented sample event payload says "after" the documented event says "head" and my tests say "head" but the alias is here as a "just in case"

PR #324, opened 2016-12-16
[JENKINS-36240] Added GHRepository.getPermission(String)

jglick:
[JENKINS-36240](https://issues.jenkins-ci.org/browse/JENKINS-36240)

Using a new API (currently still in preview) that fixes a longstanding issue.

@reviewbybees

### DIFF
--- a/src/main/java/org/kohsuke/github/GHRepository.java
+++ b/src/main/java/org/kohsuke/github/GHRepository.java
@@ -481,6 +481,18 @@ public class GHRepository extends GHObject {
     }
 
     /**
+     * Obtain permission for a given user in this repository.
+     * @param user a {@link GHUser#getLogin}
+     * @throws FileNotFoundException under some conditions (e.g., private repo you can see but are not an admin of); treat as unknown
+     * @throws HttpException with a 403 under other conditions (e.g., public repo you have no special rights to); treat as unknown
+     */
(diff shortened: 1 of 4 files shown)


src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -481,6 +481,18 @@ public boolean hasAssignee(GHUser u) throws IOException {
     }
 
     /**
+     * Obtain permission for a given user in this repository.
+     * @param user a {@link GHUser#getLogin}
+     * @throws FileNotFoundException under some conditions (e.g., private repo you can see but are not an admin of); treat as unknown
+     * @throws HttpException with a 403 under other conditions (e.g., public repo you have no special rights to); treat as unknown
+     */
+    public GHPermission getPermission(String user) throws IOException {
```
kohsuke: What if we just have this method return an enum of admin/read/write/none?

kohsuke: And what if the user parameter is `GHUser` and not `String'?

jglick:
Apparently amended in 6bfeb54f3c5116e241ad8ddd130667cd398bf459, d1c900a6206ddbeb78616a92841b0667a08a8a11, and 911e8d21a717842221f42cd49de0dd543250a807, and then released in 1.83.

PR #327, opened 2017-01-05
Expose Rate Limit Headers

stephenc:
Exposes the rate limit header responses so that consumers of the API can proactively tune their usage.

@reviewbybees

### DIFF
--- a/src/main/java/org/kohsuke/github/GitHub.java
+++ b/src/main/java/org/kohsuke/github/GitHub.java
@@ -23,12 +23,10 @@
  */
 package org.kohsuke.github;
 
-import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.ANY;
-import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.NONE;
-import static java.util.logging.Level.FINE;
-import static java.net.HttpURLConnection.HTTP_UNAUTHORIZED;
-import static org.kohsuke.github.Previews.DRAX;
-
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.introspect.VisibilityChecker.Std;
+import com.infradna.tool.bridge_method_injector.WithBridgeMethods;
 import java.io.ByteArrayInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


KostyaSha:
I guess you far away from good gh integration, so soon you will get https://github.com/kohsuke/github-api/issues/303

stephenc:
@kohsuke ping

src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -287,17 +295,61 @@ public void setConnector(HttpConnector connector) {
...
             // GitHub Enterprise doesn't have the rate limit, so in that case
             // return some big number that's not too big.
             // see issue #78
             GHRateLimit r = new GHRateLimit();
             r.limit = r.remaining = 1000000;
-            long hours = 1000L * 60 * 60;
-            r.reset = new Date(System.currentTimeMillis() + 1 * hours );
-            return r;
+            long hour = 60L * 60L; // this is madness, storing the date as seconds in a Date object
+            r.reset = new Date((System.currentTimeMillis() + hour) / 1000L );
```
xhumanoid: I am sorry, but what the crazy calculation do you do here?

hour - value in second
System.currentTimeMillis() - return millisecond

(millisecond + second)/ 1000L - strange unpredictable type of second, we can say this type of seconds as X.

By javadoc:
public Date(long date)
date   the milliseconds since January 1, 1970, 00:00:00 GMT.

if we will use new Date( X ), then we can't predict what type of date we will receive.

stephenc: Ha! Read the comment in GHRateLimit... for crazy reasons that class stores the date in Date with seconds expressed as milliseconds.

Without this fixup the reset date will be incorrect as returned by `GHRateLimit.getResetDate()`

I have comments indicating that it be madness... but actually is the fix if least harm

stephenc: Ack the adding of an hours seconds should be outside of the division, but this will still give a result in the future for GHE and as GHE does not have a rate limit...

xhumanoid: I read about "this is madness, storing the date as seconds in a Date object"
but very confused by the division.

stephenc: https://github.com/kohsuke/github-api/blob/4ee3086b6dcb46913f1341b2500a4790db5813d6/src/main/java/org/kohsuke/github/GHRateLimit.java#L31 is why we need to do the division... and that is madness... but fixing that madness is orthogonal to this enhancement

stephenc: https://github.com/kohsuke/github-api/pull/329 will fix the algebra... but it is not strictly required as a limit of no limit expiring in 3.6 seconds or expiring in 1 h is still a limit of no limit

stephenc: But without correcting for the 1000 fold factor then GHE would say it had a limit of 1000000 being refreshed on Tue Jun 20 01:41:44 IST 48997 which allows 21 API calls per year... so code that uses the rate limit to pre-throttle will run unnecessarily slow on GHE

stephenc: With #329 code running against GHE and proactively throttling will make 277 API requests per second, while without #329 it might try and make 277777 requests per second as its upper limit... which is probably unatainable

stephenc: Or were you accessing the public field directly (oh how I love the way people make fields public in Java and make migration very difficult without breaking backwards compatibility)

xhumanoid: >> is why we need to do the division... and that is madness... but fixing that madness is orthogonal to this enhancement

Yes, I agree with this, fixing this madness must doing not in this issue.

PR #336, opened 2017-02-07
Fix #335

### LINKED ISSUE #335: OkHttpConnector is broken
The package was changed to `okhttp3.OkUrlFactory;`

auchri:
Correct namespace

https://github.com/square/okhttp/blob/master/okhttp-urlconnection/src/main/java/okhttp3/OkUrlFactory.java

Fixes #335


kohsuke:
For compatibility reasons we can't change this signature, but I added a separate `OkHttp3Connector` to work with OkHttp3.

PR #337, opened 2017-02-09
Remove unused imports

sebkur:
Especially also remove the unsued import of
`javax.xml.bind.DatatypeConverter` from `GHContent` which is non-public API
as of Java 8


KostyaSha:
Makes zero sense because there is no any checker like checkstyle, they will back in any case.

PR #340, opened 2017-02-15
improved branch protection support

jgangemi:
this PR will improves the support around protected branches and bring it inline w/ the current api. what is currently there doesn't work (properly at least) and given how i interpreted the `@Preview` docs, this seems inline.


src/main/java/org/kohsuke/github/Requester.java
```diff
@@ -169,6 +169,11 @@ public Requester with(@WillClose/*later*/ InputStream body) {
         return this;
     }
 
+	public Requester withNullable(String key, Object value) {
+		args.add(new Entry(key, value));
```
jgangemi: according to this: https://developer.github.com/v3/repos/branches/#update-branch-protection, these values can be explicitly null and rather then change how `_with` worked, i created the new method.

PR #343, opened 2017-02-19
add latest release

### LINKED ISSUE #341: How can I get Latest Release of Repository?

kamontat:
I already add the latest release, so please add to master and update version so that I can use in my project
Fix #341 

PS  I don't know how to push my fixed module to the maven server so, please. @emlagowski 

Thank you

### DIFF
--- a/src/main/java/org/kohsuke/github/GHRepository.java
+++ b/src/main/java/org/kohsuke/github/GHRepository.java
@@ -300,7 +300,11 @@ public class GHRepository extends GHObject {
     }
     
     public GHRelease getLatestRelease() throws IOException {
-		return root.retrieve().to(getApiTailUrl("releases/latest"), GHRelease.class).wrap(this);
+        try {
+            return root.retrieve().to(getApiTailUrl("releases/latest"), GHRelease.class).wrap(this);
+        } catch (FileNotFoundException e) {
+            return null; // no latest release
+        }
     }
 
     public PagedIterable<GHRelease> listReleases() throws IOException {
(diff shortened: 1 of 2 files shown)


Shredder121:
It seems the indentation is off though.
And also, would a test be good for this feature?
(I'm not a member, so don't take my word on this)

emlagowski:
@kamontat I'm also not a member.

@Shredder121 taking the opportunity, do you know how to run tests locally if I have no access to repositories and organisations that are referenced from the unit tests?

KostyaSha:
You can read test code, it uses file with local configuration. Create additional account and run using it.

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -298,6 +298,10 @@ public GHRef createRef(String name, String sha) throws IOException {
     public List<GHRelease> getReleases() throws IOException {
         return listReleases().asList();
     }
+    
+	public GHRelease latestRelease() throws IOException {
```
KostyaSha: should be `getLatestRelease`,
fix indentation

emlagowski:
@KostyaSha I know that, but it still fails. Is there any list of what scopes an access token needs to pass the tests?

src/main/java/org/kohsuke/github/GHRepository.java
kamontat: fixed, Thank you

- https://github.com/kohsuke/github-api/pull/343/commits/fd859815b039a84c3dd53e38ff8d41fefc2c2b23?diff=unified

kamontat:
@Shredder121 I already testing on it in my own repository, so I sure it's work

but I don't push my test up because I test by using my own repository.

PR #346, opened 2017-02-23
Ensure that connections are closed for error responses

stephenc:
- This was endless fun to trace, but I found it at last. This should
stop the `WARNING: A connection to https://api.github.com/ was leaked.
Did you forget to close a response body?` messages in the logs when
using the OkHttpConnector.

@reviewbybees

### DIFF
--- a/src/main/java/org/kohsuke/github/GitHub.java
+++ b/src/main/java/org/kohsuke/github/GitHub.java
@@ -47,13 +47,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TimeZone;
-import java.util.concurrent.TimeUnit;
-import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import org.apache.commons.codec.Charsets;
 import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.io.IOUtils;
 
 import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.ANY;
 import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.NONE;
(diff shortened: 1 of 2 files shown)


stephenc:
No I do not have an automated test for this fix, the testing I have done is to use the old version and run some code that is known to cause the issue... wait maybe 5 minutes (can show as early as immediately, seems to be at worst every 5 minutes) and then you get the error in the logs

Then switch to the fix, rerun the same code (GHRepository.getCollaborators() with an anonymous connection is good) and then wait 15 minutes... no error logged!

src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -703,8 +702,18 @@ private boolean isPrivateModeEnabled() {
...
-            return uc.getResponseCode() == HTTP_UNAUTHORIZED
-                && uc.getHeaderField("X-GitHub-Media-Type") != null;
+            try {
+                return uc.getResponseCode() == HTTP_UNAUTHORIZED
+                        && uc.getHeaderField("X-GitHub-Media-Type") != null;
+            } finally {
+                // ensure that the connection opened by getResponseCode gets closed
+                try {
+                    IOUtils.closeQuietly(uc.getInputStream());
+                } catch (IOException ignore) {
```
oleg-nenashev: Maybe it;s a leftover from the previous fix. `closeQuietly`does not throw `IOException`s

stephenc: `uc.getInputStream()` does

stephenc: And when using the OkHttpConnector and the HTTP response code is >= 400 it will throw the IOE always (not that JVM's HttpURLConnection will do that)... given that this is the expected case for returning `true`, we need to access the error stream in order to ensure that the request is closed for the `OkHttpConnector` path... which we have to do using the error stream 🙄

KostyaSha:
@stepehenc have you tried to run existing ITs?

src/main/java/org/kohsuke/github/GitHub.java
stephenc: https://github.com/square/okhttp/blob/parent-2.7.5/okhttp-urlconnection/src/main/java/com/squareup/okhttp/internal/huc/HttpURLConnectionImpl.java#L243

stephenc:
@KostyaSha they fail for me before making changes, I do not expect them to magically pass after my changes, but I have done as extensive a testing as possible without the less than stellar automated tests

src/main/java/org/kohsuke/github/GitHub.java
stephenc: Also https://github.com/square/okhttp/blob/parent-2.7.5/okhttp-urlconnection/src/main/java/com/squareup/okhttp/internal/huc/HttpURLConnectionImpl.java#L145 which is the only way to close the request body and prevent the warning!

rsandell:
:bee:

stephenc:
@reviewbybees done, @kohsuke can you merge and release?

PR #350, opened 2017-03-29
Add application/json header when sending json data.

kounoike:
I'm using Jenkins and SonarQube GitHub plugins with [GitBucket](https://github.com/gitbucket/gitbucket).
GitBucket has GitHub-compatible API.
But it is not 100% compatible. This PR solves one of the differences between GitHub and GitBucket. And, it also solves header and body data format mismatch when sending json data.

in current version's Requester.java [buildRequest](https://github.com/kohsuke/github-api/blob/2627dc5ee4ed60c0250ce5a9e99d586d480603f2/src/main/java/org/kohsuke/github/Requester.java#L385) sends json data with content-type: application/x-www-form-urlencoded.
This behavior is incorrect. data and header mismatch. Then, GitBucket was confused by how to handle data.

In GitHub API, GitHub parses this data as json (header content-type is ignored), but this behavior is not documented. In the future this behavior may change.

I think that it is better to accept this modification to avoid future problems.

### DIFF
--- a/src/main/java/org/kohsuke/github/Requester.java
+++ b/src/main/java/org/kohsuke/github/Requester.java
@@ -385,15 +385,16 @@ class Requester {
     private void buildRequest() throws IOException {
         if (isMethodWithBody()) {
             uc.setDoOutput(true);
-            uc.setRequestProperty("Content-type", contentType);
 
             if (body == null) {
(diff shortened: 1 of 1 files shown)


kohsuke:
In order to honor the content type set by the caller of `Requester`, I fixed this problem a little differently. The fix will be in 1.89

PR #359, opened 2017-06-28
[JENKINS-45142] Retry connections after getting SocketTimeoutException

jglick:
[JENKINS-45142](https://issues.jenkins-ci.org/browse/JENKINS-45142)

@reviewbybees

### DIFF
--- a/src/main/java/org/kohsuke/github/Requester.java
+++ b/src/main/java/org/kohsuke/github/Requester.java
@@ -36,6 +36,7 @@ import java.lang.reflect.Field;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.ProtocolException;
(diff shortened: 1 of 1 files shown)


jglick:
BTW I get a ton of failures/errors running tests in this component, which I doubt are caused by my changes here.

stephenc:
- [ ] @stephenc

KostyaSha:
In my github plugins i just do retryable executions from Function and that's all.

jglick:
@reviewbybees done

jglick:
> do retryable executions from Function

Yes it is possible to work around from the caller side, but to limit the fix to the known case where a retry is definitely sensible, it would be necessary to inspect `cause`s for `instanceof SocketTimeoutException`.

Issue #360, opened 2017-07-07
Add support for committing multiple files.


siordache:
A single file can be committed by using one of the GHRepository.createContent() methods.

In order to commit multiple files, the following steps have to be taken:
1. Get the SHA of the latest commit on the branch
2. Get the tree information for that commit
3. **Create a new tree for your commit**
4. **Create the commit**
5. Link commit to the reference

The operations in steps 3 and 4 are currently not suppported by github-api. They involve:
- [creating a blob](https://developer.github.com/v3/git/blobs/#create-a-blob)
- [creating a tree](https://developer.github.com/v3/git/trees/#create-a-tree)
- [creating a commit](https://developer.github.com/v3/git/commits/#create-a-commit)

Do you find the addition of these operations useful? If yes, I will create a PR for this.

PR #361, opened 2017-07-08
issue #360: Add support for committing multiple files

siordache:
Implement #360. New classes:
- GHBlobBuilder - [create a blob](https://developer.github.com/v3/git/blobs/#create-a-blob)
- GHTreeBuilder - [create a tree](https://developer.github.com/v3/git/trees/#create-a-tree)
- GHCommitBuilder - [create a commit](https://developer.github.com/v3/git/commits/#create-a-commit)

[This gist](https://gist.github.com/siordache/63b19f515e4db6944e2411265f89e92f) shows how to use the new classes.

### DIFF
--- a/src/main/java/org/kohsuke/github/GHBlobBuilder.java
+++ b/src/main/java/org/kohsuke/github/GHBlobBuilder.java
@@ -1,0 +1,50 @@
+package org.kohsuke.github;
+
+import org.apache.commons.codec.binary.Base64;
+
+import java.io.IOException;
+
+/**
+ * Builder pattern for creating a new blob.
+ * Based on https://developer.github.com/v3/git/blobs/#create-a-blob
+ */
+public class GHBlobBuilder {
+    private final GHRepository repo;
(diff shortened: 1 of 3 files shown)


siordache:
@reviewbybees

jglick:
@siordache please do not use the above mention. @kohsuke is I think the only one who can now delete the previous two comments.

siordache:
Oh, sorry. I thought that I should use the above mention to indicate that the PR is ready to be reviewed.

KostyaSha:
This is not a jenkins, this @kohsuke owned library. Unfortunately jenkins plugins depends on it, maybe we should raise the question on forking it, because it causes delays for FOSS users. Only those who work with @kohsuke now can influence on releases speed.

jglick:
Yes, various people have requested that @kohsuke move ownership of this and some other repositories necessary for Jenkins to the @jenkinsci org, or otherwise open up access.

PR #369, opened 2017-08-08
- improve branch protection support

jgangemi:
this is a re-submission of #340. i'd like to see this get some traction and thought this the best way to grab attention.

the PR will improves the support around protected branches and bring it inline w/ the current experimental api. the current implementation doesn't work (properly at least) and given how i interpreted the @Preview docs, it did not seem i was bound to maintain existing method signatures.

i'd like to be able to add some user/team restriction tests, but that requires a repository in an organization.


jgangemi:
bueller?

PR #375, opened 2017-09-01
Add basic support for tag objects

stephenc:
@reviewbybees

### DIFF
--- a/src/main/java/org/kohsuke/github/GHRepository.java
+++ b/src/main/java/org/kohsuke/github/GHRepository.java
@@ -784,6 +784,26 @@ public class GHRepository extends GHObject {
        return GHRef.wrap(root.retrieve().to(String.format("/repos/%s/%s/git/refs", getOwnerName(), name), GHRef[].class), root);
     }
 
+
+    /**
+     * Retrieves all refs for the github repository.
+     *
+     * @return paged iterable of all refs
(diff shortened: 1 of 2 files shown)


src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -810,6 +851,18 @@ public GHRef getRef(String refName) throws IOException {
...
+
+    /**
+     * Returns the <strong>annotated</strong> tag object. Only valid if the {@link GHRef#getObject()} has a
+     * {@link GHRef.GHObject#getType()} of {@code tag}.
+     *
+     * @param sha the sha of the tag object
+     * @return the annotated tag object
+     */
+    public GHTagObject getTagObject(String sha) throws IOException {
+        return root.retrieve().to(getApiTailUrl("git/tags/" + sha), GHTagObject.class).wrap(this);
```
KostyaSha: is really only sha?

stephenc: Nope it's the tagged and the tag date and the sha of the tagged commit.

For an annotated tag object the tag ref points to a tag object that points to the commit

For a lightweight tag, the tag ref just points to the commit.

stephenc: S/tagged/tagger/

PR #378, opened 2017-09-13
bridge-method-annotation should be an optional dep

jglick:
b6e48cc4f9c7b13895673a7b4aa6399b7e236b1f introduced a hard dep on a lib which is newer than that currently bundled in core, causing problems for `requireUpperBoundDeps`.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -131,6 +131,7 @@
       <groupId>com.infradna.tool</groupId>
       <artifactId>bridge-method-annotation</artifactId>
       <version>1.14</version>
+      <optional>true</optional>
     </dependency>
     <dependency>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -131,6 +131,7 @@
       <groupId>com.infradna.tool</groupId>
       <artifactId>bridge-method-annotation</artifactId>
       <version>1.14</version>
+      <optional>true</optional>
```
KostyaSha: or scope provided?

jglick: Either seem to work pretty much the same.

vivek:
LGTM 🐝

recampbell:
@reviewbybees complete

PR #379, opened 2017-09-20
Roles for team members

amberovsky:


benchesnut:
Looks good to me. It would be very useful to be able to set roles for team members.

PR #352, opened 2017-03-30
Add support for PR reviews preview

stephenc:
@reviewbybees

PR #384, opened 2017-10-05
Add support for pr review/review comment events

mattnelson:
Add support for pr review/review comment events building on #352 

There was one non-passive change with `GHPullRequestReviewState` the enum value did not match the value returned from the github API. I verified this on github.com and an enterprise instance.


mattnelson:
@kohsuke Could I get a review on this?

Issue #348, opened 2017-03-20
Using 'since' parameter when listing users


JakubKahovec:
Hello,

when listing users the GitHub Api provides a parameter _since_ which allows you to specify an id of the user you've seen the last, when you start listing again (i.e after a crash) you skip the users you've seen. It'd great to have this parameter in the api i.e github.listUsers(sinceUserId). Do you think it would be feasible to add it there ?

Thank you

Jakub

kohsuke:
Translating the requirement: it'd be convenient to allow `PagedIterator` to skip forward without actually retrieving the objects in between.

nodoze:
I think the PageIterator simply needs to implement the "page" parameter to skip to a particular page. It support size but not page from what I can tell. 

https://developer.github.com/v3/#pagination

bitwiseman:
@nodoze 
Yes, there is enough information for PagedIterable to know how many pages it ha, you only have get the first page in order to know the total number of pages.   Seems like something that could be added without breaking anything.

bitwiseman:
In #448, another scenario was mentioned - restarting a search from a specific page.  One thing to keep in mind for searches is that the underlying data may have changed.  At that point the page position of items may have changed.  This a generally problem for any paged query though, so maybe that is something that we can leave to consumers of the API to figure out for themselves.

bitwiseman:
I think it would make sense to add a `GitHubPaginator` class that handles page traversal and navigation.  "PagedIterable` would have a method `paginator()` that would return the new class and support going to a specific page, next, previous, etc.

bitwiseman:
See #1197 
Other users would like a method that returns an iterator that only returns results for a specific page range: `pageRange(start, end)`.  Results would be returned only within that page range.  

I'm thinking the following methods: 
`startAtPage(long)`
`endAfterPage(long)`
`pageRange(long, long)`

anujhydrabadi:
@bitwiseman Hey! Thank you for this library, works great! I ended up not using this library for integrating with the paginated GitHub APIs because of the lack of support for specific page number, and wrote my own simple client instead. 

I'd like to implement this, perhaps starting with these new methods in `PagedIterator`, using just the `Link` response header, then moving to the other features that would use the `page` param.
> `hasPrevious()` `previous()` `getCurrentPageNumber()` `getPageCount()` `first()` `last()`

[...]

PR #385, opened 2017-10-06
Addressing issue #348. This will allow skipping to a page number for

nodoze:
the all the search builders . The alternative is to implement it at the page
PagedIterable level which will require changes throughout all the
classes that reference it. Since this is primarily a search issue this
solution might suffice.

Also added support for the repository search parameter "topic"


kohsuke:
Thanks for the submission, but from the API design perspective, these additions should happen to `PagedIterable` and `PagedIterator`.

PR #391, opened 2017-10-30
Add get for all organizations

scotty-g:
Add support for fetching all organizations using the [`/organizations`](https://developer.github.com/v3/orgs/#list-all-organizations) endpoint. Supports paging and allows for the page size to be specified.


scotty-g:
Are any more releases of this project planned?

PR #392, opened 2017-11-01
Now that getUser throws IOException, provide alternative for people who just want the username

stephenc:


kohsuke:
It seems to me that the right thing to do is to have methods like `GHCommitComment.getUser()` not try to populate the `GHUser` object, and instead rely on lazy population, which is already implemented. No?

That way, there is no need for the caller to do anything different.

stephenc:
Well it’s very hard for the caller to budget the rate limit if you have no idea whether the method call will spend another api call.

The use case I have is just to get the username I do not need the user object at all. Further the code paths i’m Calling from will not have the user object populated, so here the lazy fetch both costs an extra api call and performance as well as being unwanted.

If we are doing lazy we should provide accessors for the keys of the lazy objects so that shallow queries need not trigger a lookup.

kohsuke:
Right, I get the use case and where you were going with this. Reduce the API call when you just need the login name.

My point is that we don't need to define a separate signature to achieve it. We can let the caller just do `GHCommitComment.getUser().getLogin()` and have the library not make excessive HTTP calls behind the scene.

stephenc:
I give up

PR #394, opened 2017-11-06
Populate the GHPullRequest if 'mergeable' is null

psiroky:
GitHub sometimes returns a response which has the 'mergeable_state' set
to 'unknown' (ant thus non-null) and the 'mergable' is 'null'. 'isMergable()'
then returns 'null' instead of 'boolean'. This can to unexpected issues for
the caller.

I am by far not sure that this is a good approach. It can definitely lead to more API calls, which may not be desired. Please let me know if you see a better alternative for this.


kohsuke:
In 4e31636181a2d847a9c6dbb6c186e387951f86af I fixed this differently. Hope that works!

PR #398, opened 2017-11-16
Fix: state for REQUEST_CHANGES is CHANGES_REQUESTED

ubolonton:
https://developer.github.com/v3/pulls/reviews/


iburris:
I'm running into an issue that I believe is related to this. From what I can tell when you submit a pull request `REQUEST_CHANGES` should be used as the state. But, when you get an existing pull request the equivalent state returned is `CHANGES_REQUESTED`. So, it almost seems like this change should be to just leave `REQUEST_CHANGES` as is and add `CHANGES_REQUESTED("CHANGES_REQUESTED"),` or to introduce another enum containing the valid values when reading a review.

kohsuke:
Made a similar change toward 1.91

PR #401, opened 2017-11-20
Replace "new Error" with GHException

### LINKED ISSUE #400: Do not throw new Error()
https://github.com/kohsuke/github-api/search?utf8=%E2%9C%93&q=%22new+Error%22&type=

In general, this seems to be a bad idea.  See also comments in https://github.com/jenkinsci/ghprb-plugin/pull/570. This PR for the GHPRB plugin is left with the dilemma of catching throwable because there's no other choice since this API throws `java.lang.Error`.

A quote from `java.lang.Error` javadoc:

> An `Error` is a subclass of `Throwable` that indicates serious problems that a reasonable application should not try to catch.

Instead, consider throwing `IllegalStateException`.

bjoernhaeuser:
This is a possible fix for #400

Reasoning:

`GHException` is used in other places for similar purposes, therefore I went down that road. Does the changes make sense?

Closes #400

### DIFF
--- a/src/main/java/org/kohsuke/github/GHException.java
+++ b/src/main/java/org/kohsuke/github/GHException.java
@@ -1,5 +1,7 @@
 package org.kohsuke.github;
 
+import java.io.IOException;
+
 /**
  * @author Kohsuke Kawaguchi
  */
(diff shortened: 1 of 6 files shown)


kohsuke:
I've incorporated this change toward 1.91.

PR #402, opened 2017-12-04
tune for sonarsource fork

sns-seb:


src/main/java/org/kohsuke/github/AbuseLimitHandler.java
```diff
@@ -1,3 +1,22 @@
+/*
+ * GitHub API for Java
+ * Copyright (C) 2009-2016 SonarSource SA
```
KostyaSha: Why do you add your license headers to files where you are not authors?

KostyaSha:
Is it fork or what is the purpose of this PR?

PR #413, opened 2018-02-14
fix issues

eosmanov:


bitwiseman:
@kohsuke  It looks like the majority of changes were already take for this PR.  Can we take the last few?

PR #420, opened 2018-02-28
OkHttpConnector: Enforce use of TLSv1.2 to match current Github and Github Enterprise TLS support.

randomvariable:
On Feb 8, 2018, Github changed their TLS settings to be 1.2 only.
Most recent Jenkins installs are OK as Java 1.8 defaults to TLS 1.2, however some people see intermittent
or continuous failures with connecting to Github in a variety of configurations:

e.g. https://issues.jenkins-ci.org/browse/JENKINS-49761?jql=project%20%3D%20JENKINS%20AND%20component%20%3D%20github-api-plugin 

This PR creates a new TLS v1.2 only SSLContext and attaches its socket factory to the urlFactory passed to OkHttpConnector, which is used by most Github plugins in Jenkins.

### DIFF
--- a/src/main/java/org/kohsuke/github/extras/OkHttpConnector.java
+++ b/src/main/java/org/kohsuke/github/extras/OkHttpConnector.java
@@ -1,12 +1,24 @@
 package org.kohsuke.github.extras;
 
+import com.squareup.okhttp.ConnectionSpec;
 import com.squareup.okhttp.OkHttpClient;
 import com.squareup.okhttp.OkUrlFactory;
+
 import org.kohsuke.github.HttpConnector;
 
 import java.io.IOException;
+
 import java.net.HttpURLConnection;
 import java.net.URL;
+
+import java.security.KeyManagementException;
+import java.security.NoSuchAlgorithmException;
+
+import java.util.Arrays;
+import java.util.List;
+
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLSocketFactory;
 
 /**
  * {@link HttpConnector} for {@link OkHttpClient}.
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
CC @KostyaSha @lanwen @stephenc who maintain the dependent plugins

KostyaSha:
@oleg-nenashev i don't maintain it since cloudbees .... 
lanwen is not maintaing github-api-plugin, there is nothing on github-plugin side

oleg-nenashev:
Well, I added you to CC so that you know where to redirect the issues 🤷‍♂️

recena:
@kohsuke Do you agree if I merge this PR and cut a release?

src/main/java/org/kohsuke/github/extras/OkHttpConnector.java
```diff
@@ -23,10 +35,33 @@
...
 
     public HttpURLConnection connect(URL url) throws IOException {
         return urlFactory.open(url);
     }
+
+    /** Returns TLSv1.2 only SSL Socket Factory. */
+    private SSLSocketFactory TlsSocketFactory() {
+        SSLContext sc;
+        try {
+            sc = SSLContext.getInstance("TLSv1.2");
```
jtnord: @randomvariable  ow does this affect people that are behind a HTTPS proxy where they connect to github via that proxyusing say TLS 1.1 or SSL and then the proxy does the connection.  Does this lock them out as they will not be able to handshake with the proxy?

holybit:
Are there plans to cut a new version of the plugin?

src/main/java/org/kohsuke/github/extras/OkHttpConnector.java
KostyaSha: I guess it will affect, also people can change connector on their side AFAIR.

PR #424, opened 2018-03-15
Add sha1 to updateFiles 

onoguera-ob:
see https://developer.github.com/v3/repos/contents/#update-a-file


kohsuke:
We need to preserve the existing method signature, so the new additions would have to be new methods.

At this point, createContent has so many parameters that it should switch to a builder pattern.

PR #428, opened 2018-03-27
Fix imports for commons-lang3.

markwoon:
The last commit updated the commons-lang dependency to v3 but neglected to update the imports.


jgangemi:
any chance of this getting merged? it prevents `master` from compiling

kohsuke:
I think I just independently fixed this. I blindly merged someone's PR without verifying the consequence.

PR #439, opened 2018-06-06
Add support for repository searching by "topic"

l3ender:
See https://help.github.com/articles/searching-repositories/#search-by-topic

### DIFF
--- a/src/main/java/org/kohsuke/github/GHRepositorySearchBuilder.java
+++ b/src/main/java/org/kohsuke/github/GHRepositorySearchBuilder.java
@@ -55,6 +55,10 @@ public class GHRepositorySearchBuilder extends GHSearchBuilder<GHRepository> {
         return q("stars:"+v);
     }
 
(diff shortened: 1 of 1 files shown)


ToWi87:
:+1: for the idea and the code, here the related docs:
[https://developer.github.com/v3/search/#search-topics](https://developer.github.com/v3/search/#search-topics)
Question coming up here:
Are there no additions needed regarding the additional header needed as stated in the docs?

l3ender:
@ToWi87 The doc you linked to is for searching/retrieving topics (`/search/topics`). This PR is for using topics to search/retrieve repositories (`/search/repositories`).

If/when there is a desire to integrate searching of topics, then there will likely be need to spend effort on the header information that you mentioned.

Thank you.

ToWi87:
true that, sorry for the confusion.
I'd like to have the search API also work for topics but I'd wait for GitHub to release the feature without the 'beta' header.

PR #461, opened 2018-10-01
Add methods for adding/removing labels to GHIssue

evenh:
This PR adds the functionality required to add/remove labels individually from a `GHIssue`. One thing I haven't quite understood yet, is how/whether the data in this object can be refreshed (e.g. after adding a label, I want to inspect the labels present for a given issue).

I've never contributed to this project before, so please let me know if I have done something horribly wrong :-)


kohsuke:
Thanks for the change.

I massaged the changes a bit, in particular, as a matter of personal preference, I don't like the kind of `addLabel` method that does multiple things like creating a label and silently tolerating missed parameters.

PR #462, opened 2018-10-07
Add archived flag for repository

ingwarsw:
We have many repositories in our org that are archived..
That flag will allow us to skip them in processing..


kohsuke:
I added this in PR #470 without noticing yours first.

ingwarsw:
@kohsuke As long as its there its more than fine 👍

PR #470, opened 2018-10-25
Added archived attribute in GHRepository

recena:
- `archived` attribute was missed and can be interesting in many cases
- Updated parent POM from [17 to 20](https://github.com/kohsuke/pom/compare/pom-17...pom-20)
- Addressed a Apache Maven warning

```
[WARNING] 
[WARNING] Some problems were encountered while building the effective model for org.kohsuke:github-api:jar:1.95-SNAPSHOT
[WARNING] 'build.plugins.plugin.version' for org.apache.maven.plugins:maven-surefire-plugin is missing. @ line 37, column 15
[WARNING] 
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING] 
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING] 
```

@kohsuke, could you take a look? Thanks.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
   <parent>
     <groupId>org.kohsuke</groupId>
     <artifactId>pom</artifactId>
-    <version>17</version>
+    <version>20</version>
   </parent>
 
   <artifactId>github-api</artifactId>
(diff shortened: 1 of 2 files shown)


brian-mcnamara:
@recena Thanks for this, I was actually just about to implement this as I need it too! Thank you!

@kohsuke I would love to use this for our project. Can we create a release?

nielsutrecht:
@kohsuke +1 from me, I'm running into an issue where I'm mass creating pull requests that are failing on archived repositories.

Issue #472, opened 2018-11-20
Implement archive/unarchive functionality


joaoe:
Howdy.

`GHRepository` does not have an archive/unarchive feature.

To archive, a `GitHub.edit("archived", "true")` call is enough.

To unarchive, it is not possible from the REST API and need to be done manually in the web UI.

In case someone wants to implement an unarchive API:
send a POST request to `https://github.com/<organization>/<repository>/settings/archive` or `https://github.com/<organization>/<repository>/settings/unarchive`. The POST body contains:
`utf8=%E2%9C%93&authenticity_token=...&verify=<repository>`. The token `authenticity_token` is embedded in the HTML in the html form.
This will force to use a different set of URLs and scrape some HTML, so lets consider this out of scope.

andrewlalis:
would it not be more intuitive to name the method simply `archive`?

joaoe:
Hi. Could be, but archive is both a verb and a noun so for me at least, naming something as `archive` looks confusing. There are some options of verbs to use here, like `do` or `set`, or name it as `archiveRepository`.

andrewlalis:
I think that if we want to be very explicit, the most java-style way to do it would be `setArchived(boolean isArchived)`

joaoe:
Hi again. Been a bit away.

Are you one of the project maintainers ?

andrewlalis:
No, sadly I'm just someone who is interested in having this functionality implemented.

TV4Fun:
+1. This would be very nice to have.

bitwiseman:
@joaoe @TV4Fun 
I've commented on #473.  If you're interested, please update the PR to address the comments and we can get this merged.

bitwiseman:
I changed the method name to `archive()` and added a test for it.  
This will be in the next version, which should be released in the next few weeks.

lumaxis:
Just coincidentally found this and wanted to leave a note that unarchiving via API is now possible: https://github.blog/changelog/2022-12-13-unarchive-a-repository-via-the-rest-api/

bitwiseman:
@lumaxis 
Could you open a new issue asking for `unarchive` functionality to be added?

PR #473, opened 2018-11-20
Implemented GitHub.doArchive

joaoe:
`doArchive()` will mark a repository as archived.

Issue #472


src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -631,6 +631,27 @@ public void delete() throws IOException {
...
+     * <pre>
+     * org.kohsuke.github.HttpException: {
+     *     "message":"Repository was archived so is read-only.",
+     *     "documentation_url":"https://developer.github.com/v3/repos/#edit"
+     * }
+     * </pre>
+     *
+     * @throws IOException In case of any networking error or error from the server.
+     */
+    public void doArchive() throws IOException {
```
bitwiseman: Would it make more sense to have `setArchived` instead? That would support archiving and unarchiving.

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -631,6 +631,27 @@ public void delete() throws IOException {
...
+        edit("archived", "true");
+        archived = true;
```
bitwiseman: The other `set*` methods on this class don't assign to the field so I would assume it is done automatically?

PR #477, opened 2018-12-02
Add statistics API.

### LINKED ISSUE #330: Missing repository statistics
I wasn't able to find the statistics in your api https://developer.github.com/v3/repos/statistics/

martinvanzijl:
Fixes #330 - Implements the statistics API as per https://developer.github.com/v3/repos/statistics/.
Adds the following to the GHRepository class:

- getContributorStats()
- getCommitActivity()
- getCodeFrequency()
- getParticipation()
- getPunchCard()


martinvanzijl:
@bitwiseman Thanks for your comment. I apologize for taking so long to respond. (I've updated my Github notification settings now.)

Do you mean unit tests?

bitwiseman:
@martinvanzijl  
My turn to be slow to respond.  :) 
Yes, I unit tests.  We now have the start of WireMock based framework where you can make sure the values from the server are being processed correctly.

martinvanzijl:
@bitwiseman I've added some unit tests (and pushed a fix). So far the tests just check that the methods run successfully. I'll aim to add more detailed tests to see that the statistics returned are correct.

Should I be using WireMock for these?

martinvanzijl:
@bitwiseman I've added more detail to the unit tests. Could you please review?

GitHub caches statistics, and if they are out of date it will return "202" while it refreshes them. I return "null" in that case, but I could instead throw an exception, or simply call `sleep()` and try again in five seconds within the method itself. What do you think?

bitwiseman:
@martinvanzijl 
Please use the WireMock testing utility, similar to the GistTest or PullRequestTest. This will let us run tests in CI. 

I've also add you as a maintainer on 
https://github.com/github-api-test-org/github-api .   Please point your tests at that repository if possible, and if not let's discuss if there is anything else you need.

bitwiseman:
@martinvanzijl 


> GitHub caches statistics, and if they are out of date it will return "202" while it refreshes them. I return "null" in that case, but I could instead throw an exception, or simply call `sleep()` and try again in five seconds within the method itself. What do you think?

I'm not sure, I'm not familiar with this part of the code yet.

martinvanzijl:
@bitwiseman Thanks, makes sense. I'll try to make the changes as suggested.

martinvanzijl:
I've uploaded my progress so far. I moved the code to a new class `GHRepositoryStatistics` as suggested, and used the Wiremock framework.

There is more I would like to do:

 - `getContributorStats()` has the "wait-till-ready" loop in the method itself, which can be turned off with a boolean parameter. Do the same for the other "get..." methods.
 - Return classes with accessor methods (e.g. `ContributorStats.getAuthor("name")`) instead of just the PagedIterables or Lists.
 - Cache the statistics and add a "refresh()" method to force a reload of data.

src/main/java/org/kohsuke/github/GHRepositoryStatistics.java
```diff
@@ -0,0 +1,444 @@
...
+        return getContributorStats(true);
+    }
+
+    /**
+     * @param waitTillReady Whether to sleep the thread if necessary until the
+     * statistics are ready. This is true by default.
+     */
+    @SuppressWarnings("SleepWhileInLoop")
+    @SuppressFBWarnings(value = {"RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE"}, justification = "JSON API")
+    public PagedIterable<ContributorStats> getContributorStats(boolean waitTillReady) throws IOException, InterruptedException {
```
bitwiseman: There's a similar need for retry loop with PullRequests and checking for whether they are mergable (and getting `mergeCommitSha`).   Onc we publish this, we are commited to supporting it going forward, but I'm not sure yet that this is model I'd like for follow.  

Would you be willing to mark this as `@Preview @Deprecated`?  That will give the project the freedom to make breaking changes (such as moving this behavior to a common helper the does `pollFor(...)`).

src/main/java/org/kohsuke/github/GHRepositoryStatistics.java
```diff
@@ -0,0 +1,444 @@
...
+        }
+
+        public String toString() {
+            return "Day " + getDayOfWeek() + " Hour " + getHourOfDay() + ": " +
+                    getNumberOfCommits() + " commits";
+        }
+    }
+
+    String getApiTailUrl(String tail) {
+        return repo.getApiTailUrl(tail);
```
bitwiseman: All of the children are under `stats/`, so that string should move in here. 
```suggestion
        return repo.getApiTailUrl("stats/" + tail);
```

src/main/java/org/kohsuke/github/GHRepositoryStatistics.java
martinvanzijl: Makes sense, I've marked it as suggested.

Should I make the default value `false`, to bring it in line with the other getters?

src/main/java/org/kohsuke/github/GHRepositoryStatistics.java
martinvanzijl: Thanks, I've made this change.

src/main/java/org/kohsuke/github/GHRepositoryStatistics.java
bitwiseman: Meh.  Let's leave it.  It is definitely better behavior.

Issue #376, opened 2017-09-07
Issue events


oliviercailloux:
I miss the ability to find events pertaining to an issue.

Specifically, I am trying to find the first close date of a given issue, and who was assigned to the issue at that time.

This should be possible with https://developer.github.com/v3/issues/events/.

Have I overlooked something?

savindra:
same problem, how to get events under an issue?

PR #479, opened 2018-12-09
Add issue events API

martinvanzijl:
Fix for issue #376 (add issue events API).

Added the following methods:

 - GHIssue.listEvents()
 - GHRepository.listIssueEvents()
 - GHRepository.getIssueEvent(int id)


src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -1597,4 +1597,89 @@ String getApiTailUrl(String tail) {
         if (tail.length()>0 && !tail.startsWith("/"))    tail='/'+tail;
         return "/repos/" + getOwnerName() + "/" + name +tail;
     }
+
+    public static class IssueEventInfo {
```
bitwiseman: @martinvanzijl 
Does this need to be a child class inside `GHRepository`?

martinvanzijl: @bitwiseman 
I've moved it to a separate class called `GHIssueEvent`.

martinvanzijl:
I've uploaded my latest progress.

I want to make the unit tests more robust by creating an issue within the test itself, editing it and making sure the appropriate events are created. I also want to test the fields of the `GHIssueEvent` object itself.

bitwiseman:
Cool just tell me when you're ready for review.

martinvanzijl:
@bitwiseman 
I've updated the unit tests. Can you please review?

Issue #459, opened 2018-09-28
Get the GHAuthorization-s for an authenticated user


betehess:
Once authenticated, I don't see a way to get the scope of the OAuth token that was used. More generally, I don't see how to fetch all the authorization tokens for the user.

I believe that should be doable: https://developer.github.com/v3/oauth_authorizations/#list-your-authorizations .

PR #481, opened 2018-12-10
Added method to list authorizations

martinvanzijl:
Fixes issue #459.

Added method `GitHub.listMyAuthorizations()`.


martinvanzijl:
@bitwiseman I added a basic unit test. Can you please check?

I'm not sure how to check that the authorizations themselves are correct, since they depend on the user running the test. Can I change to a "known user" in the test itself?

bitwiseman:
@martinvanzijl 
TL;DR - What you've written is good enough.

Kohsuke had a separate dummy login that he used for this, but that doesn't scale to more people.  Where possible I'd like tests to use the `github-api-test-org` since multiple people can use it interchangeably. what that isn't possible (as here), then the snapshot you take is good enough. The snapshot becomes the "known user"/stable data.  😄

You can then add checks against that stable data and anyone running the test should get the same result.  Maybe mark the test using the `snapshotNotAllowed()` method and explain why.  

FYI: I'd like to have robust tests that can be re-snapshotted freely, but if that can't be accomplished without a lot of extra work, I'm okay with tests that verify the functionality in CI using canned data that may require work to update if the tests change.

Issue #466, opened 2018-10-17
[Gists] Edit Gists Support


QzSG:
Since we can already get Gist by ID, will support to edit gists be implemented?

PR #484, opened 2018-12-19
Add support for editing Gists

martinvanzijl:
Fixes issue #466.

Added the following methods:

- GHGist.setDescription
- GHGist.addFile
- GHGist.updateFile
 - GHGist.renameFile

NOTE: I could not get the deleteFile() method to work. I have left it commented-out.


bitwiseman:
@martinvanzijl  There's actually `GistBuilder`, so it would be better if we had a `GistUpdater` class to do the updating.

martinvanzijl:
@bitwiseman 
Thanks for reviewing. I created the `GHGistUpdater` class to do the updating as suggested, and added unit tests. Can you please check?

I still can't get the `deleteFile` method to work.

src/main/java/org/kohsuke/github/GHGistUpdater.java
```diff
@@ -0,0 +1,60 @@
...
+        files = new LinkedHashMap<>();
+    }
+
+    public GHGistUpdater addFile(String fileName, String content) throws IOException {
+        updateFile(fileName, content);
+        return this;
+    }
+
+//    // This method does not work.
+//    public GHGistUpdater deleteFile(String fileName) throws IOException {
```
bitwiseman: @martinvanzijl 
This probably has to do with the way that values are passed via maps.  
I suspect passing `null` results in the key not being set or at least not set to GitHub.

Issue #491, opened 2019-02-01
GHTeam.add fails


arykov:
GHTeam.add fails due to use of incorrect roles case when calling github. Problem is caused by using .with("role", role.name()) instead of .with("role", role)

PR #492, opened 2019-02-01
fixed membership role problem with case

arykov:
Fix for problem described in issue #491

### DIFF
--- a/src/main/java/org/kohsuke/github/GHTeam.java
+++ b/src/main/java/org/kohsuke/github/GHTeam.java
@@ -140,7 +140,7 @@ public class GHTeam {
      */
     public void add(GHUser user, Role role) throws IOException {
         org.root.retrieve().method("PUT")
-                .with("role", role.name())
+                .with("role", role)
                 .to(api("/memberships/" + user.getLogin()), null);
     }
 


arykov:
While I share your sentiment, I don't see the original test for the functionality I fixed that you want me to modify. Care to point me in the right direction?

bitwiseman:
@arykov 
Hmm... Looks like there aren't any for that class... 😭 
You could use https://github.com/kohsuke/github-api/blob/master/src/test/java/org/kohsuke/github/GHOrganizationTest.java as a general template and go from there.  

That said, I see the passing enum calls in to this method which does the toLowerCase that is needed:
https://github.com/kohsuke/github-api/blob/d593e584dd3b8504015d2bc145caaabe02ee08a4/src/main/java/org/kohsuke/github/Requester.java#L155-L162

It would be great to have a test ensuring this continues to work, but I understand that may be too involved.  Would you be willing to take a swing at it?

arykov:
I don't see myself able to devote sufficient amount of time to creating this test from scratch in the foreseeable future.  Sorry about this.

bitwiseman:
@arykov  No problem.  This is still a reasonable change.  I'll go ahead and merge.

Issue #495, opened 2019-02-12
Unable to replace a label in an issue because GHIssue.labels list is not updated


rmetzger:
Consider the following code:

```java
GHRepository repo = gitHub.getRepository("myacc/myrepo");
GHIssue issue = repo.getIssue(4);
// assume "test" exists
issue.removeLabels("test");
issue.addLabels("test1");
```

Expected result: Issue `#4` has one label: `test1`
Actual result:  Issue `#4` has two labels: `test`, `test1`

Cause:
In `GHIssue`, the `labels` field is not updated when the `setLabels(String... labels)` method is called. `addLabels("test1")` calls `getLabels()` in `_addLabels()`, which returns the cached `GHIssue.list` field, still containing the `test` label.

PR #496, opened 2019-02-12
Update GHissue.label accordingly when labels change

rmetzger:
Update Findbugs Plugin to 3.0.5 to be compatible with Maven 3.6.0

This addresses #495.


src/main/java/org/kohsuke/github/GHIssue.java
```diff
@@ -215,7 +218,8 @@ public void assignTo(GHUser user) throws IOException {
     }
 
     public void setLabels(String... labels) throws IOException {
-        editIssue("labels",labels);
+        GHIssue updatedIssue = editIssue("labels", labels);
+        this.labels = updatedIssue.labels;
```
bitwiseman: @rmetzger  Why do the other `set*` methods to set their fields as well?

bitwiseman:
@rmetzger 
Ah, this is the issue also mentioned in #475.

bitwiseman:
@rmetzger I'd like to see this solved by moving to a Builder/Updater pattern instead. We're using elsewhere to positive effect.  Would you up for implementing that instead?

src/main/java/org/kohsuke/github/GHIssue.java
```diff
@@ -180,8 +180,11 @@ private void edit(String key, Object value) throws IOException {
         new Requester(root)._with(key, value).method("PATCH").to(getApiRoute());
     }
 
-    private void editIssue(String key, Object value) throws IOException {
-        new Requester(root)._with(key, value).method("PATCH").to(getIssuesApiRoute());
+    /**
+     * @return the updated GHissue object.
+     */
+    private GHIssue editIssue(String key, Object value) throws IOException {
+        return new Requester(root)._with(key, value).method("PATCH").to(getIssuesApiRoute(), GHIssue.class);
```
bitwiseman: As I've learned more, I've found `Requester` supports update in place as well: 
```suggestion
new Requester(root)._with(key, value).method("PATCH").to(getIssuesApiRoute(), this);
```

I'm not sure the `GHIssue` fields that come back from patch are actually a fully populated object. 
But if they are that would save us from needing to change the method below.

bitwiseman:
I'm going to close this for now.  Let's discuss a design that can be consistent across the project - see #584.

PR #498, opened 2019-02-19
Add GHPullRequestReview.submitted_at field

rmetzger:
This is a fix for https://github.com/kohsuke/github-api/issues/450

### DIFF
--- a/src/main/java/org/kohsuke/github/GHPullRequestReview.java
+++ b/src/main/java/org/kohsuke/github/GHPullRequestReview.java
@@ -25,6 +25,7 @@ package org.kohsuke.github;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
+import java.util.Date;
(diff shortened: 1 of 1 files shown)


rmetzger:
Yeah, I think their documentation is out of sync (I also notified GitHub about it).
But the API returns the field: https://api.github.com/repos/kohsuke/github-api/pulls/517/reviews

bitwiseman:
@rmetzger  
Ah, okay, but until it is in the docs, it might change, right? 
Also, what about updatedAt? What should be done about that? 
Maybe the a better solution is to override them and have them throw "NotImplemented"?
I'm open to discussion.

rmetzger:
Before opening this pull request, I reached out to GitHub's support, and they consider this field as stable. Apparently, the team just forgot to document it (and my question to the support was not reason enough for them to fix it :( )

rmetzger:
Thanks a lot for merging!

PR #499, opened 2019-02-20
Create PR from original repo to private org fork

anatolyD:
Use case: when fork is done into private organization then the original repo maintainer can't have any access there hence this field has to be set to `false` explicitly.

Added parameter: `maintainerCanModify` to `GHRepository.createPullRequest` method. 

- Indicates whether maintainers can modify the pull request

Original method is kept for backward compatibility defaulting the parameter to `true`.


anatolyD:
hey, yeap, that works. I can do that both manually and using API directly. 
when you say you'll need some tests what do you mean? cover the change with UTs?

PR #509, opened 2019-03-26
Fixed typo of ADMIN_HOOK variable

baymac:
Fixed the variable typo and also sending a PR to Github Plugin which uses this variable.


baymac:
Hi @KostyaSha, do you have any idea why it breaks? Is happening because I didn't recompile before commit?

KostyaSha:
You shouldn't change public fields, you will break all software using it and nobody will break  their software because of typo. This naming will stay here forever now

baymac:
okay @KostyaSha, thanks for the enlightenment. :)

PR #517, opened 2019-04-26
Adds the ability to get a repository by ID

jamesatha:
Fixes: https://github.com/kohsuke/github-api/issues/515

### DIFF
--- a/src/main/java/org/kohsuke/github/GitHub.java
+++ b/src/main/java/org/kohsuke/github/GitHub.java
@@ -474,6 +474,14 @@ public class GitHub {
         String[] tokens = name.split("/");
         return retrieve().to("/repos/" + tokens[0] + '/' + tokens[1], GHRepository.class).wrap(this);
     }
+
+    /**
+     * Gets the repository object from its ID
+     */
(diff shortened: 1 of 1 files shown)


src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -474,6 +474,14 @@ public GHRepository getRepository(String name) throws IOException {
         String[] tokens = name.split("/");
         return retrieve().to("/repos/" + tokens[0] + '/' + tokens[1], GHRepository.class).wrap(this);
     }
+
+    /**
+     * Gets the repository object from its ID
+     */
+    public GHRepository getRepositoryById(String id) throws IOException {
+        return retrieve().to("/repositories/" + id, GHRepository.class).wrap(this);
```
recena: @jamesatha Is there this endpoint documented? I was not able to find it [here](https://developer.github.com/v3/repos/).

jamesatha: Officially no. But it is supported. The issue filed gives a bit more info about it's "documentation" and how other clients support it: https://github.com/kohsuke/github-api/issues/515

PR #530, opened 2019-07-19
Add GitHub API requests logging

bozaro:
This information is very useful on GitHub API rate limit exhaustion investigation.


bitwiseman:
@bozaro  
This is looks reasonable.  Need some tests for it.  Maybe you can try out the mocking framework, I've added and create some tests the verify the logging output?

Issue #382, opened 2017-09-26
create a little MockGitHub class for tests mocking out the github REST API


jstrachan:
there's a few projects I'd really like a simple mock implementation of `GitHub` for easier testing. 

I'm fine with it being pretty basic (and defaulting to just not finding stuff ;) but if it was able to find users/orgs/repos/PRs and whatnot from in memory objects it'd be super handy.

I started a little spike using a real JAXRS server side and tried reusing the DTOs from this excellent library; but the way the DTOs work they are kinda hard wired into the Requestor; so figure it was maybe simplest to just add a little MockGitHub derived class. By barely changing the code I was able to get something basic going without breaking the `GitHub` API or implementation.

I can submit a little PR to show you the kinda thing am thinking. Its kinda an extension of the `offline` mode but just with a custom API that lets you populate some resources etc

jstrachan:
@kohsuke what do you think of the general idea? I've got a little spike going but thought I'd wait for your thoughts before i go further. Can query/add users/orgs so far - gotta figure out how to support creating repos etc...

jgangemi:
couldn't you do this w/ `wiremock` and return canned json responses?

either way, i support this idea!

ometa:
This would be an excellent addition.

jstrachan:
Here's my first spike https://github.com/jstrachan/github-api/tree/mock-api

e.g. https://github.com/jstrachan/github-api/blob/mock-api/src/test/java/org/kohsuke/github/mock/MockGitHubTest.java#L45

so we can create a mock user + list the repos for the user & create a new repository. There's an issue printing out the repository though (ownerName issues)

kohsuke:
@jstrachan , my apologies for the delay. Yes, I think better testability would be great, and ability to mock things would be wonderful. I think @stephenc worked in some Jenkins plugins to extensively mock GH*** objects from literal JSON files. How can I help make this happen?

Issue #316, opened 2016-11-26
Is there a way to make this library more test friendly?


davidxia:
Thanks for this library.

I'm trying to write unit tests that use objects like `GHRepository`, but I'm having a hard time creating a `GHRepository` instance with dummy data. There are no static constructors. It looks like the only way to create one with attributes is to deserialize from a JSON string with an `ObjectMapper`.

Can we make this library more test friendly?

kohsuke:
I'd use something like [Mockito](http://site.mockito.org/) for that purpose. Are there some specific use cases where that doesn't work nicely? Constructors with attributes will go quickly out of control with a big object like `GHRepository`, and I hate making these objects mutable.

Another idea is to provide a `HttpConnector` that records & replays the HTTP calls into/from a file, which eliminates the need for a long & verbose mock object setups.

davidxia:
If I make a mock object, I'll have to look at github-api source code and
mock tons of methods. My tests end up being pretty brittle as they have to
know the implementation of how I use this library's classes in my business
logic.

You can use static constructors for classes with few attributes and the
builder pattern for bigger ones to make immutable objects, i.e. they don't
have to have setters.

Does this make sense? I can provide an example of how using a mock can be
brittle and a bit tedious.

KostyaSha:
@davidxia please let me know if you decide to make spotify/github-api

kohsuke:
@davidxia I'm not trying to claim that mock is perfect, but I don't see a big difference between mock vs static construction. In both cases, you'd have to construct an object graph with every attribute your business logic uses. And thinking about creating all the builders for each of the `GH*` classes make me dizzy.

That's why I suggested a replay at HTTP layers. That way, people writing tests will use actual captured JSON payloads, and I don't need to write all the builders.

bitwiseman:
Added WireMock facility working great.

PR #537, opened 2019-08-31
Add WireMock testing facility

bitwiseman:
@kohsuke 
Does this look reasonable to you? 

Related to #316 and #382


kohsuke:
If I might only add for one thing, it might be worthwhile documenting that it's now possible to write this kind of tests.

IIRC @stephenc added a mechanism to do this kind of testing some time ago as well, so I thought there might have been some discoverability challenge.

PR #538, opened 2019-09-03
Namespace PR head queries with repo's owner by default

ewiegs4:
The [create a pull request API](https://developer.github.com/v3/pulls/#create-a-pull-request) does not require a namespace on the head branch when creating a pull request between two branches in the same repository.

However, the [list pull requests API](https://developer.github.com/v3/pulls/#list-pull-requests) **does** require a namespace on the `head` value in order to work properly. If the namespace is omitted, the call _ignores_ the `head` parameter entirely, returning all pull requests that match on the other criteria.

In my opinion, this seems like an oversight in the API and creates quite a pitfall.

For example, creating a pull request between two branches in the same repository looks like:
`repository.createPullRequest("title", "my-branch", "master", null);`
But the query to pull back the newly created pull request must look like:
`repository.queryPullRequests().state(GHIssueState.OPEN).head("my-org:my-branch").base("master").list();`
Forgetting the `my-org:` results in all open pull-requests into the master branch being returned, which is clearly not what was intended in the query.

This change causes the `head` value to be namespaced by default if a non-namespaced value is provided, given that providing a non-namespaced value is effectively the same as `null` and will never behave as the developer intended.


ewiegs4:
Yeah, I was a little surprised with the state of the tests. I have added a test for the "qualified" case.

I noticed that many of the tests (including these PR tests) use `github-api-test-org`, which I don't have access to. I figured the CI build would verify my tests work by using credentials with access to that org, but upon further inspection it seems like tests are skipped in the CI build, so I guess someone with access has to run the tests on their machine to make sure everything passes? Let me know if I need to do anything additional. Thanks!

bitwiseman:
@ewiegs4 
Yeah, if you plan to continue contributing I will gladly grant you access to that org so you can have an easier time writing tests.  

In the meanwhile, I took your PR as chance to merge (and the rewrite) the WireMock-based framework I've been working on.  Your test now runs (using mocked data) as part of CI.

ewiegs4:
@bitwiseman 
I use this project a fair amount so could certainly see myself contributing again in the future.

Thanks for the timely responses and guidance. Let me know if you need anything else on my end before merging.

PR #542, opened 2019-09-11
Improved OkHttpConnector caching behavior

bitwiseman:
In the Jenkins project we've seen a number of less than ideal behaviors by the OKHttp cache.  
Some instances it seems bad data gets in and then gets stuck in the cache. In other cases it appears the cache obey the `max-age=60` returned by GitHub and thus doesn't even look for new data from some requests.   

The cache behavior is fine for client-side uses such as Android where optimizing for network usage is important.  Users of this library are are generally less concerned with network usage - they turn on caching to avoid using up their itHub rate-limit budget.

This change modifies the default OkHttpConnect behavior to make OkHttp always check the cache against the GitHub site.  OkHttp still sends the response `ETag` in the request which lets GitHub return a [`304 Not Modified`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/304) if nothing has changed and does not count that request against the users rate-limit. OkHttp handles this response automatically and returns the cached data.  

This means that by using a bit more network bandwidth, this change guarantees up-to-date data while keeping the same rate-limit savings. 

This change has no effect outside the scenario where OkHttpConnector is used with caching.

This is basically the same behavior added to Jenkins [here](https://github.com/jenkinsci/github-branch-source-plugin/blob/e7ac121084bcdf1b5959668cea3fb3a22f380686/src/main/java/org/jenkinsci/plugins/github_branch_source/Connector.java#L726-L748).


bitwiseman:
I found this while doing some exploration of the OkHttpConnector with WireMock Snapshotting.  Still need tests to show how this change effects query behavior.

KostyaSha:
For mocking better just disable cache

bitwiseman:
@KostyaSha 
Generally true, but the point here is to show the behavior of caching and how the proposed change to OkHttpConnector improves the behavior.

Tests aside from this one will not have caching enabled.

bitwiseman:
Turned off the tests for now.  Caching is painful.

PR #544, opened 2019-09-17
Adding possiblity to get ssh keys

arngrimur-seal:


src/main/java/org/kohsuke/github/GHUser.java
```diff
@@ -36,6 +35,10 @@
  */
 public class GHUser extends GHPerson {
 
+    public List<GHKey> listKeys() throws IOException {
```
bitwiseman: Do you have a specific reason you used `list` instead of `get`? 

```suggestion
    public List<GHKey> getKeys() throws IOException {
```

src/main/java/org/kohsuke/github/GHUser.java
```diff
@@ -36,6 +35,10 @@
...
+        return Collections.unmodifiableList(Arrays.asList(root.retrieve().to("/users/"+login+"/keys", GHKey[].class)));
```
bitwiseman: It is possible for there to be multiple pages of keys?  I know it is unlikely for some one to have more than a page of keys, but if it is possible it might be better to use a `PagedIterable`.

src/test/java/org/kohsuke/github/UserTest.java
```diff
@@ -27,4 +26,27 @@ public void listFollowsAndFollowers() throws IOException {
         assertEquals(30, users.size());
         return users;
     }
+
+    @Test
+    public void getKeys() throws IOException {
```
bitwiseman: This is really great.  How was it using this test framework and `takeSnapshot`?  I just added it and would appreciate any feedback you can offer.

src/main/java/org/kohsuke/github/GHUser.java
bitwiseman: Ah, I see we use `list` elsewhere.  Got it.

src/test/java/org/kohsuke/github/UserTest.java
arngrimur-seal: It took a while before I understood the framework but then it was simple. Deleting and generating the test data has drawbacks since it is easy  do forget about that step but the advantage with test taking lots of time and that data is constant is a greater benefit.

src/main/java/org/kohsuke/github/GHUser.java
arngrimur-seal: I think it is a over kill using `PagedIterable`, but it is of course possible. Most people only have a single ssh key and less than 10. 

Personally I think it better to return a list and instead read all the keys in `listKeys()` . I can of course iterate until all key have been read

src/test/java/org/kohsuke/github/UserTest.java
bitwiseman: Okay, well, if you have any suggestions for how to make it easier (such as improvements to the CONTRIBUTING.md) that would be great.

src/main/java/org/kohsuke/github/GHUser.java
bitwiseman: @arngrimur-seal
I see your point.  

Okay, I just did a quick review of the code base.  Generally, `list*` methods return `PagedIterable<T>`.  If there is a method that returns a `List<T>` or a `T[]` it is called `get*`.   

I think this completely misleading naming, but I'd rather be consistent with the existing naming pattern.  So, let's change the name to `getKeys()`.  Then we're good to go.

arngrimur-seal: Not sure I understand, are we good to merge or do you want me to update the code so we  make sure all keys has been rtrieved?

src/main/java/org/kohsuke/github/GHUser.java
bitwiseman:
```suggestion
    public List<GHKey> getKeys() throws IOException {
```

src/main/java/org/kohsuke/github/GHUser.java
bitwiseman: Added a suggestion comment above.

src/test/java/org/kohsuke/github/UserTest.java
```diff
@@ -27,4 +26,27 @@ public void listFollowsAndFollowers() throws IOException {
...
+        GHUser u = gitHub.getUser("rtyler");
+        List<GHKey> ghKeys = new ArrayList<>(u.listKeys());
```
bitwiseman:
```suggestion
        List<GHKey> ghKeys = new ArrayList<>(u.getKeys());
```

PR #545, opened 2019-09-19
Support for projects

gskjold:
An extension of https://github.com/github-api/github-api/pull/483 where project columns and cards have been added, as well as some tests


src/main/java/org/kohsuke/github/GHProject.java
```diff
@@ -0,0 +1,189 @@
...
+        this.root = repo.root;
+        return this;
+    }
+
+    public GHProject wrap(GitHub root) {
+        this.root = root;
+        return this;
+    }
+
+    private void edit(String key, Object value) throws IOException {
```
bitwiseman: @gskjold 

What do you think of the builder/updater style instead? 

https://github.com/github-api/github-api/blob/master/src/main/java/org/kohsuke/github/GHRelease.java
https://github.com/github-api/github-api/blob/master/src/main/java/org/kohsuke/github/GHReleaseBuilder.java
https://github.com/github-api/github-api/blob/master/src/main/java/org/kohsuke/github/GHReleaseUpdater.java

It would reduce round trips but would add more code.  

I don't have a strong opinion.  The project doesn't seem to have one particular style it uses, I'd like to move toward a more consistent style.  I'm wondering what you think. 

(Maybe @kohsuke can comment as well if he's around.)

src/main/java/org/kohsuke/github/GHOrganization.java
```diff
@@ -193,6 +194,44 @@ public void conceal(GHUser u) throws IOException {
         root.retrieve().method("DELETE").to("/orgs/" + login + "/public_members/" + u.getLogin(), null);
     }
 
+    /**
+     * Returns the projects for this organization.
+     * @param status The status filter (all, open or closed).
+     */
```
bitwiseman:
```suggestion
     */
     @Preview @Deprectated
```

PR #562, opened 2019-10-07
GitHub workflow: enable Java matrix [ '1.8.0', '11.0.x' ]

sullis:


bitwiseman:
@sullis 
Great idea, but it looks like maybe bridge method facility is not compatible in some way. 
We use those to maintain backward compatibility.    

It looks like the indexer that would usually create the file to tell the bridge method injector what classes to process isn't running when expected.  

I've tried compiling on Java 8 and then running on java 11 and that seems okay.  If you can get the annotations I'd be happy to take this PR.

PR #579, opened 2019-10-29
Add methods to update and delete milestones.

### LINKED ISSUE #512: Methods to update milestones
Currently you can only close and reopen them, but it would be nice to also be able to edit the other properties like title, description and due date.

martinvanzijl:
Fixes #512.

Adds the following methods to GHMilestone:

- setTitle
- setDescription
- setDueOn
- delete

NOTE: For the unit test, the returned due date doesn't quite match the one passed to `setDueOn`. I guess this is due to timezone difference.


src/test/java/org/kohsuke/github/GitHubStaticTest.java
```diff
@@ -0,0 +1,72 @@
...
+        Date instantMillis = Date.from(instantNow.truncatedTo(ChronoUnit.MILLIS));
+
+        // TODO: other formats
+        String instantFormatSlash = formatDate(instantMillis, "yyyy/MM/dd HH:mm:ss ZZZZ");
+        String instantFormatDash = formatDate(instantMillis, "yyyy-MM-dd'T'HH:mm:ss'Z'");
+        String instantFormatMillis = formatDate(instantMillis, "yyyy-MM-dd'T'HH:mm:ss.S'Z'");
+        String instantSecondsFormatMillis = formatDate(instantSeconds, "yyyy-MM-dd'T'HH:mm:ss.S'Z'");
+        String instantBadFormat = formatDate(instantMillis, "yy-MM-dd'T'HH:mm'Z'");
+
+
```
martinvanzijl: I would like to see a short comment above each test to explain what is being checked. e.g. // Millisecond values should be ignored.

bitwiseman: I think 
https://github.com/github-api/github-api/blob/master/src/test/java/org/kohsuke/github/GitHubStaticTest.java#L48
covers this.   Cool?

martinvanzijl: Looks good, thanks.

Issue #582, opened 2019-10-30
GitHub app: isCredentialValid doesn't work


timja:
I'm trying to add GitHub app support to the github-branch-source Jenkins plugin, 
I'm facing 2 main issues,

1. From the GitHub class I can't tell if I'm using a GitHub app

2. isCredentialValid / getMyself don't work
```
org.kohsuke.github.GHFileNotFoundException: {"message":"Resource not accessible by integration","documentation_url":"https://developer.github.com/v3/users/#get-the-authenticated-user"}
```
Ideally `isCredentialValid ` would do something else when using a `GitHub` app and not sure if there's anything that can be done about `getMyself`

cc @PauloMigAlmeida

PauloMigAlmeida:
Hi @timja 

Both points have been discussed on #570. I really suggest you read it (I know it's bit lengthy, but that's going to answer most questions you may have)

> From the GitHub class I can't tell if I'm using a GitHub app

The ```GitHub``` class should be diligently created using the appropriate credentials/permissions for what you need... which means that if you get an instance of the ```GitHub``` class using ```new GitHubBuilder().withJwtToken(jwtToken).build()``` then you know you are 'using a Github App'.


> isValidCredential / getMyself don't work
>
> ```
> org.kohsuke.github.GHFileNotFoundException: {"message":"Resource not accessible by integration","documentation_url":"https://developer.github.com/v3/users/#get-the-authenticated-user"}
> ```

This was also discussed on #570 and I proposed to @bitwiseman that a convenience/cosmetic method was created to make it easier for developers to know how to make heads or tails of the multiple ways of authenticating to GIthub with this library.

Basically, the summarised version of why you are facing it would be:

1. App Installation Token is an OAuth token (so is personal access tokens and oauth apps)
2. This library defined the behaviour (many years ago) that if you pass an oauth token with no login then it tries to call the ```getMyself``` which for App Installation tokens flow this isn't applicable as ```/user``` isn't a GitHub App-ready endpoint. Since we may have users relying on this behaviour, we won't change it until we are ready to bump to the 2.x version of this sdk.

**How to address it:**

Before the #583 PR is approved: 
```java
GitHub githubAuthAsInst = new GitHubBuilder()
                    .withOAuthToken(appInstallationToken.getToken(), "")
                    .build();
```

After the #583 PR is approved: 
```java
GitHub githubAuthAsInst = new GitHubBuilder()
                    .withAppInstallationToken(appInstallationToken.getToken())
[...]

PR #583, opened 2019-10-30
Convenience method to auth with app installation token && documentation examples

### LINKED ISSUE #570: Support all available endpoints for Github App with preview request
Not able to access all available endpoints with Github App. Only certain endpoints that are accessible because they have this custom media type set internally.

According to [this doc](https://developer.github.com/v3/apps/available-endpoints/), accessing the Github API with Github App requires custom media type `application/vnd.github.machine-man-preview+json`. Currently, there isn't a way to set this.

PauloMigAlmeida:
Hi @bitwiseman 

This PR fixes #570 #582 things discussed there.

List of things done:

* Added a convenience method to authenticate with app installation tokens;
* Converted existing markdown doc to APT due to formatting features;
   
   The maven-site-plugin (even the 3.3 onwards) Markdown implementation is rather limited and it wasn't exactly clear, visually,  when the code snippets started when I started adding more of them.

   Before: 
image

   After:
image

   
* Replaced occurrences of kohsuke user to github-api org where applicable;

   I noticed that you transferred the github-api project to an org with the same name so I seized the opportunity to fix references in the source code to it too.

* Fixed repositoryIds method on GHAppCreateTokenBuilder;

   That was something that slipped through the cracks on my last PR that both of us didn't notice during the review. The type should be ```Long``` rather than ```Integer```. Considering that you haven't released it yet, I thought that now was the perfect opportunity to fix it. 

   In case you want me to deprecate the original method and create a second one with the right method signature instead of replacing it, let me now ... it's up to you tbh. 

**Future work:**
I plan to create another documentation page with the steps related to deal with Github apps but I decided to first focus on the several ways people can authenticate to Github with this library before anything else (most recurrent question IMHO). Also, I needed to know if I am heading in the right direction that you guys have in mind.


bitwiseman:
> I plan to create another documentation page with the steps related to deal with Github apps but I decided to first focus on the several ways people can authenticate to Github with this library before anything else (most recurrent question IMHO). Also, I needed to know if I am heading in the right direction that you guys have in mind.

File issues or use the the wiki to track future ideas and work.  I'll take a look at this shortly.

src/main/java/org/kohsuke/github/GHAppCreateTokenBuilder.java
```diff
@@ -35,7 +35,7 @@
      *
      */
     @Preview @Deprecated
-    public GHAppCreateTokenBuilder repositoryIds(List<Integer> repositoryIds) {
+    public GHAppCreateTokenBuilder repositoryIds(List<Long> repositoryIds) {
```
bitwiseman: Does this automatically cast from `List<Integer>` to `List<Long>`. 
If not, it would be better to keep both for at least one more release (with an issue filed to remove the `List<Integer>` one). 
I know we have no commitment to maintain it, but we can still try to be nice.

PauloMigAlmeida: Agreed. I will deprecate this one and add the right one. Java auto-boxing won’t be a solution unfortunately. As soon as I get to work I will do it and let you know

src/site/apt/index.apt
```diff
@@ -0,0 +1,154 @@
+What is this?
```
bitwiseman: I'm going to talk through my thoughts here : 

* I have no love of markdown - using something else works fine for me. 
* I want to encourage you to make the docs better.  From that perspective whatever format you want to use is fine by me. 
* I also want others to be able to do the same.  I hadn't heard of APT before this.  Others might not know it either and would need to learn it (though it looks easy enough). 
* I also want to be able to preview the changes on GitHub easily.  This doesn't do that - we just get text.  Bummer. 

Meh, okay, let's try this.

src/main/java/org/kohsuke/github/GHAppCreateTokenBuilder.java
bitwiseman: Nevermind!

bitwiseman: Since this hasn't been released a version with the `Integer` yet, we can just change it.

PR #584, opened 2019-10-30
Convert GHMilestone to updater

bitwiseman:
@martinvanzijl
How about this instead of the `set*` methods?


bitwiseman:
@PauloMigAlmeida  Any thoughts on this?

PauloMigAlmeida:
@bitwiseman I agree with the approach.

martinvanzijl:
@bitwiseman
My thoughts are: I recommend keeping the original methods un-deprecated, since the builder syntax is cumbersome when updating a single field, e.g.:
`milestone.update().title("New Title").update();`

The syntax also looks strange when closing a milestone, e.g.:
`milestone.update().close();`

Just `milestone.close()` makes more sense to me.

bitwiseman:
@martinvanzijl 
Yeah, I see your point.  
On the other hand, offering (and supporting) both has its own problems.
I'm going to close this for the moment.

Issue #591, opened 2019-11-05
GHRateLimit uses public fields


bitwiseman:
https://github.com/github-api/github-api/blob/master/src/main/java/org/kohsuke/github/GHRateLimit.java

Can't be changed without (very likely extensive) API breaking changes.

bitwiseman:
The public fields have been deprecated.  Will be removed in a future version.

PR #595, opened 2019-11-07
Adjust GHRateLimit to system time instead of depending on synchronization

### LINKED ISSUE #383: Parse the response Date header and expose in the Rate Limit information
See https://issues.jenkins-ci.org/browse/JENKINS-47154

The rate limit information should expose the date of the rate limits by parsing the `Date` header rather than putting the consumer at the mercy of clock synchronization.

bitwiseman:
Fixes #383
Related to #591 

@martinvanzijl @PauloMigAlmeida 
 if you have a little time to look at this over, any feedback would be appreciated. 

# Summary
If the client time is out of sync with global time, GHRateLimit could report a `resetDate` that was completely wrong - accurate to global time, but disconnected from local system time.  This wouldn't be so bad, but some clients use the `resetDate` to budget API calls.  If they believe they have 10 hours until reset and only 4000 calls available, they effective lock up.   

Further, public fields and having a "Date" field that is holding unconverted EpochSeconds just bothers me. 

I deprecated the old public fields and added private fields with getters.  I then added awareness of the Date header returned with the Http response.  Using that we can generate `calculatedSecondsUntilReset` and the create the `resetDate` by adding that to System-time that his instance was created.

As a final step, I made it so `rateLimit()` will update automatically when then GHRateLimit instance expires.


PauloMigAlmeida:
Cool! I will take a look at it during lunchtime :)

PauloMigAlmeida:
Hi @bitwiseman, 

> If the client time is out of sync with global time, GHRateLimit could report a resetDate that was completely wrong

Good catch! That is a nice way of fixing the problem. 👍 

**[Re: refactoring/improvements you made]**
The only thing I would possibly change is the way that the content from ```/rate_limit``` response is parsed. 

Currently, we're using the ```JsonRateLimit``` containing only the ```rate``` attribute which has been deprecated. Ideally, we should add the newer JSON attributes listed at https://developer.github.com/v3/rate_limit/#response and refactor all references to the ```rate``` attribute to ```resources.core```.

In case you want to go down that road, the ```Requester.noteRateLimit()``` method will have to be reworked a bit. https://github.com/github-api/github-api/blob/f28edbcf8fcb57a291aaa537bc57ea79c9bcda64/src/main/java/org/kohsuke/github/Requester.java#L354-L362

same here https://github.com/github-api/github-api/blob/f28edbcf8fcb57a291aaa537bc57ea79c9bcda64/src/main/java/org/kohsuke/github/GitHub.java#L329-L338

src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -315,25 +316,25 @@ public void setConnector(HttpConnector connector) {
...
+
     }
 
     /*package*/ void updateRateLimit(@Nonnull GHRateLimit observed) {
         synchronized (headerRateLimitLock) {
             if (headerRateLimit == null
-                    || headerRateLimit.getResetDate().getTime() < observed.getResetDate().getTime()
-                    || headerRateLimit.remaining > observed.remaining) {
+                    || headerRateLimit.getRemaining() > observed.getRemaining()
+                    || headerRateLimit.getResetEpochSeconds() < observed.getResetEpochSeconds()) {
```
PauloMigAlmeida: I must be missing something really simple but... I took a look at the git history and the issues used to discuss the original implementation (back in 2017) and I still don't get it why we wouldn't update the ```headerRateLimit``` variable every time ```updateRateLimit`` method is invoked.

Would you mind sharing with me your understanding of that if statement and/or any idea when this would be applicable?

bitwiseman: Hm, good question.  Ah, I remember now.  Multi-threading. 
If you have 10 threads all make queries and then all try to update the rate limit.  We want to pick the one with the lowest remaining count, of if a reset occurs we want to pick the one with the later reset date. 

But now that you ask, I that that the check is not quite correct.  I need to write a test that covers this and add comments.

bitwiseman:
@PauloMigAlmeida 
Deprectacted: https://developer.github.com/v3/rate_limit/#deprecation-notice - good catch.

bitwiseman:
@PauloMigAlmeida 
So, I've made some changes but it looks like making a new design that really includes the rate limit's aside from core will be another bit of work.  I think the change I've made is sufficient for this PR. And we  can add an issue to do some further deprecation and redesign to expose the other limit information.

src/main/java/org/kohsuke/github/GitHub.java
PauloMigAlmeida: ahhhhh, that makes sense  :)

PauloMigAlmeida:
I agree with you, all of that seems a bit too much for a single PR. 

If you don't mind, after you finish this refactor I would love to try implementing the newer rate limits myself. That should be fun 😄 

On the other hand, if you want to implement it, please count on me for reviewing it too

src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -315,31 +316,61 @@ public void setConnector(HttpConnector connector) {
...
+        if (candidate instanceof GHRateLimit.UnknownLimitRecord && !(current instanceof GHRateLimit.UnknownLimitRecord)) {
+            // Unknown candidate never replaces a regular record
+            return false;
+        } else if (current instanceof GHRateLimit.UnknownLimitRecord && !(candidate instanceof GHRateLimit.UnknownLimitRecord)) {
+            // Any real record should replace an unknown Record.
+            return true;
+        } else {
+            // records of the same type compare to each other as normal.
+            return current.getResetEpochSeconds() < candidate.getResetEpochSeconds()
+                || (current.getResetEpochSeconds() == candidate.getResetEpochSeconds() && current.getRemaining() > candidate.getRemaining());
```
bitwiseman: @PauloMigAlmeida  
This is the check that your question got me to fix. 
A recorder with a later reset is for sure newer that one with an earlier one. 
Only if they have the same reset should we take the one with the lower remaining count.  
Previous implementation would take the lower remaining count candidate even if it had an earlier (older) reset date.  
Thanks!
We should always update to a newer record - one w

bitwiseman:
@PauloMigAlmeida 
I've created a task #605 for exposing the new limit records and integrating the header update behavior. 
Assigning to you.  Enjoy!

bitwiseman:
@PauloMigAlmeida How about now?

src/main/java/org/kohsuke/github/GitHub.java
PauloMigAlmeida: It makes sense! I like the train of thought you used for solving it.

Issue #513, opened 2019-04-12
Access to low-level requester


Vampire:
For being able to use missing API functionality it would be nice if the `Requester` would be accessible, so that the low-level requests can be done manually or e. g. with Kotlin extension functions, but still having the lib do rate limiting and so on.

bitwiseman:
@Vampire 
Seems easy enough (`public` in a few places).  I'm a little concerned though, as then then we'd be committed to the API of the Requester and supporting it.

ranma2913:
This can be done in Groovy already, but isn't possible in Java. Probably not in Kotlin either:
![image](https://github.com/user-attachments/assets/44faf24e-51e9-4a49-b5b6-8a6d67589b20)


```groovy
GHRepository deleteSecret(GHRepository repo, String key) {
    // must use the raw request because kohsuke doesn't have a secret delete yet.
    repo.getRoot()
        .createRequest() // thank you Groovy for allowing this.
        .method("DELETE")
        .withUrlPath("/repos/${repo.getFullName()}/actions/secrets/${key}")
        .send()
    return repo
  }
```

Vampire:
Of course, with Groovy you can even access private stuff easily. And you can do it in Java or Kotlin too with using reflection. But whether you should do it, is a different question, and a stable public API would be preferable

bitwiseman:
@stianst said:

> GitHub is continuously adding new endpoints and fields to existing endpoints, while understandably this library is not always able to support all the new things quickly. The official GitHub CLI client has the same problem, but supports sending any requests using `gh api`.
> 
> It would be nice to be able to create new API requests to endpoints not supported directly by the library, something like `gitHub.createApiRequest`.
> 
> Additionally, for existing endpoints that are supported there is sometimes request parameters or response fields that are not supported. It would be nice to have a way to leverage these without waiting for this library to support them. For example:
> 
> ```
> repository.searchPullRequests().parameter("newParam", "newValue");
> issue.getField(String newField);
> ```

PR #598, opened 2019-11-08
Make low-level requester available to public

bitwiseman:
Initial proof of concept for #513. 

Doing this did help me clean up some odd rough edges in Requester, so I can see some advantages to making this part of the library more user facing and productized.  

At the same time, I think I'd probably still want to mark the `createRequester()` in such a way as to make it clear that anything that comes out of that method may change at any time.

- [ ] Tests.  Lots of tests.
- [ ] Change name of `Requester` to `GitHubRequest`? 
- [ ] Move `Requester` to `internals` package?  - again to make it clear this is available but not supported.


bitwiseman:
Things have changed too much for this PR.  I'll redo this at a later point.

PR #603, opened 2019-11-09
Add Functionality of OTP to support user 2fa

### LINKED ISSUE #602: GitHub.createToken support passing an OTP, thus supporting 2fa
#292 was not resolved despite being marked resolved. With 2fa enabled, the API requires users to go out of pipeline (into a web browser) to generate tokens. This is problematic from an instructions standpoint.

This is also not necessary. 

If create token took an OTP, then the users program could hold and prompt the user to the OTP that comes from SMS. The OTP can be added to the token request via this field:

https://developer.github.com/v3/auth/#working-with-two-factor-authentication

If the API had a field to accept the OTP, then the user would be able to use this library as standard or 2fa mode without user going out of pipeline to generate tokens.

madhephaestus:
close #602 by adding an API to request a token using an OTP.


src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -671,7 +671,26 @@ public GHAuthorization createToken(Collection<String> scope, String note, String
 
         return requester.method("POST").to("/authorizations", GHAuthorization.class).wrap(this);
     }
-
+    /**
+     * Creates a new authorization using an OTP.
+     * 
+     * Start by running createToken, if exception is thrown, prompt for OTP from user
+     * 
+     * Once OTP is recived, call this token request
```
bitwiseman:
```suggestion
     * Once OTP is received, call this token request
```

src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -671,7 +671,26 @@ public GHAuthorization createToken(Collection<String> scope, String note, String
 
         return requester.method("POST").to("/authorizations", GHAuthorization.class).wrap(this);
     }
-
+    /**
+     * Creates a new authorization using an OTP.
+     * 
+     * Start by running createToken, if exception is thrown, prompt for OTP from user
```
bitwiseman: This is what happens on the client side, right?

madhephaestus: This is the stateful use explanation. Since this is unlike any other operation in that it is stateful (request token -> rejection prompts 2f4 OTP generation server side -> user receives OTP -> user calls this function with the time sensitive OTP and this time it is not rejected and the token is returned.

madhephaestus: This raises a sticky question, how do i write a test for a stateful interaction between 2 functions like this?

madhephaestus:
I have a specific question: How do i test an API call that requires access to the users SMS 2fa system? The test would need control of the SMS receive functionality for the account used in the CI system, and the CI account would also need 2fa enabled.

madhephaestus:
When i run the test locally, i get this: 
```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running org.kohsuke.github.Github2faTest
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.411 s - in org.kohsuke.github.Github2faTest
```

But as the step 3 of setting up credentials in CONTRIBUTING.md it says that the moc data is supposed to be generated into the wiremock directory for that test. That directory is empty after the test. How do I proceed?

PauloMigAlmeida:
Hi @madhephaestus, Thanks for the contribution 👍 

Here are my 2 cents on what you could tweak.

**[Re: Creation Token flow]**
While the auth flow (when 2FA is in place) involves getting an HTTP status code 401, this library handles ```HTTP_UNAUTHORIZED``` status code by letting the ```IOException``` go up to the user-accessible API as defined here 

https://github.com/github-api/github-api/blob/f28edbcf8fcb57a291aaa537bc57ea79c9bcda64/src/main/java/org/kohsuke/github/Requester.java#L761-L762

I really appreciate it that you proposed a non-intrusive solution (kudos to you 😃 ) but having the regular method's utilisation depends on nested `try-catch` statements looks a bit unusual. 
(@bitwiseman do you agree?)

```java
GHAuthorization token=null;
try {
	token = gitHub.createToken(asList, string, "this is a test token created by a unit test");
}catch (IOException ex) {
	try {
		token = gitHub.createTokenOtp(asList, string, "", twoFactorAuthCodePrompt());
	} catch (Exception e) {
		e.printStackTrace();
	}
}
```

Another thing that developers can't infer with this approach is whether the exception has been thrown because GitHub sent a 401 (and the `X-GitHub-OTP: required; <sms/app>` header) or any other possible reason (Invalid credentials, network exception and so on).

#########

Just throwing ideas here, but what if we do the following?

We can make the `Requester.handleApiError` method a bit smarter so that when 401 is obtained we check for the existence of the ```X-GitHub-OTP``` header and handle it accordingly. 

The implementation would be something akin to it (not tested):

**PS.:** We can take advantage of the existing `GHIOException` which extends `IOException` while carries the HTTP headers. 

```java
// Don't get caught up over the name of the exception, I'm really bad at choosing names :)
public class GHOTPRequiredException extends GHIOException {
// .....
}
```
`Requester.handleApiError`
```java
if (responseCode == HttpURLConnection.HTTP_UNAUTHORIZED) // 401 Unauthorized == bad creds
    if(uc.getHeaderField("X-GitHub-OTP") != null)
        throw (IOException) new GHOTPRequiredException().withResponseHeaderFields(uc).initCause(e);
    else
        throw e; // usually org.kohsuke.github.HttpException (which extends IOException)
```

Once this is done, we can overload the `createToken` method so it can take a functional interface (that way users can integrate with any mechanism they want) while at the same time we don't break backwards compatibility. Something akin to it.

```java
@FunctionalInterface
public interface OTPGenerator {
    String run();
}
```

```java
// Original method
public GHAuthorization createToken(Collection<String> scope, String note, String noteUrl) throws IOException {
    Requester requester = new Requester(this)
            .with("scopes", scope)
            .with("note", note)
            .with("note_url", noteUrl);

    return requester.method("POST").to("/authorizations", GHAuthorization.class).wrap(this);
}

//Overloaded method
public GHAuthorization createToken(Collection<String> scope, String note, String noteUrl, OTPGenerator otpGenerator) throws IOException{
    try{
        return this.createToken(scope, note, noteUrl);
    }catch (GHOTPRequiredException ex){
        // PS:: I'm not gonna refactor/extract the common lines of code of those 2 methods to make it easier to read on this PR
        // We can also have access to the HTTP headers using the ex.getResponseHeaderFields()

        Requester requester = new Requester(this)
                .with("scopes", scope)
                .with("note", note)
                .with("note_url", noteUrl);
        // Add the OTP from the user
        requester.setHeader("x-github-otp", otpGenerator.run());
        return requester.method("POST").to("/authorizations", GHAuthorization.class).wrap(this);
    }
}
```

That way, the utilisation from the user point-of-view would be like (using the example you created on the test class):

```java
GHAuthorization token = gitHub.createToken(
        asList, 
        "TestCreateTokenOtp",
        "this is a test token created by a unit test", () -> {
            // can be anything from automated processes to user interaction.
            return whateverMethodThatGetsYouTheOTP(); 
        });
}
```

@madhephaestus  @bitwiseman  Let me know what you guys think about this approach and if you agree going forward with something along these lines. (considering that what I suggested is just a draft and many other improvements can be made on it)

also, if I'm missing something out, let me know too.

Happy coding guys 😃

madhephaestus:
@PauloMigAlmeida This! you hit the nail on the head, i was just a bit nervous to make som much of a change for my first PR with your project. I have a similar feature in my [wrapping library](https://github.com/CommonWealthRobotics/GithubPasswordManager) (I add encrypted password/token storage and management of egit credential providers). 

I would be happy to make the changes suggested above, i would basically take the snippets above and implement it in my PR. @bitwiseman Thoughts?

src/test/java/org/kohsuke/github/Github2faTest.java
```diff
@@ -0,0 +1,52 @@
...
+		String string = "Test2faTokenCreate-" + timestamp;// use time stamp to ensure the token creations do not collide with older tokens
+		GHAuthorization token=null;
+		try {
+			token = gitHub.createToken(asList, string, "this is a test token created by a unit test");
+		}catch (IOException ex) {
+			// under 2fa mode this exception is expected, and is necessary
+			// as the exception is called, GitHub will generate and send an OTP to the users SMS
+			// we will prompt at the command line for the users 2fa code 
+			try {
+				//token = gitHub.createTokenOtp(asList, string, "", twoFactorAuthCodePrompt());// prompt at command line for 2fa OTP code
```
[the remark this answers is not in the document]
bitwiseman: Ah, I see.  Yeah, it's a little harder than I first thought. 

Here's my suggestion: 
* Add checks that right information is in the `GHAuthorization` that came back from the server.
* Add a thread sleep of 30 seconds after your call to `createTokenOtp()`. 
* Add actions and checks that verify things are as they should *as though the user responded to the SMS* 
* Run the test with `takeSnapshot` with yourself as the user.  
* Respond to the SMS while the test is waiting. 
* Let the test finish successfully
* Comment out the thread sleep with info explaining this is where a person would get an SMS message and respond
* Add `assumeFalse("Test only valid when not proxying", mockGitHub.isUseProxy());` to the top of your method. 
* Clean up the permissions your account and checkin all the files. 

This will help people understand what this method is for and will ensure if someone goes to change it they have some idea of what needs to be done. 

Cool?

bitwiseman:
@PauloMigAlmeida @madhephaestus 
Yes to all of this as long as you add tests and lots of them.
This may involve adding entire new test classes in some cases because of the lack of testing up to now, but it sounds like even then this will be a win for you and a really nice addition to this library. 

Question: You said SMS, so that has me a little concerned. So, there is user interaction somewhere in here?   Where they library will be waiting for input right?   I think seeing the tests will clarify this for me.

src/test/java/org/kohsuke/github/Github2faTest.java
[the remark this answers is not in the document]
madhephaestus: ok, cool, so the takeSnapshot will record the user interaction. I would then bake in the OTP code during the capture that I get, into the unit test, to match the header data. That way it can be run in mock. Is that right?

madhephaestus:
@bitwiseman , yes SMS as in the first token request prompts GitHub to send you a 6 digit code through a side channel, in this case SMS. To actually get the test to run to generate the moc data i think i need to make a simple swing GUI inside the test. I can't think of another way to get the test to accept the OTP code to make the test pass in the capture phase. Ill then have to **delete all that** and bake in the code I got to match the captured data, which should work in CI moving forward. I promise not to commit any swing cruft ;)

bitwiseman:
@madhephaestus 
Could run it in debug, set a break point, and set the OTP via debugger? 
Or maybe run it on the commandline and enter it by hand?

Not questioning, just looking to make your life easier.  

If you do create some tool for this, we should have it (and/or instructions) checked in somewhere so we can update/recreate the test later.

madhephaestus:
@bitwiseman After much fiddeling i managed to take a snapshot of the OTP transaction and stor it along with the test. The trick of debugging the test and hand-inserting the OTP into the memory worked where every other method had not. Now I will dig into the changes suggested by @PauloMigAlmeida

madhephaestus:
@bitwiseman This PR is tested and ready for review now.

PauloMigAlmeida:
@madhephaestus Thanks 👍 

what about the overloaded ```createToken``` method so that users don't need to forcefully use nested try-catchs to use it?

```
//Overloaded method
public GHAuthorization createToken(Collection<String> scope, String note, String noteUrl, OTPGenerator otpGenerator) throws IOException{
```

Did you give up on that idea for some reason? If so, share with us your thoughts 😃

madhephaestus:
@PauloMigAlmeida give up on, no, I was trying to make the minimal changes to the code base possible.  I'm just enjoying the fact i was able to make the test data and get the unit test to pass.  If there is a strong desire from the core devs to allow me to add that much more, then ill go ahead and set it entirely up as you suggested.

madhephaestus:
@bitwiseman @PauloMigAlmeida Ok! i have made the workflow match Paulo's suggested workflow using a lambda and hiding the try/catch in the overloaded function. Mock data with the new workflow was captured and committed, and the test tweaked to match the captured date (Baked in the OTP i used). I think this should be good to merge/release maybe?

src/main/java/org/kohsuke/github/GHOTPRequiredException.java
```diff
@@ -0,0 +1,5 @@
+package org.kohsuke.github;
+
+public class GHOTPRequiredException extends GHIOException {
```
bitwiseman: Add basic JavaDoc.

src/main/java/org/kohsuke/github/OTPReceiverLambda.java
```diff
@@ -0,0 +1,7 @@
+package org.kohsuke.github;
+
+@FunctionalInterface
+public interface OTPReceiverLambda {
```
bitwiseman: Could this be a `[Supplier<String>](https://docs.oracle.com/javase/8/docs/api/java/util/function/Supplier.html)` instead? 
Is there a reason for it not to be?

src/test/java/org/kohsuke/github/Github2faTest.java
```diff
@@ -0,0 +1,32 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+
+import org.junit.Test;
+public class Github2faTest  extends AbstractGitHubWireMockTest {
+
+	@Test
+	public void test2faToken() throws IOException {
+		//assertFalse("Test only valid when not proxying", mockGitHub.isUseProxy());
```
bitwiseman:
```suggestion
		assumeFalse("Test only valid when not proxying", mockGitHub.isUseProxy());
```
The test will be skipped when proxying, but will run in CI.

src/test/java/org/kohsuke/github/Github2faTest.java
```diff
@@ -0,0 +1,32 @@
...
+		        asList, 
+		        nameOfToken,
+		        "this is a test token created by a unit test", () -> {
+		        	String data = "111878";
+		        	// TO UPDATE run this in debugger mode, put a breakpoint here, and enter the OTP you get into the value of Data
+		            return data; 
+		        });
+		assert token!=null;
+		
+		String p = token.getToken();
```
bitwiseman: Are there more checks that can be done here?  Ensure the list passed above is in the expected values of the returned `GHAuthorization` and such?

src/main/java/org/kohsuke/github/Requester.java
```diff
@@ -758,8 +758,11 @@ private InputStream wrapStream(InputStream in) throws IOException {
                 IOUtils.closeQuietly(es);
             }
         }
-        if (responseCode == HttpURLConnection.HTTP_UNAUTHORIZED) // 401 / Unauthorized == bad creds
-            throw e;
+        if (responseCode == HttpURLConnection.HTTP_UNAUTHORIZED) // 401 Unauthorized == bad creds
+            if(uc.getHeaderField("X-GitHub-OTP") != null)
```
bitwiseman: Add some comments about why this code path exists.

src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -671,7 +671,31 @@ public GHAuthorization createToken(Collection<String> scope, String note, String
...
+     * 
+     * Once OTP is received, call this token request
+     *
+     * The token created can be then used for {@link GitHub#connectUsingOAuth(String)} in the future.
+     *
+     * @see <a href="http://developer.github.com/v3/oauth/#create-a-new-authorization">Documentation</a>
+     */
+    public GHAuthorization createToken(Collection<String> scope, String note, String noteUrl, OTPReceiverLambda OTP) throws IOException{
```
bitwiseman: Add JavaDoc for the `@param`s, @return` and `@since`. 

And just a thought: 
```suggestion
    public GHAuthorization createToken(Collection<String> scope, String note, String noteUrl, Supplier<String> otpSupplier) throws IOException{
```

madhephaestus:
@bitwiseman I think I have incorporated all of your notes :)

src/main/java/org/kohsuke/github/GHOTPRequiredException.java
madhephaestus: done

src/main/java/org/kohsuke/github/Requester.java
madhephaestus: done

src/test/java/org/kohsuke/github/Github2faTest.java
madhephaestus: I check that each requested scope is returned

src/main/java/org/kohsuke/github/OTPReceiverLambda.java
madhephaestus: yup, i wasn't aware of the existing interface, changed.

src/test/java/org/kohsuke/github/Github2faTest.java
madhephaestus: fixed

src/main/java/org/kohsuke/github/GitHub.java
madhephaestus: yeah, i made this change, thanks for the heads up the generic existed!

madhephaestus:
Awesome!! What release will this feature be availible in?

bitwiseman:
@madhephaestus 
1.100.  I'd like to wait at least a few weeks before releasing, see if there is any fall out from the one year wait for 1.99.

madhephaestus:
any chance a pre-release or snapshot could be published so i can get my pipeline working and tested?

bitwiseman:
I'll get to a release later this week.  In the meanwhile, you could try https://jitpack.io/ .

PR #619, opened 2019-11-18
Removed permission field in createTeam. It is deprecated in the API

asthinasthi:
# Description 
https://developer.github.com/v3/teams/#create-team had deprecated the `permission` field but the library still expects this field. https://github.com/github-api/github-api/blob/da11702f6815f658014f33ce724f0cd2e4501081/src/main/java/org/kohsuke/github/GHOrganization.java#L387 

This PR refactors the code to remove the permission field.

# Before submitting a PR:


bitwiseman:
I rebased your changes on master to make the PR cleaner.

Thanks for this PR it is a good catch.  However, removing `permissions` would be a breaking change, which we will be avoiding until we are ready to create v2.x of this library.  So, we can't just remove this field.  

Instead you can mark the existing methods that have `permissions` as `@Deprecated` (with proper JavaDoc `@deprecated` note and so on), and create a method that does not use `permissions`.  Also, we'll need tests for both with and without permissions.

```java
public GHTeam createTeam(String name, Collection<GHRepository> repositories) throws IOException {
     // ...
}
/**
 * ... 
 * @deprecated use {@link #createTeam(String, Collection) instead.
 */
@Deprecated   
public GHTeam createTeam(String name, Permission p, Collection<GHRepository> repositories) throws IOException {
    // ...
}
```

asthinasthi:
Agreed! It was a long shot to remove those fields. 

I didn't understand why there is a separate `github-api-test-org` Aren't all tests mocked? 

Can you please point to a doc about what is different between these two orgs or elaborate here?

asthinasthi:
@bitwiseman When is the next release planned?

bitwiseman:
Done v1.101 is out.

Issue #504, opened 2019-03-12
Ability to commit textContent directly to tree, without intermediate blobs


blacelle:
https://github.com/kohsuke/github-api/issues/360 introduces a way to commit multiple files in a single commit. The library API suggests it is possible to build a tree from text content without creating blob manually with `org.kohsuke.github.GHTreeBuilder.textEntry(String, String, boolean)`.

However, this fails with:

    {"message":"Must supply tree.sha or tree.content","documentation_url":"https://developer.github.com/v3/git/trees/#create-a-tree"}

spaceghost69:
Sha seems required despite docs...?

iHatebug:
I also tried this and hit the same issue, may i know if there's any work around to commit multiple files in a single commit?

bitwiseman:
Please help us in improve this project.  
#360 was committed before we added a framework that made adding tests easier.

@siordache 
I know it has been a while but as the submitter of the original change would you be willing to add some tests we could use as examples?

siordache:
@bitwiseman I will look into it in the next few days.

bitwiseman:
Fixed in #620

siordache:
@spaceghost69 @blacelle @iHatebug
Until the fix goes live, see [this gist](https://gist.github.com/siordache/f2a895569de6f866877174376cf777d0) for a workaround.

bitwiseman:
Live now in v1.100.  Thanks @siordache!

PR #620, opened 2019-11-20
fix and unit tests for issue #504

siordache:
It seems that GitHub no longer accepts the 'content' parameter in tree entries, so I deprecated the `textEntry()` method (and also `entry()` and `shaEntry()`). Now, the preferred way to add content to the tree is to use the new methods `add(String path, String content, boolean executable)` and `add(String path, byte[] content, boolean executable)`.

Unit tests are available in: `GHTreeBuilderTest.java`.


vinz243:
Hello I know this is late but why did you remove shaEntry ? How would I go about linking other trees in my tree

Issue #519, opened 2019-05-14
Add new event types 


bitwiseman:
See https://issues.jenkins-ci.org/browse/JENKINS-57430

GitHub has added new events since this library was created.  The new events result in an exception being thrown for webhooks: 

```
java.lang.IllegalArgumentException: No enum constant org.kohsuke.github.GHEvent.CHECK_RUN
```

I don't have a list of event that need to be added aside from CHECK_RUN and CHECK_SUITE, but I'm sure there are more.

thomas-riccardi:
Bump

Among other things this limits usage in Jenkins.

strass:
I am hitting this issue in Jenkins now too

bitwiseman:
@strass  @thomas-riccardi 
It is being worked on.

PR #622, opened 2019-11-22
[JENKINS-57430] Added some new GH Event Types

alexanderrtaylor:
Added some new GH Event types so that it does not throw errors

# Description 
Adding some GH event types that are parsed by the github API

Fixes Issue #519


src/main/java/org/kohsuke/github/GHCheckRun.java
```diff
@@ -0,0 +1,72 @@
...
+public class GHCheckRun extends GHObject {
+    GHRepository owner;
+    GitHub root;
+
+    private String status;
+    private String conclusion;
+    private String name;
+    private GHPullRequest[] pullRequests;
+
+    GHCheckRun wrap(GHRepository owner) {
```
alexanderrtaylor: @Nonnull

src/main/java/org/kohsuke/github/GHCheckRun.java
```diff
@@ -0,0 +1,72 @@
...
+    public String getConclusion() {
+        return conclusion;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    GHPullRequest[] getPullRequests() throws IOException {
+        if (pullRequests.length != 0) {
```
alexanderrtaylor: if (pullRequests != null && pullRequests.length != 0)

PR #623, opened 2019-11-24
Move array population to single iterator code path

bitwiseman:
This removes a second code path through pagination.


bitwiseman:
@PauloMigAlmeida  
Any thoughts on this?  I don't have a lot of folks to ask for code review, so you get the honors. 😄

src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -425,7 +425,7 @@ URL getApiURL(String tailApiUrl) throws IOException {
     }
 
     Requester retrieve() {
-        return new Requester(this).method("GET");
+        return new Requester(this);
```
bitwiseman: `Requester` defaults to `GET` now.

src/main/java/org/kohsuke/github/Requester.java
```diff
@@ -53,13 +52,13 @@
 import java.util.Locale;
 import java.util.Map;
 import java.util.NoSuchElementException;
+import java.util.Objects;
 import java.util.function.Consumer;
 import java.util.logging.Logger;
-import java.util.regex.Matcher;
```
bitwiseman: No more regex-ing needed.

src/main/java/org/kohsuke/github/Requester.java
```diff
@@ -281,9 +283,9 @@ public Requester with(String key, Object value) {
      * @return the requester
      */
     public Requester set(String key, Object value) {
-        for (Entry e : args) {
-            if (e.key.equals(key)) {
-                e.value = value;
+        for (int index = 0; index < args.size(); index++) {
```
bitwiseman: With `final` Entry fields, we now replace the instance rather than changing the values.

src/main/java/org/kohsuke/github/Requester.java
```diff
@@ -314,6 +316,56 @@ public Requester contentType(String contentType) {
...
+     * <p>
+     * Should only be used when passing a literal URL field from a GHObject, such as {@link GHContent#refresh()} or when
+     * needing to set query parameters on requests methods that don't usually have them, such as
+     * {@link GHRelease#uploadAsset(String, InputStream, String)}.
+     *
+     * @param urlOrPath
+     *            the content type
+     * @return the requester
+     */
+    Requester setRawUrlPath(String urlOrPath) {
```
bitwiseman: Eventually, we'll want to call this method directly where it is needed, but for now `withUrlPath` has logic to decide when to use this.

src/main/java/org/kohsuke/github/Requester.java
```diff
@@ -327,89 +379,80 @@ public Requester inBody() {
...
+     *
+     * @param <T>
+     *            the type parameter
+     * @param type
+     *            the type
+     * @return {@link Reader} that reads the response.
+     * @throws IOException
+     *             if the server returns 4xx/5xx responses.
+     */
+    public <T> T[] toArray(@Nonnull Class<T[]> type) throws IOException {
```
bitwiseman: Overloading `to()` doesn't work with `T[]` and `T`.  The compiler can't handle it.  Instead moved this logic to this new method.  

This results in cleaner and clearer code - there is no case where the caller doesn't already know whether they want an array or an instance returned so why call one method and then immediately have an `if` statement.

src/main/java/org/kohsuke/github/Requester.java
```diff
@@ -327,89 +379,80 @@ public Requester inBody() {
...
 
-            buildRequest();
+        String tailApiUrl = buildTailApiUrl(urlPath);
+        setupConnection(root.getApiURL(tailApiUrl));
 
+        while (true) {// loop while API rate limit is hit
             try {
-                T result = parse(type, instance);
-                if (type != null && type.isArray()) { // we might have to loop for pagination - done through recursion
-                    final String links = uc.getHeaderField("link");
```
bitwiseman: This same work as done by the `PagingIterator<T>`.

src/test/resources/org/kohsuke/github/AppTest/wiremock/notifications/mappings/notifications-10-a97934.json
```diff
@@ -2,7 +2,7 @@
   "id": "a979348d-c6be-4cb7-8877-7c42a6f013ae",
   "name": "notifications",
   "request": {
-    "url": "/notifications?all=true&page=9&all=true",
```
bitwiseman: And by switching to one consistent path we find little bugs like this.  Yay!

PauloMigAlmeida: :)

PauloMigAlmeida:
>@PauloMigAlmeida
Any thoughts on this? I don't have a lot of folks to ask for code review, so you get the honors. 😄

My pleasure ;)

@bitwiseman looks great! This should open the possibilities for a lot of improvements that we always wanted to do 🍾 

The only thing which I still struggling is the name of the `to` method. Now that is doesn't take the URL path anymore, it sounds a bit funny. For instance:

```java
GHHook hook = root.retrieve()
                  .method("POST")
                  .with("name", name)
                  .with("active", active)
                  .with("config", config)
                  .with("events", ea)
                  .withUrlPath(collection())
                  .to(clazz());
```

In the case above it looks like it's performing a `parseTo`/`fromJson` kind of thing, while in this other below:

```java
root.retrieve().method("PATCH").withUrlPath("/user/repository_invitations/" + id).to();
```

This looks like it's performing a `execute`/`sendRequest`.

Like any other developer, I too struggle with coming up with variable/method names 😄 but if you could think of a name for the former `to()` method that can read nicely in both scenarios then it's perfect 👍

bitwiseman:
@PauloMigAlmeida 
Okay, I'll go through and clean up all the redundant `method()` calls.  Thanks!

Naming: Agreed, it could be better.  The `to()` method should really be something like `send()`.

PR #627, opened 2019-11-27
Fixed CacheControl package name

alecharp:
# Description
The `okhttp3.OkHttpConnector` class is still referencing the `okhttp.CacheControl`. However, this class is not part of OkHttp 3.x or 4.x. This creates a `ClassNotFoundException` at runtime on projects creating a `okhttp3.OkHttpConnector` with a `Cache`.

# Before submitting a PR:


bitwiseman:
No way to detect this aside from better code reviews or having a test project that only has okhttp3 and not okhttp. 😞 Thanks!

alecharp:
I figured it was a "copy-paste" issue, no big deal. Imports are not the first thing I think of and here I got angry because the class was available in the classpath. But not the _correct_ class..

bitwiseman:
Published as v1.101.

PR #633, opened 2019-12-04
Add shading of dependencies

### LINKED ISSUE #630: Shade the common dependencies in the library
Currently GitHub API declares several dependencies, including okhttp, commons-lang, commons-io, etc. These dependencies were updated in 1.99 by Dependabot, and it leads to risks when using the library in components which already include older versions of the libraries. 

One of the examples: Jenkins Core and GitHub API Plugin which bundles the library. https://github.com/jenkinsci/github-api-plugin/pull/24 updated the version there while Jenkins static checkers ignored it due to the exclusions in POM.xml: https://github.com/jenkinsci/github-api-plugin/blob/master/pom.xml#L77-L80 . Basically such approach may explode at any moment.

There are two possible solutions:

* Revert library updates in 1.99, pin the library versions
* Shade the dependencies so that GitHub API becomes more portable. I would recommend it though it may require a significant effort

bitwiseman:
Fixes #630


dependency-reduced-pom.xml
```diff
@@ -0,0 +1,585 @@
...
+            <phase>package</phase>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <artifactSet>
+                <excludes>
+                  <exclude>com.infradna.tool:bridge-method-annotation</exclude>
+                  <exclude>org.jenkins-ci:annotation-indexer</exclude>
+                  <exclude>com.squareup.*:*</exclude>
```
bitwiseman: Optional dependency.

dependency-reduced-pom.xml
```diff
@@ -0,0 +1,585 @@
...
+                  <exclude>org.jetbrains*:*</exclude>
+                  <exclude>com.github.spotbugs:*</exclude>
+                  <exclude>com.google.code.findbugs:*</exclude>
```
bitwiseman: compile time dependencies.

bitwiseman:
Hm, thinking shading should only be done in ci and release maybe?

bitwiseman:
@oleg-nenashev  Can you comment on this when you have time?

oleg-nenashev:
I need to test it manually to comment. What is your plan for `dependency-reduced-pom.xml` ?

bitwiseman:
@oleg-nenashev 
`dependency-reduced-pom.xml` get generated during build and is what is published on release.

bitwiseman:
Ah, looks like it should be ignored. https://github.com/cloudera/director-google-plugin/issues/45

bitwiseman:
@oleg-nenashev Updated.

PR #634, opened 2019-12-04
Task/minimum coverage

bitwiseman:
Adding "touched it" coverage"


src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -1935,6 +1935,9 @@ public GHHook createWebHook(URL url) throws IOException {
     @SuppressFBWarnings(value = "DMI_COLLECTION_OF_URLS",
             justification = "It causes a performance degradation, but we have already exposed it to the API")
     public Set<URL> getPostCommitHooks() {
+        if (postCommitHooks == null) {
+
```
bitwiseman: Hm, this seems wrong.

PR #642, opened 2019-12-11
Post only non-null values in request body for createTree

asthinasthi:
# Description 
Github release 2.18+ doesn't accept `sha:null` in the request body. This change fixes that. More details here: https://github.com/github-api/github-api/issues/636

# Before submitting a PR:


bitwiseman:
@asthinasthi 
This is really good.  Thanks!  This is the behavior for GHE.  Is it the same for GitHub.com?

asthinasthi:
Actually I see that we have deprecated this https://github.com/github-api/github-api/blob/d1507f26668950508e0bf242c34cdb599003991a/src/main/java/org/kohsuke/github/GHTreeBuilder.java#L111
in the latest release (I was using 1.93).

So when we use the recommended 
https://github.com/github-api/github-api/blob/d1507f26668950508e0bf242c34cdb599003991a/src/main/java/org/kohsuke/github/GHTreeBuilder.java#L129

```

    public GHTreeBuilder add(String path, byte[] content, boolean executable) {
        try {
            String dataSha = repo.createBlob().binaryContent(content).create().getSha();
            return shaEntry(path, dataSha, executable);
        } catch (IOException e) {
            throw new GHException("Cannot create binary content of '" + path + "'", e);
        }
    }
```
We create a sha & then use it in the API. We never send a null sha field.

Also, we have deprecated `shaEntry(path, dataSha, executable)` so I think we are good.

bitwiseman:
@asthinasthi 
Um, so, do you want to close this or still merge it?

asthinasthi:
We can close this. As long as I stay away from deprecated methods I should be fine.

Xcelled:
Can this be reopened and merged? I just got bit by this issue in deployed production code after a GH:E upgrade and it took me forever to track it down.

src/main/java/org/kohsuke/github/GHTreeBuilder.java
```diff
@@ -16,8 +18,10 @@
 
     private final List<TreeEntry> treeEntries = new ArrayList<TreeEntry>();
 
+    @JsonInclude(Include.NON_NULL)
```
bitwiseman:
```suggestion
    // Issue #636: Create Tree no longer accepts null value in sha field
    @JsonInclude(Include.NON_NULL)
```

Issue #126, opened 2014-09-05
github-api does in not correctly distinguish between user and organisation ownership


msperisen:
owners are always returned as GHUsers instead of GHOrganizations or GHUsers. Many github api calls return users and organisations alike and the distinction is made based on the type field. Making GHPerson reflect that:

``` java
@JsonTypeInfo(use=JsonTypeInfo.Id.NAME,
        include=JsonTypeInfo.As.PROPERTY,
        property="type")
@JsonSubTypes({
        @JsonSubTypes.Type(value=GHUser.class, name="User"),
        @JsonSubTypes.Type(value=GHOrganization.class, name="Organization"),
})
public abstract class GHPerson {
```

will break a lot of owner related githup-api calls

kohsuke:
Thanks for the Jackson mapping fragment.

I'm afraid at this point this is unfixable as it breaks too many existing method signatures.

davidkarlsen:
Could at least the "type" attribute be exposed in the API so it can be handled by the api-user? Also the site_admin property would be valuable (all-in-all all properties exposed would make sense to have on the object-model).

bitwiseman:
@davidkarlsen 
PR's are welcome.  Adding more information is not an API breaking-change.

Haarolean:
This is also a big problem when using different authentication methods (app authentication with a JWT token vs app installation authentication). All the methods are available within the same class and sometimes you just get "bad credentials" response.

Issue #589, opened 2019-11-01
Final instances vs Incomplete instances vs Update-in-place


bitwiseman:
#495 and #496 raise an interesting point. 

This project currently tends to treat instances of data objects as "final" - you request an Issue and the instance you requested never changes after that.  That instance is a stable snapshot and can be used as a key in a HashSet or Map without fear of it changing.   

We also have action and update methods on these objects. In general, these methods update the state of the server-side object the instance is based on, but do not change the data in the instance.  You call `close()` or `delete()` or one of the `set*` methods and a request is sent to change the object on the server, but the data in the local instance doesn't update to match.  This is consistent with the "final instances" model but has some problems.  There's no batching of updates (each action is another query) and since the local instance is out of date, updates can end up overwriting each other as in #495.  The only current workaround is send another request to get a fresh instance for each update.  😱 

The `Builder/Updater` pattern added to some objects addresses this by batching updates together and then returning a new instance matching those updates.  This solves the problem of stale data but results in verbose `item.update().title('New Title').update()` syntax for single field changes. It does partially unblock update-in-place as `item = item.update().title('New Title').update()`.   

Finally, some objects can be incomplete when first created - either due to their source (partial information provided as part of creating another object) or due to the target only updating on request (pull request mergability is only updated after the first time PR is queried after a new commit is pushed).  For these cases, we currently update in place.  Do we want to force users to get new instances instead?  

There's no one right way to do this, but we should look at how to make this behavior consistent across all objects.   

My opinions off hand about this: 
* `set*` method naming is misleading - best practice in my understanding is to make `get*` and `set*` methods local and lightweight.  Alternative would be to have `update*` methods that actually send updates to the server. 
* Create states for the objects and allow the parent objects and the `GitHub` object to create objects in those state.  Rather than having updaters for objects, users would get a mutable instance. Possible states for data objects:  immutable, immediate mutable, batch mutable.   Not sure how the incomplete objects figure into this yet.

bitwiseman:
A thought I had today.  `update*` methods  could return a new instance with the updated data, while `update()` would start a batch.  

So: 
```
GHLabel label = getLabel(myLabelName);
// two network calls, each returning a new instance
label = label.updateName("New Name").updateColor(newColor);

// one network call, returns new instance
label = label.update().name("New Name").color(newColor).apply();

// one network call, updates in place?  Maybe not. 
// I'd really rather have instances be effectively final so their hashCode and equality are stable. 
[...]

PR #644, opened 2019-12-13
Add type and site_admin fields to GHPerson

martinvanzijl:
# Description 
Helps with #126. This adds the `type` and `site_admin` field from the API to `GHPerson`. It does not completely solve the issue, but at least you can use `getType()` to see whether the owner is actually a User or an Organization.

NOTE: I'm not sure I need the calls to `populate()`; it worked fine without them in my unit tests. How can I be sure?


bitwiseman:
@martinvanzijl 
Short answer: I don't know but what you have done is a reasonable start.  

Slightly longer answer: `populate()` is needed for when the User or Organization is created from a response that has only partial information - but I don't know which fields those are.  😭 

Non-answer information: I started to add the `Refreshable` interface that will work without knowing which fields are missing from partial responses. However,  it isn't applied generally yet and it is basically the same as `populate()` - which I'm not sure it is the right way to go long term  (see #589 for some background).

PR #647, opened 2019-12-19
Labels with description are not preview anymore

halkeye:
# Description 
Labels with descriptions do not seem to be preview apis anymore - https://developer.github.com/v3/issues/labels/#create-a-label so don't mark them as such

# Before submitting a PR:

- [ ] Run `mvn -P ci install site ` locally. This may reformat your code, commit those changes. If this command doesn't succeed, your change will not pass CI.

### DIFF
--- a/src/main/java/org/kohsuke/github/GHLabel.java
+++ b/src/main/java/org/kohsuke/github/GHLabel.java
@@ -6,8 +6,6 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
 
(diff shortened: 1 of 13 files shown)


bitwiseman:
@halkeye Do you have time to fix the test failure or should I do it?

halkeye:
If you know how then better for you. But I can spend some time if need be.

Looks like it's on master too, and it's an extra key in the object but that's as far as I got

bitwiseman:
Ah, interesting.  How did that get in there? 😞  
Okay, I'll fix it.

PR #648, opened 2019-12-20
Enforce 'Accept' request header values

bitwiseman:
# Description
The 'Accept' header on requests can change what data is returned, so we need to
only match requests with the correct value.

# Before submitting a PR:

- [x] Run `mvn -P ci install site ` locally. This may reformat your code, commit those changes. If this command doesn't succeed, your change will not pass CI.

### DIFF
--- a/src/test/resources/org/kohsuke/github/AppTest/wiremock/blob/mappings/repos_github-api_github-api-2-db8eac.json
+++ b/src/test/resources/org/kohsuke/github/AppTest/wiremock/blob/mappings/repos_github-api_github-api-2-db8eac.json
@@ -3,7 +3,12 @@
   "name": "repos_github-api_github-api",
   "request": {
     "url": "/repos/github-api/github-api",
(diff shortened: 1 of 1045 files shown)


src/test/resources/org/kohsuke/github/AppTest/wiremock/blob/mappings/repos_github-api_github-api-2-db8eac.json
```diff
@@ -3,7 +3,12 @@
   "name": "repos_github-api_github-api",
   "request": {
     "url": "/repos/github-api/github-api",
-    "method": "GET"
+    "method": "GET",
+    "headers": {
+      "Accept": {
+        "equalTo": "text/html, image/gif, image/jpeg, *; q=.2, */*; q=.2"
```
bitwiseman: It sucks to have to add this to almost every test mapping file, but it will be done automatically in the future.

PR #649, opened 2019-12-20
Add ability to update Github labels

halkeye:
# Description 
* Add ghlabels.update
* extend ghobject for standardness (might be breaking change? but since its extend i suspect its fine)
* Change getURL signature to match ghobject

- [X] Run `mvn -P ci install site ` locally. This may reformat your code, commit those changes. If this command doesn't succeed, your change will not pass CI.


halkeye:
I didn't. I can add set name.

I think they should have fetchinto and the tests still right?

halkeye:
https://github.com/github-api/github-api/blob/6bc617c488e96ea4e8040b54754f13ec68356bbe/src/test/java/org/kohsuke/github/AppTest.java#L814-L816

Did i accidentally fix this?

src/main/java/org/kohsuke/github/GHLabel.java
```diff
@@ -13,7 +15,7 @@
  * @see GHIssue#getLabels() GHIssue#getLabels()
  * @see GHRepository#listLabels() GHRepository#listLabels()
  */
-public class GHLabel {
+public class GHLabel extends GHObject {
```
bitwiseman: Does this actually have all the fields from GHObject? I don't think it does.

bitwiseman:
@halkeye 
Sorry for delay, I was on vacation. 

> Did i accidentally fix this?

Yes, you kind of did - by using `fetchInto` instead of `send`.  But in doing so you've also changed the behavior of the class overall. 

If someone is depending on GHLabel instances to be immutable, they will get a surprise after this change.  I'm not saying it is wrong, it is just unclear what the right behavior for the library overall should be.  See #589. 

Look I'm not trying to make your life hard, I'm glad to have more contributors. I'm happy to make time to chat in real-time, so we can get this resolved and merged.

bitwiseman:
@halkeye 
You around?
I'm going to go ahead cut the next release. I'm be happy to do another one when we hammer this out as well.

bitwiseman:
@halkeye  Please take a look at #724 .

PR #665, opened 2020-01-14
Workaround for `If-Modified-Since` HTTP request header causing cache corruption

bitwiseman:
# Overview 
This PR shows a low level repro of an issue that has been plaguing Jenkins users for at least the last year.

EDIT: Similar to https://github.com/isaacs/github/issues/692 but not the same

# Key Takeaways
* The root cause of this issue is NOT Jenkins.  The issue is caused by a bug in OkHttp caching, GitHub server software, or the interaction between them. 
    * In the short-term Jenkins will still need to make changes to mitigate this.  What these steps will be is unknown.
* This is an issue that can effect any GitHub query that returns a 404 why caching is turned on.
    * The existing work-around addresses the scenario that seems to hit Jenkins most often but it is possible there are any number of other places where this issue causes problem that we have not been able to reliably track down.

# Basic Repro Steps
This basic process can be used for many queries that return 404.   Using Ref creation as an example:

1. Query CACHED connection: request git ref "refs/heads/does_not_exist" - File not found (expected)
2. Create git ref "refs/heads/does_not_exist"
3. Query CACHED connection: request git ref "refs/heads/does_not_exist" - File not found due erroneous GitHub 304 response (fail)


src/test/java/org/kohsuke/github/extras/okhttp3/GitHubCachingTest.java
```diff
@@ -0,0 +1,334 @@
...
+        // Expected: success
+        // Actual: still GHFileNotFound due to caching: GitHub incorrectly returns 304
+        // even though contents of the ref have changed.
+        //
+        // There source of this issue seems to be that 404's do not return an ETAG,
+        // so the cache falls back to using "If-Modified-Since" which is erroneously returns a 304.
+        //
+        // NOTE: This is even worse than you might think: 404 responses don't return an ETAG, but 304 responses do.
+        //
+        // Due erroneous 304 returned from "If-Modified-Since", the ETAG returned by the first 304
```
bitwiseman: This refers to https://github.com/github-api/github-api/pull/665/files#diff-02b86edbc1ffd10d08e7049e7a842fc8R9.

aglarendil:
@bitwiseman I would suggest looking into the logic of the github branch source plugin as well. The problem here as per https://issues.jenkins-ci.org/browse/JENKINS-60353 is that Jenkins checks the existence of the Jenkinsfile for the 'MERGE' head of a non-mergeable request, the one that does not exist _a priori_ and then consequently marks the pr as dead/non-buildable.

bitwiseman:
@aglarendil  
Yup, there is less than ideal behavior in the plugin (that's where my analysis to for this example came from).  However, they would not be an issue if GitHub weren't silently behaving horribly.

PR #675, opened 2020-01-23
Added ability to specify permissions for collaborators

jimmysombrero:
Modified the addCollaborators methods to require a GHOrganization.Permission parameter in order allow the API to add external collaborators with varying levels of permissions.  Added a second modifyCollaborators method with a signature that accepts a GHOrganization.Permission parameter.


jimmysombrero:
Running mvn formatter:format refuses to format the file

bitwiseman:
Hm, not sure why the formatting isn't happening.  Have you tried `mvn clean install`?

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -535,9 +535,12 @@ public GHRelease getLatestRelease() throws IOException {
      * @return the owner name
      */
     public String getOwnerName() {
-        // consistency of the GitHub API is super... some serialized forms of GHRepository populate
-        // a full GHUser while others populate only the owner and email. This later form is super helpful
-        // in putting the login in owner.name not owner.login... thankfully we can easily identify this
+        // consistency of the GitHub API is super... some serialized forms of
```
bitwiseman: Take a minute to fix this formatting please.

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -750,7 +753,7 @@ public int getSize() {
 
     /**
      * Lists all
-     * <a href="https://help.github.com/articles/assigning-issues-and-pull-requests-to-other-github-users/">the
+     * <a href= "https://help.github.com/articles/assigning-issues-and-pull-requests-to-other-github-users/">the
```
bitwiseman:
```suggestion
     * <a href="https://help.github.com/articles/assigning-issues-and-pull-requests-to-other-github-users/">the
```

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -840,23 +843,27 @@ public GHPermissionType getPermission(GHUser u) throws IOException {
...
      * @param users
      *            the users
+     * @param perm
+     *            the permission level
      * @throws IOException
      *             the io exception
      */
-    public void addCollaborators(GHUser... users) throws IOException {
-        addCollaborators(asList(users));
+    public void addCollaborators(GHOrganization.Permission perm, GHUser... users) throws IOException {
```
bitwiseman: It's fine to add new methods with new parameters, but we need to keep the original methods as well.  

If you believe the old ones are no longer valid you can mark them as deprecated.

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -1944,7 +1963,8 @@ public GHHook createWebHook(URL url) throws IOException {
     // * Retrieves all the pull requests.
     // */
     // public List<GHPullRequest> getPullRequests() throws IOException {
-    // return root.retrieveWithAuth("/pulls/"+owner+'/'+name,JsonPullRequests.class).wrap(root);
+    // return
```
bitwiseman:
```suggestion
    // return return root.retrieveWithAuth("/pulls/"+owner+'/'+name,JsonPullRequests.class).wrap(root);
```

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -1944,7 +1963,8 @@ public GHHook createWebHook(URL url) throws IOException {
...
+    // root.retrieveWithAuth("/pulls/"+owner+'/'+name,JsonPullRequests.class).wrap(root);
```
bitwiseman:
```suggestion
```

bitwiseman:
@jimmysombrero 
Are you working on windows?  Just wondering. 
Also do you have time to write tests?

jimmysombrero:
> 
> 
> @jimmysombrero
> Are you working on windows? Just wondering.
> Also do you have time to write tests?

I was working in Windows, but switched to my Mac, which fixed the formatting issues I was having.  My weekend project for this weekend is to add the missing methods back and write tests for the new methods.  I was trying to wrap my head around how the json configuration files for WireMock are created then decided to follow the instructions in CONTRIBUTING.md and see what happens.

bitwiseman:
If following the CONTRIBUTING doesn't basically "just work", I'm committed to fixing it.  That's what it's there for. 😄

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -843,6 +857,7 @@ public GHPermissionType getPermission(GHUser u) throws IOException {
      * @throws IOException
      *             the io exception
      */
+    @Deprecated
```
bitwiseman: Now that I'm looking at the api documentation for "[Add user as collaborator](https://developer.github.com/v3/repos/collaborators/#add-user-as-a-collaborator) it doesn't look like permissions are a required in all cases. So, we probably don't want the base methods deprecated. 
```suggestion
```

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -855,10 +870,25 @@ public void addCollaborators(GHUser... users) throws IOException {
      * @throws IOException
      *             the io exception
      */
-    public void addCollaborators(Collection<GHUser> users) throws IOException {
+    @Deprecated
+    public void addCollaborators(List<GHUser> users) throws IOException {
```
bitwiseman: This would be an API change.
```suggestion
    public void addCollaborators(Collection<GHUser> users) throws IOException {
```

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -855,10 +870,25 @@ public void addCollaborators(GHUser... users) throws IOException {
      * @throws IOException
      *             the io exception
      */
-    public void addCollaborators(Collection<GHUser> users) throws IOException {
+    @Deprecated
```
bitwiseman:
```suggestion
```
Same as above.

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -889,6 +919,18 @@ private void modifyCollaborators(Collection<GHUser> users, String method) throws
         }
     }
 
+    private void modifyCollaborators(GHOrganization.Permission perm, Collection<GHUser> users, String method)
```
bitwiseman: Can we make all versions call one method?  Pass `null` for permission if it isn't used? 
```suggestion
    private void modifyCollaborators(@Nonnull Collection<GHUser> users, @Nonnull String method, @CheckForNull GHOrganization.Permission perm)
```

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -855,10 +870,25 @@ public void addCollaborators(GHUser... users) throws IOException {
...
+     * Add collaborators.
+     *
+     * @param users
+     *            the users
+     * @param perm
+     *            the permission level
+     * @throws IOException
+     *             the io exception
+     */
+    public void addCollaborators(GHOrganization.Permission perm, Collection<GHUser> users) throws IOException {
```
bitwiseman:
```
    public void addCollaborators(Collection<GHUser> users, GHOrganization.Permission perm) throws IOException {
```
Since `perm` is optional, from an API design view I want it to be on the end, but then the `(GHOrganization.Permission perm, GHUser... users)` signature looks out of place.  

What's your opinion? Do we just take this as it is or reverse this and remove the other signature?

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -835,6 +835,20 @@ public GHPermissionType getPermission(GHUser u) throws IOException {
...
+     * Add collaborators.
+     *
+     * @param users
+     *            the users
+     * @param perm
+     *            the permission level
+     * @throws IOException
+     *             the io exception
+     */
+    public void addCollaborators(GHOrganization.Permission perm, GHUser... users) throws IOException {
```
bitwiseman:
```suggestion
    public void addCollaborators(GHOrganization.Permission permission, GHUser... users) throws IOException {
```

src/main/java/org/kohsuke/github/GHRepository.java
jimmysombrero: That's a tough question.  I agree with the API design view of having permissions at the end, but if we get rid of the other signature then anyone that only needs to add a single collaborator will need to create and populate a collection.  Since there is already a method that takes a collection, what do you think about changing the signature to only accept a single user?

`public void addCollaborator(GHUser user, GHOrganization.Permission permission)`

bitwiseman: Or we just accept that one signature is out of place but convenient and appropriate.

jimmysombrero: Sounds good.

PR #678, opened 2020-01-24
Harden 404 retry against exceptions

bitwiseman:
# Description 
If the 404 retry work around gets an exception while trying to determine if it should retry, that means it should not retry.  It should let execution continue on the where such exceptions are generally handled. 

This is a rare situation that is mostly added to deal with some strange mocking tests in dependent projects.


bitwiseman:
Actually #674 made me make changes to make this obsolete.

Issue #679, opened 2020-01-24
javax.net.ssl.SSLHandshakeException when using github API to fetch organization value.


abhinavg27:
This is intermittent issue. Sometime it works and most of the time it fails :

github = GitHub.connectToEnterpriseWithOAuth(appConfig.getGitHubBaseURL(), appConfig.getGitHubUserName(),
              appConfig.getGitHubUserToken());

gitOrg = github.getOrganization(gitHubBaseOrg);

and when trying to get organization values this is what I am getting sometimes:

Caused by: org.kohsuke.github.HttpException: Server returned HTTP response code: -1, message: 'null' for URL: https://github.home....-inc.net/api/v3/orgs/...
	at org.kohsuke.github.Requester.parse(Requester.java:638)
	at org.kohsuke.github.Requester.parse(Requester.java:599)
	at org.kohsuke.github.Requester._to(Requester.java:277)
	at org.kohsuke.github.Requester.to(Requester.java:239)
	at org.kohsuke.github.GitHub.getOrganization(GitHub.java:432)
Caused by: javax.net.ssl.SSLHandshakeException: sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
	at sun.security.ssl.Alerts.getSSLException(Alerts.java:192)
	at sun.security.ssl.SSLSocketImpl.fatal(SSLSocketImpl.java:1917)
	at sun.security.ssl.Handshaker.fatalSE(Handshaker.java:301)
	at sun.security.ssl.Handshaker.fatalSE(Handshaker.java:295)
	at sun.security.ssl.ClientHandshaker.serverCertificate(ClientHandshaker.java:1369)
	at sun.security.ssl.ClientHandshaker.processMessage(ClientHandshaker.java:156)
	at sun.security.ssl.Handshaker.processLoop(Handshaker.java:925)
	at sun.security.ssl.Handshaker.process_record(Handshaker.java:860)
	at sun.security.ssl.SSLSocketImpl.readRecord(SSLSocketImpl.java:1043)
	at sun.security.ssl.SSLSocketImpl.performInitialHandshake(SSLSocketImpl.java:1343)
	at sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:1371)
	at sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:1355)
[...]

PR #680, opened 2020-01-25
Handle SSLHandshakeException with connection retry

bitwiseman:
# Description 
Fixes for #679 
Also adds tests to cover more retry and exception scenarios.


src/main/java/org/kohsuke/github/Requester.java
```diff
@@ -558,6 +563,33 @@ private boolean retrySocketException(IOException e, int retries) throws IOExcept
         return false;
     }
 
+    private boolean retryInvalidCached404Response(int responseCode, int retries) throws IOException {
```
bitwiseman: This just a move.

PR #694, opened 2020-02-06
Streamline fetch and retry process

bitwiseman:
# Description 
Streamlining and clean up of fetching.  Not quite a refactoring but close. 
Rate limit changed to be updated for each retry. 
Rate and Abuse handlers are part of the retry process - they will not loop indefinitely.  They will also be detected without requiring an exception.  This means that `fetchHttpStatusCode` no longer simply returns forbidden status when rate limit is hit.


src/main/java/org/kohsuke/github/Requester.java
```diff
@@ -497,73 +497,81 @@ public InputStream fetchStream() throws IOException {
     }
 
     private <T> T _fetch(String tailApiUrl, URL url, SupplierThrows<T, IOException> supplier) throws IOException {
-        while (true) {// loop while API rate limit is hit
+        int responseCode = -1;
+        String responseMessage = null;
+
```
bitwiseman: It is better to view this in split diff view.  Two methods got squashed down and blended.

src/main/java/org/kohsuke/github/Requester.java
```diff
@@ -497,73 +497,81 @@ public InputStream fetchStream() throws IOException {
...
+        int retries = CONNECTION_ERROR_RETRIES;
+
+        do {
+            // if we fail to create a connection we do not retry and we do not wrap
+            uc = null;
```
bitwiseman: This ensures that on retry we don't somehow accidentally use the connection from the previous loop.

src/main/java/org/kohsuke/github/Requester.java
```diff
@@ -497,73 +497,81 @@ public InputStream fetchStream() throws IOException {
...
 
             try {
-                return _fetchOrRetry(supplier, CONNECTION_ERROR_RETRIES);
-            } catch (IOException e) {
-                handleApiError(e);
-            } finally {
+                // This is where the request is sent and response is processing starts
+                responseCode = uc.getResponseCode();
+                responseMessage = uc.getResponseMessage();
                 noteRateLimit(tailApiUrl);
```
bitwiseman: We note the rate limit change as soon as we have confidence that we have enough information to do so.
Better for race conditions.

src/main/java/org/kohsuke/github/Requester.java
```diff
@@ -497,73 +497,81 @@ public InputStream fetchStream() throws IOException {
...
+                if (!(isInvalidCached404Response(responseCode) || isRateLimitResponse(responseCode)
```
bitwiseman: If we have an error condition that we can detect early, skip calling the lambda.

src/main/java/org/kohsuke/github/Requester.java
```diff
@@ -497,73 +497,89 @@ public InputStream fetchStream() throws IOException {
...
+                // This is where the request is sent and response is processing starts
+                responseCode = uc.getResponseCode();
+                responseMessage = uc.getResponseMessage();
                 noteRateLimit(tailApiUrl);
+
+                // for this workaround, we can retry now
+                if (isInvalidCached404Response(responseCode)) {
+                    continue;
+                }
+                if (!(isRateLimitResponse(responseCode) || isAbuseLimitResponse(responseCode))) {
```
bitwiseman: If we have an error condition that we can detect early, skip calling the lambda.

src/main/java/org/kohsuke/github/Requester.java
```diff
@@ -497,73 +497,89 @@ public InputStream fetchStream() throws IOException {
...
+                    return supplier.get();
+                }
+            } catch (IOException e) {
+                // For transient errors, retry
+                if (handledTransientConnectionError(e, url, retries)) {
+                    continue;
+                }
+
+                throw interpretApiError(e, responseCode, responseMessage, url, retries);
```
bitwiseman: If the error is not transient, process it if needed and throw.

src/main/java/org/kohsuke/github/Requester.java
```diff
@@ -1023,57 +1039,36 @@ private InputStream wrapStream(InputStream in) throws IOException {
     /**
      * Handle API error by either throwing it or by returning normally to retry.
      */
-    void handleApiError(IOException e) throws IOException {
-        int responseCode;
-        try {
```
bitwiseman: This case is handled in the main `_fetch` method now.  We pass in the best known values and don't have to guard against bad state here.

src/main/java/org/kohsuke/github/Requester.java
```diff
@@ -497,73 +497,101 @@ public InputStream fetchStream() throws IOException {
...
-            // java.net.URLConnection handles 404 exception as FileNotFoundException,
-            // don't wrap exception in HttpException to preserve backward compatibility
-            throw e;
-        } catch (IOException e) {
+        throw new GHIOException("Ran out of retries for URL: " + url.toString());
+    }
 
-            if (!retryConnectionError(e, retries)) {
-                throw new HttpException(responseCode, responseMessage, uc.getURL(), e);
+    private void detectOTPRequired(int responseCode) throws GHIOException {
```
bitwiseman: Moved from `handleApiError`.

PR #697, opened 2020-02-13
Refactor `Requester` into multiple smaller classes

bitwiseman:
# Description 
Refactoring from `GitHub` and `Requester` being monolithic centers of work to having the work split among smaller classes.  `GitHub` is still the root of interactions, but holds a reference to a `GitHubClient` that has all the connection and communication. `Requester` still remains as a thin helper that will be removed in a future change.

# Improvements 
* Abstraction/encapsulation of  `HttpUrlConnection` interactions in `GitHubHttpUrlConnectionClient`
* Builders and mostly immutable instances - better thread safety and clarity
* Client/Request/Response model in line with modern Http frameworks
* `PagedIterable` is the sole entry point for retrieval of all item arrays, lists, or iterables. 
* `*Iterable` and `*Iterator` classes encapsulate all the work done to retrieve pages of items

This change encapsulates most of the dependencies on `HttpUrlConnection` as much as possible, to allow moving to other more modern frameworks at some point.  This also separates the concerns more clearly. 

There's still some work to be done here, but this change is more than enough for now.


src/main/java/org/kohsuke/github/Requester.java
```diff
@@ -23,368 +23,24 @@
...
-
-import static java.util.Arrays.asList;
-import static java.util.logging.Level.*;
-import static org.apache.commons.lang3.StringUtils.defaultString;
-import static org.kohsuke.github.GitHub.MAPPER;
-import static org.kohsuke.github.GitHub.connect;
 
 /**
- * A builder pattern for making HTTP call and parsing its output.
+ * A thin helper for {@link GitHubRequest.Builder} that includes {@link GitHubClient}.
```
bitwiseman: This is the really the summary of this change: Refactoring from `Requester` being monolithic center of work to having the work split between `GitHubClient`, `GitHubRequest`, and `GitHubResponse` among others.

src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -535,17 +403,21 @@ public GHRateLimit rateLimit() throws IOException {
      *             the io exception
      */
     @WithBridgeMethods(GHUser.class)
-    public GHMyself getMyself() throws IOException {
-        requireCredential();
+    GHMyself getMyself() throws IOException {
```
bitwiseman:
```suggestion
    public GHMyself getMyself() throws IOException {
```

src/main/java/org/kohsuke/github/GitHubRequest.java
```diff
@@ -0,0 +1,651 @@
...
+    public static Builder<?> newBuilder() {
+        return new Builder<>();
+    }
+
+    /**
+     * Gets the final GitHub API URL.
+     */
+    @Nonnull
+    static URL getApiURL(String apiUrl, String tailApiUrl) throws MalformedURLException {
+        if (tailApiUrl.startsWith("/")) {
```
kshultzCB: Is there a circumstance where `tailApiUrl` will _not_ start with a "/"? I didn't see any tests that behave this way, so just wondering how we'd run up against this, and what it looks like when we do.

src/main/java/org/kohsuke/github/GitHubRequest.java
```diff
@@ -0,0 +1,651 @@
...
+     *
+     * @param en
+     *            Enum to be transformed
+     * @return a String containing the value of a Github constant
+     */
+    static String transformEnum(Enum<?> en) {
+        // by convention Java constant names are upper cases, but github uses
+        // lower-case constants. GitHub also uses '-', which in Java we always
+        // replace with '_'
+        return en.toString().toLowerCase(Locale.ENGLISH).replace('_', '-');
```
kshultzCB: Comments like this are super helpful. :)

Are there other places where we might need to do such a conversion? I only see one instance of `replace('_', '-')` in the whole plugin. Might be fine, it just surprised me that there aren't more.

src/main/java/org/kohsuke/github/GitHubRequest.java
```diff
@@ -0,0 +1,651 @@
...
+         *            the value
+         * @return the request builder
+         */
+        public B with(String key, Object value) {
+            if (value != null) {
+                args.add(new Entry(key, value));
+            }
+            return (B) this;
+        }
+
```
kshultzCB: This is a somewhat arbitrary place for me to put this question, but it's as good as any. There are a lot of `with(thing, thing)` methods here with different parameters. Why is that? I mean, I'm sure there's a perfectly good reason for it, I just am having trouble putting it together on my own.

src/main/java/org/kohsuke/github/PagedIterator.java
```diff
@@ -38,29 +42,23 @@
...
-        while (current == null || current.length <= pos) {
-            if (!base.hasNext()) {// no more to retrieve
-                current = null;
-                pos = 0;
-                return;
-            }
-
-            current = base.next();
-            wrapUp(current);
+        if ((current == null || current.length <= pos) && base.hasNext()) {
```
kshultzCB: Not really specific to this PR, but, I could use a quick overview of the relationships between `current`, `pos`, and `base`. I thought I had it figured out for a minute but I guess I don't.

bitwiseman:
From my discussion of this with @kshultzCB - This is very much a refactor with no significant performance changes.

src/main/java/org/kohsuke/github/PagedIterator.java
bitwiseman: I can pick better names for these and give them a little more commenting.

src/main/java/org/kohsuke/github/GitHubRequest.java
bitwiseman: I'm not sure.  These were here when I arrived.  I'm not sure it they are still all needed, per se.  I think I tried to remove them at one point and started getting errors ... but I don't remember now.

src/main/java/org/kohsuke/github/GitHubRequest.java
bitwiseman: Good question.  There are a small number of cases where the API returns a URL that is then used for other things.  

https://github.com/github-api/github-api/pull/697/files#diff-64dfb4dfbe21520e8735454a113c45a7R548-R560


https://github.com/github-api/github-api/blob/ff3136df7069d22d937daa8f2cddc357c777e07e/src/main/java/org/kohsuke/github/GHContent.java#L225-L227

https://github.com/github-api/github-api/blob/ff3136df7069d22d937daa8f2cddc357c777e07e/src/main/java/org/kohsuke/github/GHContent.java#L236-L240

Also, for paginated result, the full url for the next page is provided in the `Link` header , so when we call `getApiURL` we want to use that url, whatever it happens to be.

https://github.com/github-api/github-api/blob/dc33e28452532b94c6d46e435c192ae043849ca9/src/main/java/org/kohsuke/github/GitHubPageResponseIterator.java#L68-L77

I don't love this lack of clarity, but I didn't feel confident that I had tests for all cases where that URL didn't start with `/`, so I've kept the behavior where this method just handles it.  That said, I should change L77 to use `setRawUrl`.  Thanks!

src/main/java/org/kohsuke/github/GitHubRequest.java
bitwiseman: The other places are mostly done automatically by Jackson.   

https://github.com/github-api/github-api/pull/697/files#diff-04099021093ec34af9d55a855cc1e6fdR85-R90

Hm, the `toLowerCase` may not be needed due to line 88.  And line 89 means that, given a field like `create_date`, Jackson will automatically look for a field with the name `createdDate` and then fall back to `create_date` if it doesn't find one. 

However, this setting was only added recently, there are still a lot private fields with underscores in them.

src/main/java/org/kohsuke/github/GitHubRequest.java
bitwiseman: Opened #699 to come back to this.

src/main/java/org/kohsuke/github/GitHubRequest.java
bitwiseman: Opened #700 to clean this up.

bitwiseman:
@kshultzCB  Thanks!  Great feedback.  Made updates to address and filed a couple of issue to track future work.

Issue #714, opened 2020-02-24
GHOrganization::listMembersWithFilter does not create correct URL


Sage-Pierce:
**Describe the bug**
When trying to list members with a certain filter (or role), a malformed URL is created and GitHub returns `Not Found`

**To Reproduce**
1. Try to list members of an org with any filter by calling `GHOrganization.listMembersWithFilter("all")`

**Expected behavior**
Proper URL should be created and a list of members returned

**Desktop (please complete the following information):**
 - OS: MacOS High Sierra

**Additional context**
Would also be a good opportunity to allow the `role` query parameter

PR #715, opened 2020-02-24
[#714]Fix query parameter construction of org member filter

Sage-Pierce:
# Description 
Taking a stab at what's mentioned in #714 


src/test/java/org/kohsuke/github/GHOrganizationTest.java
```diff
@@ -76,6 +77,15 @@ public void testInviteUser() throws IOException {
         // assertTrue(org.hasMember(user));
     }
 
+    @Test
+    public void testListMembersWithRole() throws IOException {
+        GHOrganization org = gitHub.getOrganization(GITHUB_API_TEST_ORG);
+
+        List<GHUser> admins = org.listMembersWithRole("admin").asList();
```
bitwiseman: The bug was for `listMembersWithFilter`.  Can you add a test for that as well?

src/test/java/org/kohsuke/github/GHOrganizationTest.java
```diff
@@ -76,6 +77,15 @@ public void testInviteUser() throws IOException {
...
+
+        assertFalse(admins.isEmpty());
```
bitwiseman: Go ahead and be specific about size and who is in the list.  A non-empty list is going to be enough to ensure this doesn't break.

src/test/java/org/kohsuke/github/GHOrganizationTest.java
Sage-Pierce: Done!

PR #716, opened 2020-02-25
Use InjectableValue to provide response info

bitwiseman:
# Description 
Jackson can do binding of injected values for us rather than needing to special case them.  

A continuation of this change would be to remove most of the initialization `wrapUp(root)` methods and add a binding for `GitHub` member. 

# Before submitting a PR:

### DIFF
--- a/src/main/java/org/kohsuke/github/GHObject.java
+++ b/src/main/java/org/kohsuke/github/GHObject.java
@@ -1,5 +1,6 @@
 package org.kohsuke.github;
 
+import com.fasterxml.jackson.annotation.JacksonInject;
 import com.infradna.tool.bridge_method_injector.WithBridgeMethods;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.apache.commons.lang3.builder.ReflectionToStringBuilder;
@@ -31,6 +32,19 @@ public abstract class GHObject {
     protected String updated_at;
 
     GHObject() {
+    }
+
+    /**
+     * Called by Jackson
+     * 
+     * @param responseInfo
+     *            the {@link GitHubResponse.ResponseInfo} to get headers from.
+     */
+    @JacksonInject
+    protected void setResponseHeaderFields(@CheckForNull GitHubResponse.ResponseInfo responseInfo) {
+        if (responseInfo != null) {
+            responseHeaderFields = responseInfo.headers();
+        }
     }
 
     /**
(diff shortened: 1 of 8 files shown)


src/main/java/org/kohsuke/github/GitHubClient.java
```diff
@@ -443,30 +445,6 @@ private static IOException interpretApiError(IOException e,
...
-     * currently they do not have access to {@link GitHubResponse.ResponseInfo} after the
-     *
-     * @param responseInfo
-     *            the response info
-     * @param readValue
-     *            the object to consider adding headers to.
-     * @param <T>
-     *            type of the object
-     */
-    private static <T> void setResponseHeaders(GitHubResponse.ResponseInfo responseInfo, T readValue) {
```
bitwiseman: This is the special case code that we had before.

src/main/java/org/kohsuke/github/GHObject.java
```diff
@@ -33,6 +34,19 @@
...
     }
 
+    /**
+     * Called by Jackson
+     * 
+     * @param responseInfo
+     *            the {@link GitHubResponse.ResponseInfo} to get headers from.
+     */
+    @JacksonInject
+    protected void setResponseHeaderFields(@CheckForNull GitHubResponse.ResponseInfo responseInfo) {
```
bitwiseman: This is what we do instead of the special case.

src/main/java/org/kohsuke/github/GHRateLimit.java
```diff
@@ -308,31 +306,37 @@ private UnknownLimitRecord() {
...
+         * @param responseInfo
+         *            the response info
          */
-        @SuppressFBWarnings(value = "URF_UNREAD_PUBLIC_OR_PROTECTED_FIELD", justification = "Deprecated")
-        public Record(int limit, int remaining, long resetEpochSeconds, @CheckForNull String updatedAt) {
+        @JsonCreator
+        Record(@JsonProperty(value = "limit", required = true) int limit,
+                @JsonProperty(value = "remaining", required = true) int remaining,
+                @JsonProperty(value = "reset", required = true) long resetEpochSeconds,
+                @JacksonInject @CheckForNull GitHubResponse.ResponseInfo responseInfo) {
```
bitwiseman: Also, since we now have the the `responseInfo` injected during binding, we can make `Record` fully immutable.  We don't have "recalculate" after creation.

Issue #599, opened 2019-11-08
Replace `wrapUp` and `wrap()` methods with `@JacksonInject`


bitwiseman:
This is very much a code smells issue, not a user facing feature. 

There's quite a bit of code and complexity devoted to added `root` references to objects after they are received.  I've thought about various class structures to clean the up, but really the right thing to do seems to be using the `@JacksonInject` annotation.  We'd add the `root` instance to the mapper when reading and let the injector handle the assignment. 

See https://fasterxml.github.io/jackson-databind/javadoc/2.9/com/fasterxml/jackson/databind/ObjectMapper.html#reader-com.fasterxml.jackson.databind.InjectableValues-

bitwiseman:
After some attempt at this I found that some parts of the project depend on the late binding behavior or provided by the current `wrap*` methods. 
 
The `root` field is generally safe to bind early (it alway is preset before the query starts) but aside from that many `owner`s have scenarios in which they are not available until late or are even optional.  

It might be possible to create a `Binding` class that would hold these reference an allow them to be resolved later, but I'm not sure how useful that is.  

Also, the `@JacksonInject` entries must preset in the `InjectableValues` or else the reader throws, making optional/late binding even more of a pain.  This may not be worth the effort.

bitwiseman:
If nothing else changing the name of `wrapUp()` and `wrap()` to `lateBind()`

PR #720, opened 2020-02-27
Enable injection of GitHub and other values into mapping

bitwiseman:
# Description 
Progress on #599.  While there will alway be a need for late binding in some scenarios, binding of `GitHub` reference can always be done during mapping.   Other actions that were being done after mapping can also sometimes be done during mapping by using a constructor.  Values not initialized in the constructor are bound via reflection as they always were.

# Before submitting a PR:

### DIFF
--- a/src/main/java/org/kohsuke/github/GHGist.java
+++ b/src/main/java/org/kohsuke/github/GHGist.java
@@ -1,12 +1,13 @@
 package org.kohsuke.github;
 
+import com.fasterxml.jackson.annotation.JacksonInject;
+import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import org.apache.commons.lang3.StringUtils;
 
 import java.io.IOException;
 import java.net.URL;
 import java.util.Collections;
-import java.util.HashMap;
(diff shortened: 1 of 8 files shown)


src/main/java/org/kohsuke/github/GHGist.java
```diff
@@ -43,7 +57,7 @@
      *             the io exception
      */
     public GHUser getOwner() throws IOException {
-        return root.intern(owner);
+        return owner;
```
bitwiseman: owner is alway interned during construction now.

src/main/java/org/kohsuke/github/GHGist.java
```diff
@@ -139,31 +153,7 @@ public GHGistFile getFile(String name) {
...
-        this.owner = root.getUser(owner);
-        this.root = root;
-        wrapUp();
-        return this;
-    }
-
-    private void wrapUp() {
-        for (Entry<String, GHGistFile> e : files.entrySet()) {
-            e.getValue().fileName = e.getKey();
-        }
```
bitwiseman: Actions that were being done after instantiation are now done in the constructor, allowing for those values to be `final`.

PR #723, opened 2020-02-29
Complete checks api

XiongKezhi:
# Description 
Adding more properties to GHCheckRun (in progress) and implementing GHCheckSuite (in progress).

# Before submitting a PR:

- [X] Run `mvn -D enable-ci clean install site` locally. This may reformat your code, commit those changes. If this command doesn't succeed, your change will not pass CI.


src/main/java/org/kohsuke/github/GHCheckRun.java
```diff
@@ -62,6 +88,12 @@ public String getHeadSha() {
         return headSha;
     }
 
+    /**
+     * Gets the pull requests participated in this check run.
+     *
+     * @return Pull requests of this check run
+     */
+
```
recena: I assume that empty line is not needed.

XiongKezhi: Deleted. Thanks!

bitwiseman:
@XiongKezhi 
I'm about to start reviewing this but want to start by saying thank you.  This is looks like a big improvement.

src/main/java/org/kohsuke/github/GHCheckRun.java
```diff
@@ -72,11 +103,106 @@ public String getHeadSha() {
...
+     * Represents an output in a check run to include summary and other results.
+     *
+     * @see <a href="https://developer.github.com/v3/checks/runs/#output-object">documentation</a>
+     */
+    public static class Output {
+        private String title;
+        private String summary;
+        private String text;
+        private int annotationsCount;
+        private URL annotationsUrl;
```
bitwiseman: Do we have a `CheckRunAnnotations` yet? If not we'll need that added eventually as well.

bitwiseman:
By the way, you have this marked as draft but this seems like a perfectly reasonable PR to submit and complete on its own if you choose to do so.

src/main/java/org/kohsuke/github/GHCheckRun.java
```diff
@@ -21,8 +21,18 @@
     private String conclusion;
     private String name;
     private String headSha;
+    private String nodeId;
+    private String externalId;
+    private String startedAt;
+    private String completedAt;
+    private URL htmlUrl;
```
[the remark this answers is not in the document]
XiongKezhi: We can inherit the `url` and `getUrl`, but the `getHtmlUrl()` is abstract in `GHObject`, so I have to overwrite it.

src/main/java/org/kohsuke/github/GHCheckRun.java
```diff
@@ -21,8 +21,18 @@
     private String conclusion;
     private String name;
     private String headSha;
+    private String nodeId;
+    private String externalId;
+    private String startedAt;
+    private String completedAt;
```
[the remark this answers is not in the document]
XiongKezhi: I just saw the `GHObject` uses `String` for such timestamps at first and convert them into `Date` in getters. But yes, using `Date` from the beginning is better. I've changed it.

src/main/java/org/kohsuke/github/GHCheckRun.java
XiongKezhi: That's where I'm confused. The response (or event payload) from GitHub does not include `annotation` objects, but they are needed when users are making requests. So does our implementation have to make sure it supports both usages?

XiongKezhi:
> By the way, you have this marked as draft but this seems like a perfectly reasonable PR to submit and complete on its own if you choose to do so.

It's because I'm still implementing the `GHCheckSuite`, I'll make it a normal PR after that. Appreciate your review!

XiongKezhi:
> Just so I understand it: the reason why there aren't more tests in here is because, for general reading of check runs (outside of events) we need to have a `GHCheckSuite` for them right?

I'm not sure which kinds of tests our project requires, and I normally do not test the getters and setters, but I assume we have to test them here because the values are bound by Jackson? Also, I'm not sure how complete our tests have to be, for example, there is a `GHRepository`in our `GHEventPayload.CheckRun`, do we have to test whether each value of `GHRepository`is correct? I think such a recursive style is not necessary, so I just test some basic properties like `id` or `name` of the objects in `CheckRun` to make sure that Jackson reaches those.

src/main/java/org/kohsuke/github/GHCheckRun.java
sladyn98: Most probably it should support. Would be beneficial for users in the longer term

src/test/java/org/kohsuke/github/GHEventPayloadTest.java
```diff
@@ -320,8 +323,32 @@ public void checkRunEvent() throws Exception {
...
         assertThat(event.getRepository().getName(), is("Hello-World"));
         assertThat(event.getAction(), is("created"));
-        assertThat(event.getCheckRun().getName(), is("Octocoders-linter"));
-        assertThat(event.getCheckRun().getHeadSha(), is("ec26c3e57ca3a959ca5aad62de7213c562f8c821"));
+
+        // Checks the deserialization of check_run
+        GHCheckRun checkRun = event.getCheckRun();
+        assertThat(checkRun.getName(), is("Octocoders-linter"));
+        assertThat(checkRun.getHeadSha(), is("ec26c3e57ca3a959ca5aad62de7213c562f8c821"));
+        assertThat(checkRun.getStatus(), is("completed"));
```
sladyn98: @XiongKezhi I guess there are multiple states of the check run that need to be tested for, Is this the only one?

XiongKezhi: Yeah, that's one thing we can do. Before, I just changed the original JSON in order to cover more methods for convenience, but later I found more sample responses [here](https://developer.github.com/v3/checks/runs/), they are much better and maybe we can use them later.

src/main/java/org/kohsuke/github/GHCheckRun.java
bitwiseman: It isn't required for this PR, I'm just noting it for later.

src/main/java/org/kohsuke/github/GHCheckRun.java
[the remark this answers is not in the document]
bitwiseman: > > Just so I understand it: the reason why there aren't more tests in here is because, for general reading of check runs (outside of events) we need to have a `GHCheckSuite` for them right?
> 
> I'm not sure which kinds of tests our project requires, and I normally do not test the getters and setters, but I assume we have to test them here because the values are bound by Jackson? Also, I'm not sure how complete our tests have to be, for example, there is a `GHRepository`in our `GHEventPayload.CheckRun`, do we have to test whether each value of `GHRepository`is correct? I think such a recursive style is not necessary, so I just test some basic properties like `id` or `name` of the objects in `CheckRun` to make sure that Jackson reaches those.

I'm going to give you vague-ish answer.  We require test coverage for any new code/features added to the project, but that is a recent decision, so there are lots of gaps  in test coverage for the project overall.  You do not need to do recursive tests, but we do need tests that show the objects are behaving correctly overall.  You can save yourself some testing by not having public setters - then you only need to touch each of the getters at least once. Write a set of tests that your feel are reasonable and we can iterate to add more if they seem needed.

bitwiseman:
> > Just so I understand it: the reason why there aren't more tests in here is because, for general reading of check runs (outside of events) we need to have a `GHCheckSuite` for them right?
> 
> I'm not sure which kinds of tests our project requires, and I normally do not test the getters and setters, but I assume we have to test them here because the values are bound by Jackson? Also, I'm not sure how complete our tests have to be, for example, there is a `GHRepository`in our `GHEventPayload.CheckRun`, do we have to test whether each value of `GHRepository`is correct? I think such a recursive style is not necessary, so I just test some basic properties like `id` or `name` of the objects in `CheckRun` to make sure that Jackson reaches those.

I'm going to give you vague-ish answer.  We require test coverage for any new code/features added to the project, but that is a recent decision, so there are lots of gaps  in test coverage for the project overall.  You do not need to do recursive tests, but we do need tests that show the objects are behaving correctly overall.  You can save yourself some testing by not having public setters - then you only need to touch each of the getters at least once. Write a set of tests that your feel are reasonable and we can iterate to add more if they seem needed.

src/main/java/org/kohsuke/github/GHEventPayload.java
```diff
@@ -150,6 +150,47 @@ void wrapUp(GitHub root) {
...
+            return checkSuite;
+        }
+
+        /**
+         * Gets repository.
+         *
+         * @return the repository
+         */
+        public GHRepository getRepository() {
+            return repository;
```
bitwiseman:
```suggestion
            repository.root = root;
            return repository;
```

src/main/java/org/kohsuke/github/GHCommit.java
```diff
@@ -38,12 +34,43 @@
...
+        public String getId() {
+            return id;
+        }
+
+        /**
+         * Gets timestamp of the commit, used by {@link GHCheckSuite} when a {@link GHEvent#CHECK_SUITE} comes
+         *
+         * @return timestamp of the commit
+         */
+        public Date getTimestamp() {
```
bitwiseman: I don't see it on this page: https://developer.github.com/v3/repos/commits/#get-a-single-commit 

I see this:
https://developer.github.com/v3/activity/events/types/#webhook-payload-example-1

It possible the "issue" in `GHCheckSuite` is not a `GHIssue`?  But instead something else?

XiongKezhi: sry, didn't notice that [commit page](https://developer.github.com/v3/repos/commits/#get-a-single-commit) and pay much attention to the structure. I'll add it inside `GHCheckSuite`.

oleg-nenashev:
Nice one!

PR #724, opened 2020-03-02
Add Builder/Creator/Updater for GHLabel

bitwiseman:
Here's another possible way to address #589 .  This PR uses `GHLabel` as an example.

@rmetzger @halkeye @PauloMigAlmeida @recena @res0nance - There's still a bunch to be done here (`site` currently fails due to missing JavaDocs for example), but if you look at the changes to the tests you can see what this new syntax will look like. My intent is to create a pattern that can be used consistently across the project. What do you think of this design? 

Highlights: 
* Support for batching changes.  Batching looks like this: 
```java
label.update().someName(value).otherName(value).done()
```
* Support for single changes.  They look like this:
```java
label.set().someName(value); 
label.set().otherName(value)
```
* Support for dynamically controlled "update-in-place" vs "update-to-new-instance" behavior (not active or tested yet). As in #649, there are times people want/expect update-in-place and other times they don't. This design will allow for both, though for now it only does "update-to-new-instance".
* Significant code reuse - the real win of this POC is that these differing scenarios do not each require wildly differing code paths.  Batch/single updating is provided by almost identical code paths.  In the case of `GHLabel`, new instance creation does too (though in other classes creation may diverge more).  "update-in-place" vs "update-to-new-instance" behavior actually identical except for one `if-else` in the the base class.  This means that we can get this added functionality without significant added complexity. 

Technical notes:
* Tried to add `final` to guarantee immutability but decided against it - While guaranteeing immutability via `final` fields sounds great, it results in a lot of extra code to make it work with Jackson in a way that the compiler doesn't choke on.


src/test/java/org/kohsuke/github/AppTest.java
```diff
@@ -784,44 +784,47 @@ public void testRepoLabel() throws IOException {
...
+
+            // instances are immutable, but update returns a new updated instance.
+            assertEquals(t.color(), "123456");
+            assertEquals(t.description(), "");
+            assertEquals(t2.color(), "000000");
+            assertEquals(t2.description(), "It is dark!");
 
             t = r.getLabel("test");
-            t.setDescription("this is also a test");
+            GHLabel t3 = t.set().description("this is also a test");
```
bitwiseman: Single Update

src/test/java/org/kohsuke/github/AppTest.java
```diff
@@ -784,44 +784,47 @@ public void testRepoLabel() throws IOException {
...
+            assertEquals(t.name(), t2.name());
+            assertEquals(t.color(), "123456");
+            assertEquals(t.color(), t2.color());
+            assertEquals(t.description(), "");
+            assertEquals(t.description(), t2.description());
+            assertEquals(t.url(), t2.url());
+
+            // update works on multiple changes in one call
+            // t.setColor("");
+            t2 = t.update().color("000000").description("It is dark!").done();
```
bitwiseman: Batch update

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -1842,9 +1842,7 @@ public GHCommitStatus createCommitStatus(String sha1, GHCommitState state, Strin
      *             the io exception
      */
     public PagedIterable<GHLabel> listLabels() throws IOException {
-        return root.createRequest()
-                .withUrlPath(getApiTailUrl("labels"))
-                .toIterable(GHLabel[].class, item -> item.wrapUp(this));
+        return GHLabel.readAll(this);
```
bitwiseman: Moved all the calls that create instances of `GHLabel` into that class.

src/test/java/org/kohsuke/github/AppTest.java
```diff
@@ -800,16 +800,19 @@ public void testRepoLabel() throws IOException {
...
             assertEquals(t.getUrl(), t2.getUrl());
 
-            t.setColor("000000");
+            // update works on multiple changes in one call
+            // t.setColor("");
+            t2 = t.update().color("000000").description("It is dark!").done();
 
-            // This is annoying behavior, but it is by design at this time.
-            // Verifying so we can know when it is fixed.
+            // instances are immutable, but update returns a new updated instance.
```
bitwiseman:
```suggestion
            // instances behave as immutable by default. Update returns a new updated instance.
```

src/main/java/org/kohsuke/github/AbstractBuilder.java
```diff
@@ -0,0 +1,106 @@
...
+     *            the value of the field
+     * @return either a continuing builder or an updated data record
+     * @throws IOException
+     *             if an I/O error occurs
+     */
+    @Nonnull
+    protected S with(@Nonnull String name, Object value) throws IOException {
+        requester.with(name, value);
+        if (returnType.equals(intermediateReturnType)) {
+            return intermediateReturnType.cast(done());
```
res0nance: This threw me off the first time I read this code to realize that this is what ensures setters still make the REST API call instead of batching

res0nance:
Overall I like the idea because of the ability to batch changes together. The api looks very intuitive. The base class is a little confusing for people intending to create a new API but it looks as simple as it can be while providing the benefit of reusability across the project.

src/main/java/org/kohsuke/github/AbstractBuilder.java
bitwiseman: I should make it a method or boolean field with comments to make it clearer.

bitwiseman: @res0nance How about now? Better?

src/main/java/org/kohsuke/github/GHLabel.java
```diff
@@ -48,41 +69,23 @@ public String getColor() {
      *
      * @return the description
      */
+    @Nonnull
     public String getDescription() {
         return description;
     }
 
-    GHLabel wrapUp(GHRepository repo) {
```
bitwiseman: Replaced `wrapUp()` with `lateBind()`.

src/main/java/org/kohsuke/github/AbstractBuilder.java
```diff
@@ -0,0 +1,147 @@
...
+     * Finishes an update, committing changes.
+     *
+     * This method may update-in-place or not. Either way it returns the resulting instance.
+     *
+     * @return an instance with updated current data
+     * @throws IOException
+     *             if there is an I/O Exception
+     */
+    @Nonnull
+    public R done() throws IOException {
```
bitwiseman: Any suggestions for a better method name than `done()`?   `send()`?  `submit()`?  
We can't use `commit()` because it will almost surely collide with a git "commit" at least conceptually if not literally.

res0nance: BulkChange in core uses `commit` and `abort`. abort is not needed. Maybe `send` is a good choice. or `call`?

bitwiseman: The two places this will get used look something like this:
```java
GHLabel label = GHLabel.create().name(name).color(color).description(description).xxxx();
label.update().color(color2).description(description2).xxxx();
```
`send()` could work.  `call` not as much. I'd just like to make sure we've considered alternatives before releasing this.

bitwiseman: Okay, well, this api is internal, so changing it later is not a breaking API change.  I'll go with the current verbs for now and we can change them later if needed.

src/main/java/org/kohsuke/github/AbstractBuilder.java
res0nance: Much better

src/main/java/org/kohsuke/github/AbstractBuilder.java
bitwiseman: Hm, doh, no, what was I thinking this is public.  Ah, but I can use the same structure we already have for API Previews and ship this now.  😄There we go.

PR #738, opened 2020-03-12
Get commit or tag signature verified flag

### LINKED ISSUE #737: Get commit or tag signature verified flag
It is observed that some commits/tags are verified by user signature. Depending on this signature authorization. A `verified` flag is set to `true/false`.

Request to add this as a feature request. Such that when we collect individual commit or tag information. We get an associated signature `verified` flag as well.

For more details please check [here](https://help.github.com/en/github/authenticating-to-github/managing-commit-signature-verification).

sourabhsparkala:
This fixes #737

# Description 

- A new entity GHVerification.java has been added which would be reflecting Verification flag.
- Updating GHCommit.java and GHTagObject.java with GHVerification.
- Altering few test cases AppTest.java and GHTagTest.java to verify if the Verification entity is being picked up.
- A separate test class SignatureVerificationTest.java with the associated wiremock test resources


src/main/java/org/kohsuke/github/GHVerification.java
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * The commit/tag would be signed by user. This object would hold the verification status. Whether the Commit/Tag is
+ * signed or not.
+ *
+ * @author Sourabh Sarvotham Parkala
+ */
+@SuppressFBWarnings(value = { "UWF_UNWRITTEN_PUBLIC_OR_PROTECTED_FIELD", "UWF_UNWRITTEN_FIELD", "NP_UNWRITTEN_FIELD" },
+        justification = "JSON API")
+public class GHVerification {
+    private String reason, signature, payload;
```
artem-smotrakov: Would it make sense to create a enum which holds all possible reasons? The following page defines possible reasons (please see the "Signature verification object" section):

https://developer.github.com/v3/repos/commits/

src/main/java/org/kohsuke/github/GHVerification.java
```diff
@@ -0,0 +1,52 @@
...
+    private boolean verified;
+
+    /**
+     * Gets boolean value.
```
artem-smotrakov: Minor: How about "Indicates whether GitHub considers the signature in this commit to be verified"? (borrowed from https://developer.github.com/v3/repos/commits/)

src/main/java/org/kohsuke/github/GHVerification.java
```diff
@@ -0,0 +1,52 @@
...
+     * Gets signature used for the verification.
+     *
+     * @return null if not signed else encoded signature.
+     */
+    public String getSignature() {
+        return signature;
+    }
+
+    /**
+     * Gets payload used for the verification.
```
artem-smotrakov: I would suggest "Gets the payload that was signed."

sourabhsparkala:
Thank you for the review @artem-smotrakov. Please let me know if there are any more changes.

src/main/java/org/kohsuke/github/GHReason.java
```diff
@@ -0,0 +1,34 @@
...
+    UNKNOW_SIGNATURE_TYPE("unknown_signature_type"),
+    NO_USER("no_user"),
+    UNVERIFIED_EMAIL("unverified_email"),
+    BAD_EMAIL("bad_email"),
+    UNKNOWN_KEY("unknown_key"),
+    MALFORMED_SIGNATURE("malformed_signature"),
+    INVALID("invalid"),
+    VALID("valid");
+
+    private final String value;
```
artem-smotrakov: Not sure if `value` is necessary.

src/main/java/org/kohsuke/github/GHReason.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * The possible values for reason in verification object from github.
+ * 
+ * @see <a href="https://developer.github.com/v3/repos/commits/#signature-verification-object">List of possible reason
+ *      values</a>
+ * @author Sourabh Sarvotham Parkala
+ */
+public enum GHReason {
+    EXPIRED_KEY("expired_key"),
```
bitwiseman: We need at test for each of these values. Or at least a bunch of them.  Even if you have to manually create them.

src/main/java/org/kohsuke/github/GHReason.java
bitwiseman: @artem-smotrakov 
In general, you'd be correct.  But I think the default mapping for enums is all-caps and convert-underscores-to-dashes. Which would result in `EXPIRED_KEY` converting to `expired-key` and vice versa.  Once tests are added, we can see whether this is needed or not.

src/test/java/org/kohsuke/github/AppTest.java
```diff
@@ -584,6 +584,8 @@ public void testCommitShortInfo() throws Exception {
         GHCommit commit = r.getCommit("86a2e245aa6d71d54923655066049d9e21a15f23");
         assertEquals(commit.getCommitShortInfo().getAuthor().getName(), "Kohsuke Kawaguchi");
         assertEquals(commit.getCommitShortInfo().getMessage(), "doc");
+        assertFalse(commit.getCommitShortInfo().getVerification().getVerified());
+        assertEquals(commit.getCommitShortInfo().getVerification().getReason(), GHReason.UNSIGNED.get());
```
bitwiseman: As noted above, we need tests for at least a few more `GHReason` values (specifically multi-word values).

bitwiseman:
@sourabhsparkala You see we need some tests right?

sourabhsparkala:
> @sourabhsparkala You see we need some tests right?

@bitwiseman I am working on the tests now.

sourabhsparkala:
@bitwiseman I have added the test cases as advised.

sourabhsparkala:
@bitwiseman Thank you for reviewing and merging the PR.

Issue #520, opened 2019-05-22
Add support for the Checks API


serkankirbas:
The Github v3 API now supports a first-class notion of [check runs](https://developer.github.com/v3/checks/runs/) and [check suites](https://developer.github.com/v3/checks/suites/). Rather than binary pass/fail build statuses, integrations can report rich statuses, annotate lines of code with detailed information, and kick off reruns. 

Github checks are becoming the de facto approach for implementing bots and for users who want to develop custom checks. I think supporting Checks API should be a high priority.

bitwiseman:
Agreed. Those would be great!  Do you have time to submit a PR?

XiongKezhi:
Does the work to represent the Check Suite event payload still need any help?  I am trying to implement the Checks API as a Jenkins plugin for a GSoC idea (see [GitHub Checks API for Jenkins Plugins ](https://jenkins.io/projects/gsoc/2020/project-ideas/github-checks/)), and I'm happy to help with this, so it can make my later implementations easier!

bitwiseman:
It looks like we have read access to checks runs and checkrun events, but not suites and write access. 

https://github.com/github-api/github-api/search?q=GHCheckRun&unscoped_q=GHCheckRun

You'll want to look at the links in the description and implement those. I'd suggest starting off with some smaller PRs or open a draft PR so I can give you pointers as you go along.

XiongKezhi:
Thanks, the links are really helpful, I was strange about those class without setters before. I'll start off with a draft PR on CheckSuite and CheckSuiteEvent.

bitwiseman:
Fixed by #723 .

XiongKezhi:
Sorry, I just saw this message. Also, until I saw those builders and  [#724](https://github.com/github-api/github-api/pull/724), I thought our project left other creating works to the users themselves because we have already provided a pluggable Http client. I wonder whether our current ways to implement those CRUD methods are to follow the design of [#724](https://github.com/github-api/github-api/pull/724)? Like for creating a `GHCommit`, we use a `GHCommitBuilder`, so we also need a `GHCheckRunBuilder` or `GHCheckSuiteBuilder` to create them?

bitwiseman:
@XiongKezhi 
#724 is still just POC. I've heard positive feedback so far, but it isn't quite ready for wide use. But, yes, you probably want to create builder and updater classes.

bitwiseman:
[...]

PR #753, opened 2020-03-25
GHRepository.createCheckRun

jglick:
# Description 

Continues #723 & #740 to [create a check run](https://developer.github.com/v3/checks/runs/#create-a-check-run). Partially addresses #520.

I am leery of building this on top of #724, which I can barely understand but anyway does not feel appropriate here since for example a `GHCheckRun` is not a complete model object: it has an `Output` nested object with an `annotationsCount` & `annotationsUrl` but there is not currently any separate model for the actual annotation data. _Some_ sort of builder pattern is needed, but I find the style used here (inspired by `GHPullRequestReviewBuilder`) more intuitive and straightforward and it is clear what Java method actually corresponds to the REST call.

# Before submitting a PR:


XiongKezhi:
Do we also need another `updateCheckRun` method for [this](https://developer.github.com/v3/checks/runs/#update-a-check-run) here? They are almost the same except for the URL and request method.

jglick:
> Do we also need another `updateCheckRun` method

Yes, could be added as well. AFAICT it is not a top priority since “creating” a run with the same name as an existing one effectively overwrites it.

src/main/java/org/kohsuke/github/GHCheckRun.java
```diff
@@ -51,19 +51,20 @@ GHCheckRun wrap(GitHub root) {
     }
 
     /**
-     * Gets status of the check run. It can be one of "queue", "in_progress", or "completed"
```
jglick: `queued` actually

src/main/java/org/kohsuke/github/GHCheckRun.java
```diff
@@ -51,19 +51,20 @@ GHCheckRun wrap(GitHub root) {
...
      * @return Status of the check run
+     * @see GHCheckRunStatus
      */
     public String getStatus() {
         return status;
     }
 
     /**
-     * Gets conclusion of a completed check run. It can be one of "success", "failure", "neutral", "cancelled",
-     * "time_out", or "action_required".
```
jglick: `timed_out` actually

src/main/java/org/kohsuke/github/GHCheckRunAnnotationLevel.java
```diff
@@ -0,0 +1,29 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.kohsuke.github;
+
+public enum GHCheckRunAnnotationLevel {
```
bitwiseman: Out of curiosity, why not make this an inner class of `GHCheckRun`?  (`GHCheckRun.AnnotationLevel`)

src/test/java/org/kohsuke/github/GHCheckRunBuilderTest.java
```diff
@@ -0,0 +1,55 @@
...
+        GHCheckRun checkRun = gitHub.getRepository("jglick/github-api-test")
+                .createCheckRun("foo", "4a929d464a2fae7ee899ce603250f7dab304bc4b")
+                .withStatus(GHCheckRunStatus.COMPLETED)
+                .withConclusion(GHCheckRunConclusion.SUCCESS)
+                .withDetailsURL("http://nowhere.net/stuff")
+                .withExternalID("whatever")
+                .withStartedAt(new Date(999_999_000))
+                .withCompletedAt(new Date(999_999_999))
+                .withOutput("Some Title", "what happened…")
+                .withAnnotation("stuff.txt", 1, GHCheckRunAnnotationLevel.NOTICE, "hello to you too")
```
jglick: The formatter plugin obliterates attempts to indent this readably.

src/main/java/org/kohsuke/github/GHCheckRunAnnotationLevel.java
jglick: I could. Did not see a clear convention here.

bitwiseman: I was more interested in what your reasoning was. In this case I would lean towards the inner class since it is the `AnnotationLevel` for  `GHCheckRun` only, not for any other class, but that feels like a somewhat weak reason.

src/main/java/org/kohsuke/github/GHCheckRunBuilder.java
```diff
@@ -0,0 +1,358 @@
...
+     * @param summary
+     *            as in GitHub documentation
+     * @return use {@link DraftOutput#done} to continue
+     */
+    public @NonNull DraftOutput withOutput(@NonNull String title, @NonNull String summary) {
+        return new DraftOutput(this, title, summary);
+    }
+
+    /**
+     * Drafts an action section.
```
bitwiseman: Add links to the pertinent GitHub docs, such as https://developer.github.com/v3/checks/runs/#actions-object.

src/main/java/org/kohsuke/github/GHCheckRunBuilder.java
```diff
@@ -0,0 +1,358 @@
...
+    /**
+     * Drafts the output section.
+     *
+     * @param title
+     *            as in GitHub documentation
+     * @param summary
+     *            as in GitHub documentation
+     * @return use {@link DraftOutput#done} to continue
+     */
+    public @NonNull DraftOutput withOutput(@NonNull String title, @NonNull String summary) {
```
bitwiseman: Since this is an inner class could this be shortened to `Output`?

src/test/java/org/kohsuke/github/GHCheckRunBuilderTest.java
```diff
@@ -0,0 +1,88 @@
...
+        GHCheckRunBuilder.DraftOutput output = gitHub.getRepository("jglick/github-api-test")
+                .createCheckRun("big", "4a929d464a2fae7ee899ce603250f7dab304bc4b")
+                .withConclusion(GHCheckRun.Conclusion.SUCCESS)
+                .withOutput("Big Run", "Lots of stuff here »");
+        for (int i = 0; i < 101; i++) {
+            output.withAnnotation("stuff.txt", 1, GHCheckRun.AnnotationLevel.NOTICE, "hello #" + i).done();
+        }
+        GHCheckRun checkRun = output.done().create();
+        assertEquals("completed", checkRun.getStatus());
+        assertEquals("Big Run", checkRun.getOutput().getTitle());
```
bitwiseman: Check OutputSummary as well.

src/test/java/org/kohsuke/github/GHCheckRunBuilderTest.java
```diff
@@ -0,0 +1,88 @@
...
+        assertEquals(101, checkRun.getOutput().getAnnotationsCount());
+        assertEquals(537614832, checkRun.id);
+    }
+
+    @Test
+    public void createCheckRunNoAnnotations() throws Exception {
+        GHCheckRun checkRun = gitHub.getRepository("jglick/github-api-test")
+                .createCheckRun("quick", "4a929d464a2fae7ee899ce603250f7dab304bc4b")
+                .withConclusion(GHCheckRun.Conclusion.NEUTRAL)
```
bitwiseman: Conclusion is optional when `competedAt` is not set.  need test for that.

src/test/java/org/kohsuke/github/GHCheckRunBuilderTest.java
```diff
@@ -0,0 +1,88 @@
...
+                .withConclusion(GHCheckRun.Conclusion.SUCCESS)
+                .withDetailsURL("http://nowhere.net/stuff")
+                .withExternalID("whatever")
+                .withStartedAt(new Date(999_999_000))
+                .withCompletedAt(new Date(999_999_999))
+                .withOutput("Some Title", "what happened…")
+                .withAnnotation("stuff.txt", 1, GHCheckRun.AnnotationLevel.NOTICE, "hello to you too")
+                .withTitle("Look here")
+                .done()
+                .withImage("Unikitty", "https://i.pinimg.com/474x/9e/65/c0/9e65c0972294f1e10f648c9780a79fab.jpg")
```
bitwiseman: Image only has one optional field maybe it doesn't need a builder? 

Are images not readable via api? I didn't see them...

src/test/java/org/kohsuke/github/GHCheckRunBuilderTest.java
```diff
@@ -0,0 +1,88 @@
...
+        GHCheckRun checkRun = gitHub.getRepository("jglick/github-api-test")
+                .createCheckRun("foo", "4a929d464a2fae7ee899ce603250f7dab304bc4b")
+                .withStatus(GHCheckRun.Status.COMPLETED)
+                .withConclusion(GHCheckRun.Conclusion.SUCCESS)
+                .withDetailsURL("http://nowhere.net/stuff")
+                .withExternalID("whatever")
+                .withStartedAt(new Date(999_999_000))
+                .withCompletedAt(new Date(999_999_999))
+                .withOutput("Some Title", "what happened…")
+                .withAnnotation("stuff.txt", 1, GHCheckRun.AnnotationLevel.NOTICE, "hello to you too")
```
bitwiseman: maybe `.addAnnotation(...)` or even `addAnnotations(DraftAnnotation... annotations)`.  Then you wouldn't need to be calling `done()` after each one.

src/test/java/org/kohsuke/github/GHCheckRunBuilderTest.java
bitwiseman: Also same comment as for `withAnnotation()` method signature.

src/test/java/org/kohsuke/github/GHCheckRunBuilderTest.java
jglick: `.addAnnotation(…)` as an alias for `.done().withAnnotation(…)` is possible, but does not improve clarity to my mind. `addAnnotations` makes it harder to customize each one, and anyway the typical caller is going to be creating these one by one in a loop, not hard-coded.

src/test/java/org/kohsuke/github/GHCheckRunBuilderTest.java
jglick: > Image only has one optional field maybe it doesn't need a builder?

Preferred to be consistent.

> Are images not readable via api?

Not sure offhand. For purposes of this PR I am focussing on creating check runs, not loading them.

src/test/java/org/kohsuke/github/GHCheckRunBuilderTest.java
jglick: Not intending to verify every single field here, only the parts that involve subtle logic.

src/test/java/org/kohsuke/github/GHCheckRunBuilderTest.java
jglick: You mean when `status` is not `completed`? Again, these are just aspects of the GitHub semantics which are uninteresting to test here since they are not exercising novel and subtle code idioms.

src/test/java/org/kohsuke/github/GHCheckRunBuilderTest.java
bitwiseman: It looks like there is a special character in the OutputSummary.  I'd like to see that it works. I'd also be fine you made the summary simple text.

jglick:
I suppose unrelated flake:

```
java.lang.AssertionError: 
if header instance expires but queried instance is valid, ratelimit() uses it without asking server
Expected: not sameInstance(<GHRateLimit {core {remaining=4897, limit=5000, resetDate=Fri Mar 27 18:17:03 GMT 2020}search {remaining=999999, limit=1000000, resetDate=Fri Mar 27 19:17:00 GMT 2020}graphql {remaining=999999, limit=1000000, resetDate=Fri Mar 27 19:17:00 GMT 2020}integrationManifest {remaining=999999, limit=1000000, resetDate=Fri Mar 27 19:17:00 GMT 2020}}>)
     but: was <GHRateLimit {core {remaining=4897, limit=5000, resetDate=Fri Mar 27 18:17:03 GMT 2020}search {remaining=999999, limit=1000000, resetDate=Fri Mar 27 19:17:00 GMT 2020}graphql {remaining=999999, limit=1000000, resetDate=Fri Mar 27 19:17:00 GMT 2020}integrationManifest {remaining=999999, limit=1000000, resetDate=Fri Mar 27 19:17:00 GMT 2020}}>
	at org.kohsuke.github.AbstractGitHubWireMockTest.assertThat(AbstractGitHubWireMockTest.java:261)
	at org.kohsuke.github.GHRateLimitTest.executeExpirationTest(GHRateLimitTest.java:417)
	at org.kohsuke.github.GHRateLimitTest.testGitHubRateLimitExpirationServerFiveMinutesAhead(GHRateLimitTest.java:343)
```

src/test/java/org/kohsuke/github/GHCheckRunBuilderTest.java
bitwiseman: @jglick 
I'm looking at https://developer.github.com/v3/checks/runs/#parameters

> Required if you provide completed_at or a status of completed. 

May not be super interesting but we should verify this not providing conclusion works.  If you remove this line from this test, we get that. 

Frankly, we should also have a test verifying the error returned when conclusion is not set but `status` is `completed`.  The test framework makes adding tests pretty quick.

src/test/java/org/kohsuke/github/GHCheckRunBuilderTest.java
bitwiseman: Okay, so:
```
.add(new Annotation("stuff.txt", 1, GHCheckRun.AnnotationLevel.NOTICE, "hello to you too")
    .withTitle("Look here"))
.add(new Annotation("stuff.txt", 3, GHCheckRun.AnnotationLevel.NOTICE, "bye")
    .withTitle("Look there"))
.add(new Image("Unikitty", "https://i.pinimg.com/474x/9e/65/c0/9e65c0972294f1e10f648c9780a79fab.jpg")
.withCaption("Princess Unikitty"))
```

src/main/java/org/kohsuke/github/GHCheckRunBuilder.java
```diff
@@ -0,0 +1,370 @@
...
+     */
+    public @NonNull Output withOutput(@NonNull String title, @NonNull String summary) {
+        return new Output(this, title, summary);
+    }
+
+    /**
+     * Drafts an action section.
+     *
+     * @param label
+     *            as in GitHub documentation
```
bitwiseman: What happens when the max size of this is exceeded?  Not mentioning that limit here will cause more pain for consumers of this library.

src/main/java/org/kohsuke/github/GHCheckRunBuilder.java
bitwiseman: There's a maximum of three actions allowed.  But the GitHub docs do not make it clear what happens when actions are exceeded.  We should at least mention the limit here.

src/main/java/org/kohsuke/github/GHCheckRunBuilder.java
```diff
@@ -0,0 +1,370 @@
...
+            return output;
+        }
+
+    }
+
+    /**
+     * @see <a href="https://developer.github.com/v3/checks/runs/#actions-object">documentation</a>
+     */
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    public static final class Action {
```
bitwiseman: Does this need to be `public`?  It is never needed by users.

src/main/java/org/kohsuke/github/GHCheckRunBuilder.java
```diff
@@ -0,0 +1,370 @@
...
+         *            as in GitHub documentation
+         * @param message
+         *            as in GitHub documentation
+         * @return use {@link Annotation#done} to continue
+         */
+        public @NonNull Annotation withAnnotation(@NonNull String path,
+                int line,
+                @NonNull GHCheckRun.AnnotationLevel annotationLevel,
+                @NonNull String message) {
+            return withAnnotation(path, line, line, annotationLevel, message);
```
sladyn98: Max limit for annotations ?

src/main/java/org/kohsuke/github/GHCheckRunBuilder.java
jglick: I would rather not duplicate backend logic here. It is up to callers to read the GitHub documentation and supply appropriate parameters accordingly. The limits may change over time, remember, and since Checks is still in preview there could be _new_ restrictions later. The library should limit itself to collecting data and shipping it off.

src/main/java/org/kohsuke/github/GHCheckRunBuilder.java
jglick: Nope, >50 is automatically handled, see above.

src/main/java/org/kohsuke/github/GHCheckRunBuilder.java
jglick: I was leaving it `public` in case the Jackson databinding required it to be, but can check if tests still pass if it is made less accessible.

src/test/java/org/kohsuke/github/GHCheckRunBuilderTest.java
jglick: Perhaps. Not quite as friendly to code completion. Can look into it.

src/test/java/org/kohsuke/github/GHCheckRunBuilderTest.java
jglick: > we should verify this not providing conclusion works

Sure.

> If you remove this line from this test

Yeah, so long as the code coverage report does not then start complaining that there is insufficient coverage percentage for some nested class, as it was yelling at me before…

> we should also have a test verifying the error returned when…

Can do that. As a general observation, GitHub does not document error codes. Even when I have raised support requests and specifically complained that documentation neglects to specify the error code for some corner case, they have _rejected_ the request saying that their general guidelines mention error responses (when in fact those guidelines are so vague that they could easily be interpreted in several ways).

src/main/java/org/kohsuke/github/GHCheckRunBuilder.java
jglick: As elsewhere, I think it is pointless to even mention these sorts of limitations in Javadoc here, as the limits could be changed at any time and the code in this repository would become stale and misleading. (For example, some fields accept quite arbitrary-seeming lengths such as 20 characters.) It is the caller’s responsibility to follow GitHub’s current limits, and deal with whatever errors result if those are exceeded.

jglick: (The exception being the 50 annotation limit, since that has a documented and tested solution—adding annotations in batches of 50 or less—which can be implemented in the library.)

jglick:
BTW you need to revisit tests from #595, which have flaked twice in this PR’s builds. Deleting `Thread.sleep` calls causes test failures, which is a “bad code smell”.

PR #755, opened 2020-03-26
#724 broke the site goal from .github/PULL_REQUEST_TEMPLATE.md

jglick:
Guess you should check this in CI!


bitwiseman:
It's is run in CI, but apparently doesn't report failure?

jglick:
Or maybe the bad nesting of tags ceased to be a warning in JDK 11, which CI seems to run for this purpose; I was using JDK 8.

bitwiseman:
Yes, that is it.

PR #756, opened 2020-03-26
Shorten generated  test resource paths

bitwiseman:
# Description 
I ran into the windows path limit again and decided to make it less likely to be exceeded in future.

This change makes the file names for test resources even shorter and more readable but changing the from UUIDs to indexes. 

# Before submitting a PR:


jglick:
I am afraid this is going to clash with #753, not sure how to fix.

bitwiseman:
You can delete the test resource files and regenerate them or leave them as is. The test resources you generated should still work, they just have a different naming convention.

PR #767, opened 2020-04-01
Installation repository event

XiongKezhi:
# Description 
[InstallationEvent ](https://developer.github.com/v3/activity/events/types/#installationevent) and [InstallationRepositoriesEvent](https://developer.github.com/v3/activity/events/types/#installationrepositoriesevent).

# Before submitting a PR:


src/main/java/org/kohsuke/github/GHEventPayload.java
```diff
@@ -207,6 +208,164 @@ void wrapUp(GitHub root) {
         }
     }
 
+    /**
+     * Represents a repository object specifically for {@link Installation} and {@link InstallationRepositories}.
+     */
+    @SuppressFBWarnings(value = { "UWF_UNWRITTEN_FIELD" }, justification = "JSON API")
+    public static class InstallationRepository {
```
XiongKezhi, line 215: Construct a new object to represent the repository since the payload only contains a small part of `GHRepository`; it would to safer to use. But not so sure where we should put it.

bitwiseman: The pattern has been to reuse existing classes.  Then there are two options:
1. **eagerly populate all the items via `wrapup()/lateBind()`** - this results in a multiple api requests up front but returns stable fully populated objects
2. **only refresh item when user requests data that is missing from partial** - can save a ton of requests but adds complexity to target class (missing fields need to be guarded with checks for partial state. 

In this case, the partial is so thin you should probably go with option 1.

bitwiseman: @XiongKezhi 
Hm, but is it possible to get an event for a Repository what the repository is private and can't be populated using the current credentials?

src/main/java/org/kohsuke/github/GHEventPayload.java
```diff
@@ -207,6 +208,164 @@ void wrapUp(GitHub root) {
...
+
+        /**
+         * Gets repositories
+         *
+         * @return the repositories
+         */
+        public List<InstallationRepository> getRepositories() {
+            return repositories;
+        };
+    }
```
bitwiseman, line 306: @XiongKezhi  You'll want to add an override for `wrapUp(root)` to bind `GHAppInstallation`

src/main/java/org/kohsuke/github/GHEventPayload.java
XiongKezhi: > 
> 
> @XiongKezhi
> Hm, but is it possible to get an event for a Repository what the repository is private and can't be populated using the current credentials?

The repository installation events will be sent to the APP with [at least metadata read-only permission](https://developer.github.com/v3/apps/permissions/#metadata-permissions) to that repository, so shouldn't it be the users' responsibility if they use a client without the proper credentials instead of a client with the APP's JWT token?

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -2792,4 +2792,17 @@ public GHTagObject createTag(String tag, String message, String object, String t
...
                 .wrap(this);
     }
+
+    /**
+     * Repopulates this object.
+     *
+     * @throws java.io.IOException
+     *             The IO exception
+     */
+    public void refresh() throws IOException {
```
bitwiseman, line 2802: For now let's make this package private and call it `populate`:
```suggestion
    void populate() throws IOException {
```

src/main/java/org/kohsuke/github/GHEventPayload.java
```diff
@@ -300,9 +302,30 @@ public GHAppInstallation getInstallation() {
...
+                throw new IllegalStateException(
+                        "Expected check_suite payload, but got something else. Maybe we've got another type of event?");
+            else
+                installation.wrapUp(root);
+
+            if (repositories != null && !repositories.isEmpty()) {
+                try {
+                    for (GHRepository singleRepo : repositories) { // warp each of the repository
+                        singleRepo.wrap(root);
+                        singleRepo.refresh();
```
bitwiseman, line 262:
```suggestion
                        singleRepo.populate();
```

src/test/java/org/kohsuke/github/GHEventPayloadTest.java
```diff
@@ -405,7 +405,6 @@ public void InstallationRepositoriesEvent() throws Exception {
         assertThat(event.getRepositorySelection(), is("selected"));
 
         assertThat(event.getRepositoriesAdded().get(0).getId(), is(186853007L));
-        assertThat(event.getRepositoriesAdded().get(0).getNodeId(), is("MDEwOlJlcG9zaXRvcnkxODY4NTMwMDc="));
```
bitwiseman: Why remove these?

XiongKezhi: The previous `GHRepository` didn't have `node_id` property. Now I added it.

PR #773, opened 2020-04-05
Add ability to unset the milestone of an issue

### LINKED ISSUE #444: Milestones can not be unset
It is not possible to unset milestones.
To unset a PATCH with null for the milestone value needs to be sent.

However that is not possible due to two problems:
* [public void setMilestone(GHMilestone milestone)](https://github.com/kohsuke/github-api/blob/master/src/main/java/org/kohsuke/github/GHIssue.java#L207) throws a null pointer exception if null is passed.
A check like that solves this problem:
```
public void setMilestone(GHMilestone milestone) throws IOException {
        if(milestone != null){
            editIssue("milestone",milestone.getNumber());
        } else {
            editIssue("milestone",null);
        }

    }
```

* This leads to the second problem. The [editIssue ](https://github.com/kohsuke/github-api/blob/master/src/main/java/org/kohsuke/github/GHIssue.java#L181) method calls the method _with on the Requester which only adds arguments if the value of that argument is not null.

I solved this problem by having a editIssueNullable that calls withNullable instead of _with, but I am not sure if that is a good way to solve that problem.

martinvanzijl:
# Description
One can now unset the milestone of an issue using `issue.setMilestone(null)`.

Fixes #444.

Having the `editNullable` method alongside the `edit` method is a little ugly. Let me know if there is a better way to do this.


martinvanzijl:
I got an email saying the CI run failed (but seems successful above). Should I merge changes from latest master?

src/main/java/org/kohsuke/github/GHIssue.java
```diff
@@ -277,15 +284,19 @@ public void setBody(String body) throws IOException {
...
-     *             the io exception
+     *             The io exception
      */
     public void setMilestone(GHMilestone milestone) throws IOException {
-        edit("milestone", milestone.getNumber());
+        if (milestone == null) {
+            editNullable("milestone", null);
+        } else {
+            edit("milestone", milestone.getNumber());
+        }
```
bitwiseman, line 299:
```suggestion
        editNullable("milestone", null);
```

bitwiseman:
@martinvanzijl  
Re: Having the `editNullable`.  
It's not beautiful but I'm okay with it.

PR #774, opened 2020-04-05
Allow "refs/" prefix in parameter to GHRepository.getRef()

### LINKED ISSUE #518: Inconsistency between createRef and getRef
Currently createRef expects a fully qualified ref name ('refs/heads/master') while getRef prepends the 'refs/' and will fail if called with the same argument as createRef.
This is confusing, I suggest to align the APIs or at least prevent getRef from auto-adding 'refs/' to the path if the provided name already starts with 'refs/'.

martinvanzijl:
# Description 
Allows having the "refs/" prefix in parameter to `GHRepository.getRef()`. This is optional, so "refs/heads/master" will work, as well as "heads/master".

Fixes #518.


martinvanzijl:
There is a CI failure about "addCollaborators". From the log:

`[ERROR] addCollaborators(org.kohsuke.github.GHRepositoryTest)  Time elapsed: 0.046 s  <<< ERROR!`

I don't think I've changed this method, though.

src/test/resources/org/kohsuke/github/GHRepositoryTest/wiremock/addCollaborators/__files/orgs_github-api-test-org-a3d2b552-58b8-4028-b731-d00420496ca8.json
```diff
@@ -1,25 +0,0 @@
-{
```
bitwiseman, line 1: @martinvanzijl  It looks like you modified the test resources for this, whether you meant to or not.  This would be where the error is coming from.

martinvanzijl: @bitwiseman Sorry, my mistake. Should be fixed now.

PR #781, opened 2020-04-13
Deprecate set methods that should not be public

bitwiseman:
# Description 
These set methods should never have been made public. They serve no purpose. 

@timja @PauloMigAlmeida 

# Before submitting a PR:

### DIFF
--- a/src/main/java/org/kohsuke/github/GHApp.java
+++ b/src/main/java/org/kohsuke/github/GHApp.java
@@ -39,7 +39,9 @@ public class GHApp extends GHObject {
      *
      * @param owner
      *            the owner
+     * @deprecated Do not use this method. It was added due to incomplete understanding of Jackson binding.
      */
+    @Deprecated
     public void setOwner(GHUser owner) {
         this.owner = owner;
     }
@@ -58,7 +60,9 @@ public class GHApp extends GHObject {
      *
      * @param name
      *            the name
+     * @deprecated Do not use this method. It was added due to incomplete understanding of Jackson binding.
      */
+    @Deprecated
     public void setName(String name) {
         this.name = name;
     }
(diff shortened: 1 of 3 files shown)


src/main/java/org/kohsuke/github/GHApp.java
```diff
@@ -39,7 +39,9 @@ public GHUser getOwner() {
      *
      * @param owner
      *            the owner
+     * @deprecated Do not use this method. It was added due to incomplete understanding of Jackson binding.
```
timja, line 42: Looks like jackson code is handled here:

https://github.com/github-api/github-api/blob/796c644c4ac0f4637b6289ed10a1ac8654c7d7b4/src/main/java/org/kohsuke/github/GitHubClient.java#L738

I assume it's doing field injection in someway?

I'm used to either constructor or setter binding with jackson, haven't seen this setup before

PauloMigAlmeida:
I agree with the changes. Nice work @bitwiseman

src/main/java/org/kohsuke/github/GHApp.java
bitwiseman: @timja It's not exactly handled there.  Yes, the creation of the reader and adding of injected values happens there. 

Other than the value injection, binding happens just like it usually does.   
 
The point of this change is that the setter methods for Jackson properties do not need to be public. Jackson can access internal or private setters, and will even automatically bind fields that do not have setters or constructor paramters (regardless of access limitations).  See https://github.com/github-api/github-api/blob/master/src/main/java/org/kohsuke/github/example/dataobject/ReadOnlyObjects.java for some examples of what I've learned.

bitwiseman: @timja  I'm happy to continue discussing this topic, but it didn't seem like something that needed to block merging this change.

PR #782, opened 2020-04-14
Support issue event attributes

chids:
# Description 

For certain [issue events](https://developer.github.com/v3/issues/timeline/) there's a specific attribute present with details for that event type. For example, assigning an issue will include the `assignee` attribute which contains the user that was (un)assigned.

For such events these attributes are quite crucial as without them one can only determine that a label was added, a user was unassigned etc but not _which_ label, user and so on.

I tried to do something of a bare minimum change here in order to show what I feel is missing. Hence I deliberately didn't venture into trying to provide strong typing for each event for the sake of keeping the PR small.

Happy to elaborate/rewrite/restructure if another approach is preferred.


CONTRIBUTING.md
```diff
@@ -23,9 +23,7 @@ This the default behavior.
 
     `mvn install -Dtest.github.useProxy -Dtest=WireMockStatusReporterTest`
 
-4. The above should report no test failures and include the following console output:
-
-    `WireMockStatusReporterTest: GitHub proxying and user auth correctly configured for user login: <your login>`
+4. The above should report no test failures
```
bitwiseman, line 26: Hm, I wonder why that was changed.  I'll have to look into that.  Could you please file an issue for this and reverse this change?

chids: 👉 #785

bitwiseman:
@chids 
FYI, I didn't include this in 1.111 because that is a hotfix for a breaking issue.  Not a comment on the quality of this PR.  

Could you add a link to the timeline documentation you mentioned and to https://developer.github.com/v3/issues/events/ in the javadoc for the class.  

My main question to you is this: is it possible, reasonable, or helpful to do strong typing?  I'm not saying we should or shouldn't, I haven't looked at this part of the API yet.  I agree this is a reasonable minimal change but once we publish it, we'll need to maintain it going forward. I'd like you opinion, pros and cons.

chids:
> _Could you add a link to the timeline documentation you mentioned and to https://developer.github.com/v3/issues/events/ in the javadoc for the class._

👍, done.

> _My main question to you is this: is it possible, reasonable, or helpful to do strong typing? I'm not saying we should or shouldn't, I haven't looked at this part of the API yet. I agree this is a reasonable minimal change but once we publish it, we'll need to maintain it going forward. I'd like you opinion, pros and cons._

While poking around the code to explore how to support this I felt that strongly typing all events feels like asking for whack-a-mole maintenance since I'm assuming Github will expand on this whenever new features are being released.

That in combination with already having typed support for `GHLabel`, `GHMilestone` and `GHUser` such that you'll get a typed object for the attribute, if it's present, made me feel that this might not only be the simplest but also a quite pragmatic approach.

I mean this is pretty far down the API so to speak (`get a repo > get an issue > get its events > get the specific details of one such event`) that I personally feel it's fine for this library to assume that I know what I'm doing and throw a NPE in my face if I don't.

The one thing I did consider was to do the attributes as `Optional`'s but I didn't explore that.

chids:
To clarify: to strongly type this, since it's a list of mixed things, I imaging it'd be something like `List<GHIssueEvent>` where `GHIssueEvent` is abstract and subclassed for each event type, hence one would still have to do the is-this-a-label-event-dance in order to cast to a `GHIssueLevelEvent`. So I'm not convinced it'd add that much utility.

chids:
So, spent some time tinkering with this and ended up with a Consumer based approach that might strike an OK balance between types and ease of maintenance. See the last commits.

Regardless of what you think we should do here: if you're OK with creating a repo for issue events in `github-api-test-org` and add me as a collaborator to that repo I can create a test setup with issues, milestones and whatnot and use that for the testing, rather than my own repo 🙂

bitwiseman:
I agree with your reasoning to not strongly type these.  The whack-a-mole is kind of the nature of this project, but your point about `List<GHIssueEvent>` are a strong case IMO.  I've made you an owner of github-api-test-org.  Do whatever you need to.

> The one thing I did consider was to do the attributes as `Optional`'s but I didn't explore that.

What would this gain us? Making it clear what fields are universal versus required? Again I'm interested in your thoughts if you have time.  

Reason I ask: There are a number of places where the API returns incomplete versions of objects and we have to `populate()` the full version.  However, there are cases where detecting this is more difficult and error prone than I'd like.   (If/when we implement GraphQL support everything becomes optional.)

bitwiseman:
> So, spent some time tinkering with this and ended up with a Consumer based approach that might strike an OK balance between types and ease of maintenance. See the last commits.

Up to you.

chids:
> I've made you an owner of github-api-test-org. Do whatever you need to.

🙇

> > The one thing I did consider was to do the attributes as `Optional`'s but I didn't explore that.
> 
> What would this gain us? Making it clear what fields are universal versus required? Again I'm interested in your thoughts if you have time.

Merely as a signal that the value might not be there, again I think it's OK to assume the user knows what they're down this far down. Just something that struck me, not a fan myself.

> > So, spent some time tinkering with this and ended up with a Consumer based approach that might strike an OK balance between types and ease of maintenance. See the last commits.
> 
> Up to you.

I'd opt for the first approach just because it's simpler and most straight forward.

bitwiseman:
@chids 
I think I'd prefer to do the non-Consumer implementation for now.  I think I see where you're going with the consumer but it feels like something could be made available separately.

bitwiseman:
@chids  I saved your Consumer work in my fork at in a branch of the same name ( https://github.com/bitwiseman/github-api/tree/support-issue-event-attributes ) and force-pushed this PR back to the pre-Consumer state.

PR #791, opened 2020-04-18
Add global node_id to GHObject + GHTeam extends GHObject

ingwarsw:
# Description 
That PR is adding Global Node ID according to https://developer.github.com/v4/guides/using-global-node-ids/
Im trying to mix v3 and v4 (graphQL) of GH API together and without Global Node IDs its really hard.

Extending GHObject in GHTeam changed signature of ID, so its back incompatible change (Im not sure why GHTeam was not extending GHObject, in my opinion there is no reason for not doing it).

Please let me know whats best idea to test if it will work on all objects that extends GHObject..

# Before submitting a PR:

### DIFF
--- a/src/main/java/org/kohsuke/github/GHOrganization.java
+++ b/src/main/java/org/kohsuke/github/GHOrganization.java
@@ -127,7 +127,7 @@ public class GHOrganization extends GHPerson {
                 .withUrlPath(String.format("/orgs/%s/teams", login))
                 .toIterable(GHTeam[].class, item -> item.wrapUp(this));
     }
-
+    
     /**
      * Gets a single team by ID.
      *
@@ -144,6 +144,21 @@ public class GHOrganization extends GHPerson {
                 .withUrlPath(String.format("/organizations/%d/team/%d", id, teamId))
                 .fetch(GHTeam.class)
                 .wrapUp(this);
+    }
+    
+    /**
+     * Gets a single team by ID.
+     *
+     * @param teamId
+     *            id of the team that we want to query for
+     * @return the team
+     * @throws IOException
+     *             the io exception
+     *
+     * @deprecated Use {@link GHOrganization#getTeam(long)}
+     */
+    public GHTeam getTeam(int teamId) throws IOException {
+        return getTeam((long)teamId);
     }
 
     /**
--- a/src/main/java/org/kohsuke/github/GHOrganization.java
+++ b/src/main/java/org/kohsuke/github/GHOrganization.java
@@ -139,7 +139,7 @@ public class GHOrganization extends GHPerson {
      *
      * @see <a href= "https://developer.github.com/v3/teams/#get-team-by-name">documentation</a>
      */
-    public GHTeam getTeam(int teamId) throws IOException {
+    public GHTeam getTeam(long teamId) throws IOException {
         return root.createRequest()
                 .withUrlPath(String.format("/organizations/%d/team/%d", id, teamId))
                 .fetch(GHTeam.class)
--- a/src/main/java/org/kohsuke/github/GHOrganization.java
+++ b/src/main/java/org/kohsuke/github/GHOrganization.java
@@ -137,13 +137,11 @@ public class GHOrganization extends GHPerson {
      * @throws IOException
      *             the io exception
      *
-     * @see <a href= "https://developer.github.com/v3/teams/#get-team-by-name">documentation</a>
+     * @deprecated Use {@link GHOrganization#getTeam(long)}
      */
-    public GHTeam getTeam(long teamId) throws IOException {
-        return root.createRequest()
-                .withUrlPath(String.format("/organizations/%d/team/%d", id, teamId))
-                .fetch(GHTeam.class)
-                .wrapUp(this);
+    @Deprecated
+    public GHTeam getTeam(int teamId) throws IOException {
+        return getTeam((long) teamId);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


bitwiseman:
This project is not ready to make backward incompatible API changes.  When we do it, it will be a much larger change with a bunch of other clean ups. Please propose how you'd make this change without breaking api compatibility.

src/main/java/org/kohsuke/github/GHOrganization.java
```diff
@@ -139,7 +139,7 @@ public GHCreateRepositoryBuilder createRepository(String name) {
      *
      * @see <a href= "https://developer.github.com/v3/teams/#get-team-by-name">documentation</a>
      */
-    public GHTeam getTeam(int teamId) throws IOException {
+    public GHTeam getTeam(long teamId) throws IOException {
```
bitwiseman, line 142: We can create a deprecated version of this method and add new method.  We can't just change it.

src/main/java/org/kohsuke/github/GHTeam.java
```diff
@@ -129,15 +130,6 @@ public void setPrivacy(Privacy privacy) throws IOException {
         root.createRequest().method("PATCH").with("privacy", privacy).withUrlPath(api("")).send();
     }
 
-    /**
-     * Gets id.
-     *
-     * @return the id
-     */
-    public int getId() {
```
bitwiseman, line 137: We have a `BridgeMethods` to handle cases like these.  Preserving backward compatibility while changing method return types. 

https://github.com/github-api/github-api/blob/7d842175f7984e33b3d97346cba3b20a1f905031/src/main/java/org/kohsuke/github/GHObject.java#L75

ingwarsw: Seems like it works only at runtime..
IDE still shows errors when trying to use it to return `int`..

But i assume its ok..

ingwarsw:
@bitwiseman There is also class that have implemented node_id on its own..
I have changed it to extend GHObject but Im not sure if it should go here (or to extra PR)
here is changed code (I did not checked tests or anything so far for it)
https://github.com/ingwarsw/github-api/commit/f36f160056536ad287263863d78e596230847d4d

src/main/java/org/kohsuke/github/GHOrganization.java
```diff
@@ -139,13 +139,28 @@ public GHCreateRepositoryBuilder createRepository(String name) {
...
```
bitwiseman, line 142:
```
     * @see <a href= "https://developer.github.com/v3/teams/#get-team-by-name">documentation</a>
     * @deprecated Use {@link GHOrganization#getTeam(long)}
     */
    @Deprecated	     
    public GHTeam getTeam(int teamId) throws IOException {
```

src/main/java/org/kohsuke/github/GHOrganization.java
```diff
@@ -139,13 +139,28 @@ public GHCreateRepositoryBuilder createRepository(String name) {
...
+     *
+     * @param teamId
+     *            id of the team that we want to query for
+     * @return the team
+     * @throws IOException
+     *             the io exception
+     *
+     * @deprecated Use {@link GHOrganization#getTeam(long)}
+     */
+    public GHTeam getTeam(int teamId) throws IOException {
```
bitwiseman, line 160:
```suggestion
     */
    public GHTeam getTeam(long teamId) throws IOException {
```

src/main/java/org/kohsuke/github/GHTeam.java
bitwiseman: I'm not sure this is okay.  If I had: 
```
int a = team.getId();
```

Does it compile still with this change?

src/test/java/org/kohsuke/github/AppTest.java
```diff
@@ -291,7 +291,7 @@ public void testShouldFetchTeam() throws Exception {
         GHOrganization organization = gitHub.getOrganization(GITHUB_API_TEST_ORG);
         GHTeam teamByName = organization.getTeams().get("Core Developers");
 
-        GHTeam teamById = gitHub.getTeam(teamByName.getId());
+        GHTeam teamById = gitHub.getTeam((int) teamByName.getId());
```
bitwiseman, line 294: Right... so this means there will be build breaks for clients and we can't have that.

src/main/java/org/kohsuke/github/GHTeam.java
ingwarsw: ` incompatible types: possible lossy conversion from long to int`

src/test/java/org/kohsuke/github/AppTest.java
ingwarsw: That would mean we would have not to extend `GHObject`..
And thats actually really good idea..

Im not sure how bridge methods should work (but in java you cannot have 2 methods with same signature just different return type), but probably it just work at runtime..
At compile time it for sure gives error..

ingwarsw: If you want me to remove extension and implement node id in GHTeam without issues..

ingwarsw: @bitwiseman From code perspective I see that such "breaking" changes was already being mage few times (that changes are binary compatible because of that bridge methods)

See here
https://github.com/github-api/github-api/commit/b6e48cc4f9c7b13895673a7b4aa6399b7e236b1f
https://github.com/github-api/github-api/commit/edd9a2d5b6057a00afbe1d6c7cf329262713afc3

And GHObject was also moved from int to long not far away...
https://github.com/github-api/github-api/pull/388

So I think its safe enought to do it..

bitwiseman:
Added tests and did some reformatting.

ingwarsw:
@bitwiseman Thanks..

bitwiseman:
Will release the next version in the next week or so.  If you have any ideas on supporting GitHub API v4, I'd be interested to hear them.

ingwarsw:
@bitwiseman To connect to GH v4 API Im using "Allegro" framework (https://github.com/apollographql/apollo-android)

Calls itself are rather simple..
Only part that Is common to Github and I hat to write small wrapper over framework was paging and rate limits..

For now Im just using GraphQL to:
- Setup branch protections (v4 allows setting branch protection with rules not branch per branch so In my case for some repos I was able ro replace 500 rules for 5)
- Get list of all users with SAML emails.. to check with our AD and remove ones that are not there anymore.

PR #817, opened 2020-05-18
Consider header rate limit information for getRateLimit()

### LINKED ISSUE #802: getRateLimit() ignores header rate limit
**Describe the bug**
I use a GitHub Enterprise instance that does not support the `/rate_limit` endpoint (it returns a 404 error) but it _does_ return `X-RateLimit` headers.

`org.kohsuke.github.GitHubClient#getRateLimit` only checks the `/rate_limit` endpoint, and thus returns an "unknown" rate limit in this case, even though there is a configured rate limit, causing the client to potentially exceed that limit.

To add insult to injury, on our GHE instance calls to `/rate_limit` count against our rate limit, so we exceed the limit that much faster. 😄 I believe there are plans to address this, but I don't know the timeframe.

**Expected behavior**
`org.kohsuke.github.GitHubClient#getRateLimit` should return valid rate limit information if the server supports either the `/rate-limit` endpoint or the `X-RateLimit` headers.

Ideally the client would cache the knowledge of whether the `/rate_limit` endpoint is available, and fall back to exclusively using the headers in this case.

bitwiseman:
# Description 
When `getRateLimit()` call gets a 404 response, use the header rate limit information if it is available.  

Fixes #802

@timja Could you take a look at this? 

# Before submitting a PR:

### DIFF
--- a/src/main/java/org/kohsuke/github/GitHubClient.java
+++ b/src/main/java/org/kohsuke/github/GitHubClient.java
@@ -210,21 +210,31 @@ abstract class GitHubClient {
     /**
      * Gets the current rate limit from the server.
      *
+     * For some versions of GitHub Enterprise, the {@code /rate_limit} endpoint returns a {@code 404 Not Found}. In
+     * that, if {@link #lastRateLimit()} is not {@code null} and is not expired, it will be returned. Otherwise, a
+     * placeholder {@link GHRateLimit} instance with {@link GHRateLimit.UnknownLimitRecord}s will be returned.
+     *
      * @return the rate limit
      * @throws IOException
      *             the io exception
      */
+    @Nonnull
     public GHRateLimit getRateLimit() throws IOException {
-        GHRateLimit rateLimit;
+        GHRateLimit result;
         try {
-            rateLimit = fetch(JsonRateLimit.class, "/rate_limit").resources;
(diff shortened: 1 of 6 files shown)


src/main/java/org/kohsuke/github/GitHubClient.java
```diff
@@ -210,21 +210,31 @@ public boolean isAnonymous() {
...
-        GHRateLimit rateLimit;
+        GHRateLimit result;
         try {
-            rateLimit = fetch(JsonRateLimit.class, "/rate_limit").resources;
+            result = fetch(JsonRateLimit.class, "/rate_limit").resources;
         } catch (FileNotFoundException e) {
-            // GitHub Enterprise doesn't have the rate limit
-            // return a default rate limit that
-            rateLimit = GHRateLimit.Unknown();
+            // GitHub Enterprise doesn't have the rate_limit endpoint
```
timja, line 227: I don’t think this statement is right anymore at least on recent versions from reading the docs,

My understanding is that it is available but there’s a global flag for turning it off or on.

Having it off means that it will return a 404

bitwiseman, line 227:
```suggestion
            // For some versions of GitHub Enterprise, the rate_limit endpoint returns a 404.
```

PR #830, opened 2020-05-27
Handle header and endpoint rate limit responses consistently

### LINKED ISSUE #605: Integrate full rate limit information
#595 fixed an existing bug and enable retrieving rate limits aside from core, but it doesn't expose them and doesn't integrate them with the header update.

I think what we want to do here is unify header/non-header update into one `ratelimit` field on GitHub that is synchronized which updates `core` from header record but preserves the other records.  It will also need to check expiration on all the records in a rate limit instance.

bitwiseman:
# Description 
This change removes the separation between header rate limit information and the `/rate_limit` endpoint. All rate limit information is merged into a single `GHRateLimit` instance that is updated/replaced as needed.  If responses are processed out of order, records are updated to use the latest information. 

The response header rate limit information may not be present for some versions/configurations of GitHub Enterprise. When present for successful queries, the header rate limit information will also be present for requests that return a 404.  The response header rate limit information (when present) applies only to the target endpoint of the current request - the response to a request to the `/search` endpoint will have header information for the `search` record but not for any other endpoints.  Information from individual responses are used to update the appropriate record of the cached `GHRateLimit`. 

The `/rate_limit` endpoint returns either 404 or information for all four areas. When `/rate_limit` returns a 404, it may still have header information in the response which will apply to the `core` rate limit record (from #817). 

 

* `getRateLimit()` still always requests new information.  This change retains #817 and makes that behavior apply to other endpoints such as `/search` as well. 
* `lastRateLimit()` never returns `null`, instead defaulting to a `GHRateLimit` with all expired records.  
* Added an internal `GitHubClient.ratelimit(GitHubRateLimitSpecifier)` method, that requests new data only when the specified limit record has expired. 
* As regular records expire they are replaced by `UnknownRateLimitRecord`s.  
* Requests that apply to rate limits other than `Core` must specify their rate limit record manually via `GitHubRequest.Builder.ratelimit(GitHubRateLimitSpecifier)`. 
* This library does not currently use `GraphQL` or `IntegrationManifest` but when it does this change will work there as well. 

Fixes #605

# Before submitting a PR:


src/main/java/org/kohsuke/github/GHRateLimit.java
```diff
@@ -25,22 +26,21 @@
...
 public class GHRateLimit {
-
     /**
      * Remaining calls that can be made.
      *
      * @deprecated This value should never have been made public. Use {@link #getRemaining()}
      */
     @Deprecated
-    public int remaining;
+    public final int remaining;
```
bitwiseman, line 35: If there were people changing these values, I'm okay breaking them.

src/main/java/org/kohsuke/github/GHRateLimit.java
```diff
@@ -220,45 +247,107 @@ public int hashCode() {
...
-     * Has a large limit and long duration so that it will doesn't expire too often.
      *
      * @since 1.100
      */
     public static class UnknownLimitRecord extends Record {
 
-        // One hour
-        private static final long unknownLimitResetSeconds = 60L * 60L;
+        /**
+         * This is set to a somewhat small duration, rather than a long one.
```
bitwiseman, line 305: This is a notable change.

src/main/java/org/kohsuke/github/GHRateLimit.java
```diff
@@ -220,45 +247,107 @@ public int hashCode() {
...
+         *
+         * {@link GitHubClient#rateLimit(String)} will not query for a new rate limit until the current Record expires.
+         * When {@link GitHubClient} is initialized it includes a default {@link GHRateLimit#Default()} which has
+         * expired rate limit records. This guarantees that the first time {@link GitHubClient#rateLimit(String)} is
+         * called TODO
```
bitwiseman, line 310:
```suggestion
         * called the rate limit will refresh. 
```

src/main/java/org/kohsuke/github/GHRateLimit.java
```diff
@@ -220,45 +247,107 @@ public int hashCode() {
...
+         * Not for use outside this class
+         *
+         * @param resetEpochSeconds
+         *            the epoch second time when this record will expire.
+         */
+        private UnknownLimitRecord(long resetEpochSeconds) {
+            super(unknownLimit, unknownRemaining, resetEpochSeconds);
+        }
+
+        static synchronized Record current() {
```
bitwiseman, line 342: TODO comments.

src/main/java/org/kohsuke/github/GHRateLimit.java
bitwiseman: TODO javadoc

src/main/java/org/kohsuke/github/GHRateLimit.java
```diff
@@ -66,16 +66,43 @@
...
+
+    @Nonnull
+    static GHRateLimit Unknown(@Nonnull GitHubRateLimitSpecifier endpoint) {
+        return fromHeaderRecord(UnknownLimitRecord.current(), endpoint);
     }
 
     @Nonnull
-    static GHRateLimit fromHeaderRecord(Record header) {
-        return new GHRateLimit(header, new UnknownLimitRecord(), new UnknownLimitRecord(), new UnknownLimitRecord());
+    static GHRateLimit fromHeaderRecord(@Nonnull Record header, @Nonnull GitHubRateLimitSpecifier rateLimitSpecifier) {
```
bitwiseman, line 82: TODO javadoc

src/main/java/org/kohsuke/github/GHRateLimit.java
```diff
@@ -66,16 +66,43 @@
     private final Record integrationManifest;
 
     @Nonnull
-    static GHRateLimit Unknown() {
-        return new GHRateLimit(new UnknownLimitRecord(),
-                new UnknownLimitRecord(),
-                new UnknownLimitRecord(),
-                new UnknownLimitRecord());
+    static GHRateLimit Default() {
```
bitwiseman, line 69: TODO javadoc

src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -373,12 +373,19 @@ public String getApiUrl() {
     }
 
     /**
-     * Gets the current rate limit.
+     * Gets the current rate limit from the server.
+     *
+     * For some versions of GitHub Enterprise, the {@code /rate_limit} endpoint returns a {@code 404 Not Found}. In that
+     * case, if the most recent {@link GHRateLimit} will be returned.
```
bitwiseman, line 379: TODO javadoc

Mention header information

src/main/java/org/kohsuke/github/GitHubClient.java
```diff
@@ -210,69 +211,106 @@ public boolean isAnonymous() {
...
     public GHRateLimit getRateLimit() throws IOException {
+        return getRateLimit(GitHubRateLimitSpecifier.NONE);
+    }
+
+    @Nonnull
+    GHRateLimit getRateLimit(@Nonnull GitHubRateLimitSpecifier rateLimitSpecifier) throws IOException {
         GHRateLimit result;
         try {
-            result = fetch(JsonRateLimit.class, "/rate_limit").resources;
+            GitHubRequest request = GitHubRequest.newBuilder()
```
bitwiseman, line 233: TODO, consider making `fetch()` take a `builder` instead of a `urlPath`?

src/main/java/org/kohsuke/github/GitHubClient.java
```diff
@@ -512,58 +550,25 @@ private static boolean isInvalidCached404Response(GitHubResponse.ResponseInfo re
...
-        String resetString = responseInfo.headerField("X-RateLimit-Reset");
-        if (StringUtils.isBlank(resetString)) {
-            // if we are missing a header, return fast
-            return;
-        }
-
-        int limit, remaining;
-        long reset;
         try {
+            String limitString = Objects.requireNonNull(responseInfo.headerField("X-RateLimit-Limit"),
```
bitwiseman, line 554: There as a lot of trying not to throw.  Now we use exceptions like sane people.

src/main/java/org/kohsuke/github/GHRateLimit.java
```diff
@@ -221,44 +264,112 @@ public int hashCode() {
...
-     * <p>
-     * Has a large limit and long duration so that it will doesn't expire too often.
      *
      * @since 1.100
      */
     public static class UnknownLimitRecord extends Record {
 
-        // One hour
-        private static final long unknownLimitResetSeconds = 60L * 60L;
+        private static final long defaultUnknownLimitResetSeconds = Duration.ofSeconds(30).toMillis() / 1000;
```
timja, line 325: this seems a bit strange? 

```suggestion
        private static final long defaultUnknownLimitResetSeconds = Duration.ofSeconds(30).getSeconds();
```

bitwiseman: @timja 
There is a `Duration.toSeconds();` but only in Java 11+.

`getSeconds()` is not quite the same as `toSeconds()`, but you are correct for this case it can be considered the same.

src/main/java/org/kohsuke/github/GitHubRateLimitChecker.java
```diff
@@ -122,24 +136,26 @@ void checkRateLimit(GitHubClient client, GitHubRequest request) throws IOExcepti
...
-     *            the url path of the request
-     * @return the {@link RateLimitChecker} for a url path.
+     * @param rateLimitTarget
+     *            the rate limit to check
+     * @return the {@link RateLimitChecker} for a particular target
      */
     @Nonnull
-    private RateLimitChecker selectChecker(@Nonnull String urlPath) {
-        if (urlPath.equals("/rate_limit")) {
+    private RateLimitChecker selectChecker(@Nonnull RateLimitTarget rateLimitTarget) {
```
bitwiseman, line 146: Instead of trying to guess corresponding rate limit checker from the url path, depend on the requester to provide that information.  Much simpler.

src/main/java/org/kohsuke/github/GitHubClient.java
```diff
@@ -654,37 +641,6 @@ private boolean isPrivateModeEnabled() {
...
-     * {@link GHRateLimit.Record}s are never replaced by {@link GHRateLimit.UnknownLimitRecord}s. Candidates with
-     * resetEpochSeconds later than current record are more recent. Candidates with the same reset and a lower remaining
-     * count are more recent. Candidates with an earlier reset are older.
-     *
-     * @param candidate
-     *            {@link GHRateLimit.Record} constructed from the response header information
-     * @param current
-     *            the current {@link GHRateLimit.Record} record
-     */
-    static boolean shouldReplace(@Nonnull GHRateLimit.Record candidate, @Nonnull GHRateLimit.Record current) {
```
bitwiseman, line 671: Merge and equality logic moved into `GHRateLimit`.

PR #851, opened 2020-06-17
GHTeamBuilder#parentTeamId now accepts a long instead of an int

### LINKED ISSUE #850: GHTeamBuilder#parentTeamId should accept a long
**Describe the bug**

The identifier for a GHTeam returns a long, while the team builder accepts an int. This causes
a compile error if not handled.

**To Reproduce**

I detected this issue because while provisioning teams using the following snippet:

```java
public void example(GHTeam parentTeam){
organization.createTeam(teamName)
                .privacy(Privacy.CLOSED)
                .parentTeamId(parentTeam.getId())  <-- Problematic!
                .description(description)
                .create();
}
```

The following compile error is raised:

```text
java: incompatible types: possible lossy conversion from long to int
```

This is due to a recent change: fdcf74eaf27dd455e41af4150ea78649952b3b20 where GHTeam
extends GHObject, which causes an inheritance of the GHObject#getId() that is returning a long,
instead of the previous version of GHTeam#getId() that was returning an int.

**Expected behavior**

We should be able to "chain" the creation of a team with a direct call without the need for type conversion.

**Additional context**
I think that this should be as easy as GHTeamBuilder accepting a long instead of an int:

```diff
- public GHTeamBuilder parentTeamId(int parentTeamId);
+ public GHTeamBuilder parentTeamId(long parentTeamId);
```

MarcosCela:
Closes #850


MarcosCela:
First time using WireMock but I think that this should do the trick!

PR #900, opened 2020-07-25
Fixes modifyCollaborators for multiple users

### LINKED ISSUE #868: modifyCollaborators() fails with more than one user
https://github.com/hub4j/github-api/blob/8fd827040b29471f9fad942344e22cf8f5906e3c/src/main/java/org/kohsuke/github/GHRepository.java#L980

The URL path in the Requester object gets re-used so the first user is successfully processed, but subsequent users fail because the URL pat still includes the previous user in addition to the current user in the loop.

```
Given a list of two users: 'userA', 'userB'
Loop1 Path: /repos/the-org/the-repo/collaborators/userA
Loop2 Path: /repos/the-org/the-repo/collaborators/userA/repos/the-org/the-repo/collaborators/userB
```

There's a unit test for the method, but this case isn't covered since it tests with a single-user List.

gastaldi:
Fixes #868

# Description 
This PR makes sure that the Requester instance is not reused per user in the modifyCollaborators method

# Before submitting a PR:

- [X] Run `mvn clean compile` locally. This may reformat your code, commit those changes.

# When creating a PR: 


bitwiseman:
@gastaldi 
Thanks for the PR! 

I liked your formatting changes but they are separate from this PR, so I pulled them out and merged them first. 
Next time please put formatting changes unrelated to the current change in a separate commit (or even a separate PR).  

Your fix is perfectly good, but I made me realize that the `withUrlPath()` method change I made a while back was a bad idea.  I pulled that out.  

Good PR, and thanks again for contributing!

PR #909, opened 2020-07-30
GHCheckRun.getPullRequests public

### LINKED ISSUE #908: GHCheckRun's getPullRequests should be public
**Describe the bug**
The check run event class does not expose pullrequests as can be seen in https://github.com/hub4j/github-api/blob/bffa78c1b8ffb87dca87337259b636d53dc8841a/src/main/java/org/kohsuke/github/GHCheckRun.java#L110

**To Reproduce**
N/A

**Expected behavior**
This should be made public so that consumers can get a check's pull requests list.

**Additional context**
N/A

JKalash:
# Description 
Closes #908 (GHCheckRun's getPullRequests should be public)

# Before submitting a PR:

# When creating a PR: 


src/main/java/org/kohsuke/github/GHCheckRun.java
```diff
@@ -29,7 +30,7 @@
     private URL detailsUrl;
     private Output output;
     private GHApp app;
-    private GHPullRequest[] pullRequests;
+    private List<GHPullRequest> pullRequests;
```
bitwiseman, line 33: I think I'd rather keep the array here and have `getPullRequests()` return an `UnmodifiableList` created from it.  
Or maybe return an `Iterable`.  

I don't think there's a strong reason to change to a List, but I'm open to discussing it.

JKalash: I do not have any strong opinions on the need to return an `UnmodifiableList` / `Iterable`  vs changing this to a `List`. Will update 😄

src/main/java/org/kohsuke/github/GHCheckRun.java
```diff
@@ -106,14 +109,17 @@ public String getHeadSha() {
      * Gets the pull requests participated in this check run.
      *
      * @return Pull requests of this check run
+     * @throws IOException
+     *             the io exception
      */
-    GHPullRequest[] getPullRequests() throws IOException {
+    public Iterator<GHPullRequest> getPullRequests() throws IOException {
```
bitwiseman, line 115: Sorry, I mispoke.  We want to return an `Iterable`.

```suggestion
    public Iterable<GHPullRequest> getPullRequests() throws IOException {
```

JKalash: Ah right, I guess instead of having the method signature be `Iterable` we can set it to `List` specifically? Updated as such.

JKalash:
@bitwiseman is this fine to merge now?

bitwiseman:
@JKalash  
You didn't have any tests so I thought I'd try to make some.  It took a bit more work than I thought - it uncovered some bugs.

src/main/java/org/kohsuke/github/GHCheckRun.java
```diff
@@ -105,15 +128,22 @@ public String getHeadSha() {
     /**
      * Gets the pull requests participated in this check run.
      *
-     * @return Pull requests of this check run
+     * Note this field is only populated for events. When getting a {@link GHCheckRun} outside of an event, this is
+     * always empty.
```
bitwiseman, line 132: @JKalash 
Keep this in mind.  This is took me a while to figure out. I'm not sure if it is expected behavior from GitHub but it is consistent.

PR #939, opened 2020-09-08
- bulk update of repository options

jgangemi:
# Description 
allows bulk updating of repository settings instead of making individual calls for each.

# Before submitting a PR:

# When creating a PR: 


jgangemi:
any updates on getting this merged?

jgangemi:
bueller? as someone who has contributed to this project before, it's super frustrating to do this work and then see the PR languish and other things merged ahead of it.

jgangemi:
@bitwiseman hi! what needs to be done on my end to get some traction on this?

bitwiseman:
@jgangemi  
Sorry for delay, I was on leave.  I've just cut a release, but that is not a comment on this change.  

This is a solid, high-quality change. Thanks for including tests and javadoc.  Great work!

I've invited you to the `hub4j-test-org`.  Please join and rerecord test files using the org.   

Also, as long as you're going to do this level of work, I'd be grateful if you would try implementing this using [`AbstractBuilder` ](https://github.com/hub4j/github-api/blob/master/src/main/java/org/kohsuke/github/AbstractBuilder.java) framework that I used in [`GHLabel`](https://github.com/hub4j/github-api/blob/master/src/main/java/org/kohsuke/github/GHLabel.java).  You'll end up with a little smaller set of changes and gain the use o the `update()/set()` batching behavior from `AbstractBuilder`.  With the inheritance the way you have it now, I think what you have is pretty close already.

src/test/java/org/kohsuke/github/AbstractGitHubWireMockTest.java
```diff
@@ -244,6 +260,20 @@ protected void kohsuke() {
         // assumeTrue(login.equals("kohsuke") || login.equals("kohsuke2"));
     }
 
+    private GHCreateRepositoryBuilder getCreateBuilder(String name, boolean useOrg) throws IOException {
+        GitHub github = getGitHubBeforeAfter();
+
+        if (useOrg) {
```
bitwiseman, line 266: Instead of having this be a parameter that gets passed down through the call stack, this should be a setting like `mockGitHub.isUseProxy()` but on `AbstractGitHubWireMockTest`, something like `isTestWithUserAccount()`. This will let users create, debug, and record tests using their account, but also easily turn it off and re-record from the test account later.

jgangemi:
@bitwiseman no worries. i'll take a look at what you suggested and get going on these changes.

src/test/java/org/kohsuke/github/AbstractGitHubWireMockTest.java
jgangemi: i feel like i followed this from some other example but i can see about addressing this.

jgangemi:
@bitwiseman no worries. i'll take a look at what you suggested and get going on these changes.

jgangemi:
@bitwiseman sorry for the delay on my end, changes made. let me know your feedback.

i set the default behavior to use the "org" to run tests and to set a property for a user account. i updated the `contributing` doc accordingly.

Issue #941, opened 2020-09-21
Github apps token should be automatically refreshed


MarcosCela:
**Describe the bug**

Currently, when we authenticate GitHub as an application, as per the official documentation:

- github-api: https://github-api.kohsuke.org/githubappappinsttokenauth.html

Example code:
```java
GitHub gh = new GitHubBuilder().withJwtToken(jwtToken).build();
GHAppInstallation installation = gh.getApp().getInstallationByOrganization("myorg");
GHAppInstallationToken ghAppInstallationToken = installation.createToken().create();
// At this point, ghAppInstallationToken should be valid to authenticate as an app
// installation for a given organization. The problem is that it will eventually expire
// in ~1hour
return new GitHubBuilder()
.withAppInstallationToken(ghAppInstallationToken.getToken())
.withJwtToken(jwtToken)
.withConnector(connector)
.build();
```

The problem with this is that the token is only valid for a while (1 hour in my tests),
and we would need to update it depending on the expiration time that can
be obtained in:

```java
 ghAppInstallationToken.getExpiresAt();
```

**To Reproduce**

Create a GitHub application, and perform an installation. Obtain the application installation token,
and use it. Approximately 1 hour later (for my manual tests) the token will expire and it will be
no longer valid.

**Expected behavior**

Once the GitHub builder /GitHub object is configured with a valid token, it should
automatically refresh the Token. Alternatively, a refresh mechanism should be provided
to manually refresh credentials.

**Discussion**

- Should the library provide a way of manually refreshing the Token, or instead automatically refresh it?

I can take care of this once we achieve some basic points of what would be the ideal outcome.

bitwiseman:
@MarcosCela 
Automatic refresh would be nice for most users. This could be done similar to [`RateLimitChecker`s](https://github.com/hub4j/github-api/blob/master/src/main/java/org/kohsuke/github/GitHubBuilder.java#L397). Instead of accepting a `jwtToken` we could accept a `CredentialProvider`.   Before each call to the server, the `GitHubClient` would ask the provider for a token, and provider would provide a current token, updating as needed.

[...]

PR #945, opened 2020-09-25
Feat/credential provider refresh

MarcosCela:
# Description 

This PR attempts to solve: #941


MarcosCela:
@bitwiseman Hi Liam!
Just as you said in #941, this adds a CredentialProvider. Right now the changes don't really do much. I still have to force the lookup of the credentials for each request (not just at the initial setup), and then implement the refresh logic for the provider.
Will keep the PR updated, feedback is welcome!

bitwiseman:
@MarcosCela 

If I understand correctly, we'll need to do the second approach in order for refresh to work for longer than the expiration time of the JWT Token.  And JWT Tokens have a shorter life span than the GitHub App Tokens.

src/main/java/org/kohsuke/github/GitHubClient.java
```diff
@@ -112,25 +108,26 @@
         this.connector = connector;
 
         if (oauthAccessToken != null) {
-            encodedAuthorization = "token " + oauthAccessToken;
+            this.credentialProvider = new ImmutableCredentialProvider(String.format("token %s", oauthAccessToken));
```
bitwiseman, line 111:
```suggestion
            this.credentialProvider = ImmutableCredentialProvider.fromOAuthToken(String.format("token %s", oauthAccessToken));
```

src/main/java/org/kohsuke/github/GitHubClient.java
```diff
@@ -112,25 +108,26 @@
...
             if (jwtToken != null) {
-                encodedAuthorization = "Bearer " + jwtToken;
+                this.credentialProvider = new ImmutableCredentialProvider(String.format("Bearer %s", jwtToken));
             } else if (password != null) {
                 String authorization = (login + ':' + password);
                 String charsetName = StandardCharsets.UTF_8.name();
-                encodedAuthorization = "Basic "
+                String encodedAuthorization = "Basic "
                         + Base64.getEncoder().encodeToString(authorization.getBytes(charsetName));
+                this.credentialProvider = new ImmutableCredentialProvider(encodedAuthorization);
```
bitwiseman, line 120:
```suggestion
this.credentialProvider = ImmutableCredentialProvider.fromLoginAndPassword(login, password);
```

src/main/java/org/kohsuke/github/GitHubClient.java
```diff
@@ -112,25 +108,26 @@
...
             } else {// anonymous access
-                encodedAuthorization = null;
+                this.credentialProvider = new ImmutableCredentialProvider(null);
```
bitwiseman, line 122:
```suggestion
                this.credentialProvider = ImmutableCredentialProvider.ANONYMOUS;
```

src/main/java/org/kohsuke/github/GitHubClient.java
bitwiseman: I would also suggest moving all these into `GitHubBuilder` and have a method `GitHubBuilder.withCredentialProvider(CredentialProvider)` which can be internal for now.  Then this constructor just sets `this.credentialProvider = credentialProvider` pass from the builder.

MarcosCela:
> @MarcosCela
> 
> If I understand correctly, we'll need to do the second approach in order for refresh to work for longer than the expiration time of the JWT Token. And JWT Tokens have a shorter life span than the GitHub App Tokens.

Sorry, bad explanation. As I see it right now we have 2 choices:
- Users of the library provide a JWTTokenProvider that abstracts the creation of the JWT token from the library. We would not need to include any external libraries.
- We implement the creation of the JWT token in the library, and users will just provide an application ID and a private key. This will need to include some external libraries:
```xml
 <dependency>
      <groupId>io.jsonwebtoken</groupId>
      <artifactId>jjwt-api</artifactId>
    </dependency>
    <dependency>
      <groupId>io.jsonwebtoken</groupId>
      <artifactId>jjwt-impl</artifactId>
    </dependency>
    <dependency>
      <groupId>io.jsonwebtoken</groupId>
      <artifactId>jjwt-jackson</artifactId>
    </dependency>
```

Will address the other issues, thank you for your time and insight!

MarcosCela:
As it's right now, an example on how to use this will be as follows:

- First, you create a GitHubBuilder and configure it as follows:

```java
 GitHub installation(
        HttpConnector connector,
        JWTTokenProvider jwtTokenProvider) throws IOException {
        
         return new GitHubBuilder()
            .withConnector(connector)
            .withCredentialProvider(new OrgInstallationCredentialProvider("myOrg", jwtTokenProvider)).build();
    }
```


Where OrgInstallationCredentialProvider is an Oauth2 Token provider that uses another JWTTokenProvider. With the JWT token, we request a new installation token when the current one has expired. Users need to provide a JWTTokenProvider implementation. E.g:


```java
@Slf4j
@Component
@RequiredArgsConstructor
public class JWTTokens implements JWTTokenProvider {

    private static final long MINUTES_10 = Duration.buildByMinutes(10).getMilliseconds();


    private final PrivateKey privateKey;



    @Override
    public String token() {
        log.debug("Getting a new JWT token");
        return getJWT();
    }


    public String getJWT() {
        long nowMillis = System.currentTimeMillis();
        Date now = new Date(nowMillis);


        // Let's set the JWT Claims
        JwtBuilder builder = Jwts.builder()
            .setIssuedAt(now)
            .setIssuer("myApplicationId")
            .signWith(privateKey, SignatureAlgorithm.RS256);

        // if it has been specified, let's add the expiration
        if (MINUTES_10 > 0) {
            long expMillis = nowMillis + MINUTES_10;
            Date exp = new Date(expMillis);
            builder.setExpiration(exp);
        }

        // Builds the JWT and serializes it to a compact, URL-safe string
        return builder.compact();
    }

}
```


We could simplify this and provide the creation of the JWT token from a given private key and applicationId, but it would mean to add dependencies to the library. I think that it's better to let users chose what kind of JWT libraries they use.

We are of course missing tests.
What do you think?

bitwiseman:
@MarcosCela 
Adding dependencies - you could follow the pattern of putting an implementation of the CredentialsProvider in the `extras` package and make JWT an optional dependency.  That would let users choose to use that provider, letting this project provide a tested stable implementation, while also not forcing people to take dependencies they don't need.

MarcosCela:
will check that out, I was not aware of that extras thingy. Thank you!

bitwiseman:
> I think that it's better to let users chose what kind of JWT libraries they use.

Are there a lot of different JWT token libraries?  

If there is a specific and clear advantage to not generating the JWT token inside this library it is completely fine to leave that to the user.  But as I understand it, the only reason users get JWT tokens is to immediately turn around and get GitHub App credential.  If that is the case, why force them to think about it?  With the way you've implemented it, looks like you could do both. Allow people to have control AND provide a default implementation in the `extras` package the people could use if they don't need that control.

bitwiseman:
@MarcosCela 
This is looking very good.  Having token generation handled outside of the `GitHub` class is great.

src/main/java/org/kohsuke/github/GitHubBuilder.java
```diff
@@ -278,6 +279,11 @@ public GitHubBuilder withOAuthToken(String oauthToken, String user) {
         return this;
     }
 
+    public GitHubBuilder withCredentialProvider(final CredentialProvider credentialProvider) {
```
bitwiseman, line 282: I'd like to see all the other credential methods call this one. 

For example: 
```
    public GitHubBuilder withOAuthToken(String oauthToken, String user) {
        withCredentialProvider(ImmutableCredentialProvider.fromOauthToken(oauthToken));
        this.user = user;
        return this;
    }
```

That would let us remove the `oauthToken` and `password` instance fields from the class (and down the call chain).

src/main/java/org/kohsuke/github/GitHubClient.java
```diff
@@ -111,33 +93,45 @@
...
+    @Nullable
+    private String getCurrentUser(String login, String jwtToken, Consumer<GHMyself> myselfConsumer) throws IOException {
+        if (login == null && this.credentialProvider.getEncodedAuthorization() != null && jwtToken == null) {
+            try {
+                GHMyself myself = fetch(GHMyself.class, "/user");
+                if (myselfConsumer != null) {
+                    myselfConsumer.accept(myself);
+                }
+                return myself.getLogin();
+            } catch (IOException e) {
```
bitwiseman, line 129: Why add this `try-catch`?  It changes the user visible behavior.

MarcosCela: If we use a credentialProvider, we no longer know what kind of credentials we have here. Some credentials don't support checking who is making the call (api call to "/user"), and will fail, so the only way that I see to check if the credentials being used support "user lookup", is to actually make the call and handle the case where the call fails (e.g: installation tokens).

src/main/java/org/kohsuke/github/GitHubBuilder.java
MarcosCela: If we do this we lose the ability to know what kind of credentials we are using, and we will have to (I think) request the user to know if the current credentials are "user-aware" (GET "/user"). This will break a ton of tests that don't have the mapping for that call (not a big deal IMO), but also add another API call on GHBuilder#build(), and this does change semantics. We could potentially make it "lazy", so that we only lookup the user when it's actually requested, and not on setup, but I have to look into this.

MarcosCela:
> > I think that it's better to let users chose what kind of JWT libraries they use.
> 
> Are there a lot of different JWT token libraries?

No that I am aware of, but let's say that you use Spring boot (it does have JWT modules), and don't want to use the ones in the example...I find it a little bit silly to also add 3 dependencies for this little functionality, I really like your idea of providing an optional implementation in the "extras" and declare the dependencies as required.

MarcosCela:
Added examples on the extra package, ready for review and for merging.

bitwiseman:
@MarcosCela 
Sorry for the delay, I was offline for a bit.  I'll look at this this week.

MarcosCela:
No worries, thanks for your time!!!

MarcosCela:
Hi again! any chance that we can squeeze this into the next release? can I help somehow?

src/main/java/org/kohsuke/github/GitHubBuilder.java
bitwiseman: > If we do this we lose the ability to know what kind of credentials we are using

I feel like this isn't true.  There should be some way to do this. I might not be explaining myself well though.

src/main/java/org/kohsuke/github/GitHubClient.java
bitwiseman: > If we use a credentialProvider, we no longer know what kind of credentials we have here. 

What about having a method on `CredentialsProvider`?  `getLogin(Consumer<GHMyself> myselfConsumer)` or something similar?  Then `getCurrentUser()` wouldn't check `jwtToken` or `login` or whatever, the credential itself would decide what to do.  `getLogin()` on `JWTTokenProvider` would always return `null`, while for `fromOauthToken` it would have code from lines 124-128.

bitwiseman: Hmm, or maybe a `CredentialsProvider` would have `boolean requiresMyself()` to control whether this code is called and then `getLogin(@CheckForNull GHMyself)` that is always called, `return credentialProvider.getLogin(myself);`.

src/main/java/org/kohsuke/github/extras/auth/OrgInstallationCredentialProvider.java
```diff
@@ -0,0 +1,70 @@
...
+import java.nio.file.Paths;
+import java.security.NoSuchAlgorithmException;
+import java.security.spec.InvalidKeySpecException;
+import java.util.Date;
+
+/**
+ * This helper class provides an example on how to authenticate a GitHub instance with an installation token, that will
+ * be automatically refreshed when required.
+ */
+public class OrgInstallationCredentialProvider implements CredentialProvider {
```
bitwiseman, line 15:
```suggestion
public class OrgInstallationCredentialProvider extends org.kohsuke.github.OrgInstallationCredentialProvider implements CredentialProvider {
```

src/main/java/org/kohsuke/github/extras/auth/JWTTokenProvider.java
```diff
@@ -0,0 +1,83 @@
...
+    private PrivateKey loadPrivateKey(Path keyPath)
+            throws NoSuchAlgorithmException, InvalidKeySpecException, IOException {
+
+        byte[] keyBytes = Files.readAllBytes(keyPath);
+        PKCS8EncodedKeySpec spec = new PKCS8EncodedKeySpec(keyBytes);
+        KeyFactory kf = KeyFactory.getInstance("RSA");
+        return kf.generatePrivate(spec);
+    }
+
+    public String getJWT() {
```
bitwiseman, line 57:
```suggestion
    private String getJWT() {
```

src/test/java/org/kohsuke/github/OrgInstallationCredentialProviderTest.java
```diff
@@ -0,0 +1,32 @@
...
+        gitHub.getClient().credentialProvider = ImmutableCredentialProvider.fromJwtToken("myToken");
+
+        OrgInstallationCredentialProvider provider = new OrgInstallationCredentialProvider("testOrganization", gitHub);
+
+        provider.getEncodedAuthorization();
+    }
+
+    @Test
+    public void validJWTTokenAllowsOauthTokenRequest() throws IOException {
+        gitHub.getClient().credentialProvider = ImmutableCredentialProvider.fromJwtToken("valid-token");
```
bitwiseman, line 22: Are the credentials returned from `fromJwtToken` actually immutable?  I thought the token expired every 10 minutes.

src/main/java/org/kohsuke/github/extras/auth/OrgInstallationCredentialProvider.java
```diff
@@ -0,0 +1,70 @@
...
+
+    private final GitHub gitHub;
+
+    private final String organizationName;
+
+    private String latestToken;
+
+    private Date validUntil;
+
+    public OrgInstallationCredentialProvider(String organizationName, GitHub gitHub) {
```
bitwiseman, line 25: Having this take a `GitHub` at construction time makes it more difficult to work with, but it is probably something we can tweak/improve later.

bitwiseman:
@MarcosCela 
I'd really like to get those other parameters  (`jwttoken`) out of the constructors for `GitHub` and `GitHubClient`. I've pointed in a direction I think will work.  If you can take a swing at it that would be great.  If not, I'll try it as soon as I have time.  

I could be wrong and this really doesn't make sense.  If so just tell me, but I want to make sure we've really tried before we give up on it.  

Thanks for your hard work and patience.

src/main/java/org/kohsuke/github/extras/auth/OrgInstallationCredentialProvider.java
MarcosCela: I will try to give it a go following your suggestions, thanks a lot!!!!

bitwiseman:
@MarcosCela 
Cloud you take a look at the changes I've made and make sure the make sense to you?

MarcosCela:
> @MarcosCela
> Cloud you take a look at the changes I've made and make sure the make sense to you?

Definetly, I will take a look as soon as I can, thanks again for your time and dedication!!

MarcosCela:
This looks extremely promising, to perform automatic access token refresh, I can do something like this:

```java
    CredentialProvider prov = new JWTTokenProvider("applicationId", new File("/my/path.extension"));

        GitHub jwtTokenRequester = new GitHubBuilder()
            .withCredentialProvider(prov)
            .build();
        
        GitHub gh = new GitHubBuilder().withCredentialProvider(() -> {
            // Missing all the refresh token logic, this is requesting a new token all the time (for simplicity)
            return jwtTokenRequester.getApp().getInstallationById("installationId").getAccessTokenUrl();
        }).build();
        
        // From now on, I will just use "gh" for the rest of my program logic, that defers to the inlined
        // credential provider, that uses a second GitHub instance "jwtTokenRequester" to refresh the access token
        // sounds convoluted
```

With your changes I think that the client is extremely flexible in terms of authentication, looks finished to me!
Perhaps I can add an example of an implementation for the "inline" credential provider that requests an installation
token in the extras package?
Thanks again for your time!

bitwiseman:
@MarcosCela 

I wouldn't use this as code as the example.  I really don't want people to do it the way you've shown above. I want them to implement the `bind(GitHub)` method.  I definitely do not want people to inline credential providers.  Maybe the right thing to do is have a separate interface or base class?  Or make `bind(GitHub)` not `default`? 

```java
    CredentialProvider jwtProvider = new JWTTokenProvider("applicationId", new File("/my/path.extension"));

    GitHub gh = new GitHubBuilder().withCredentialProvider(new OrgInstallationCredentialProvider("testOrganization", jwtProvider)).build();
```

Oh, we need to change `OrgInstallationCredentialProvider` to `AppInstallationCredentialProvider` (or at least `OrgAppInstallationCredentialProvider`).

Also, what do you think of changing `CredentialProvider` to `AuthorizationProvider` as that is more accurate to what these are doing now - they are providing the `Authorization` header value.

MarcosCela:
> @MarcosCela
> 
> I wouldn't use this as code as the example. I really don't want people to do it the way you've shown above. I want them to implement the `bind(GitHub)` method. I definitely do not want people to inline credential providers. Maybe the right thing to do is have a separate interface or base class? Or make `bind(GitHub)` not `default`?
> 
> ```java
>     CredentialProvider jwtProvider = new JWTTokenProvider("applicationId", new File("/my/path.extension"));
> 
>     GitHub gh = new GitHubBuilder().withCredentialProvider(new OrgInstallationCredentialProvider("testOrganization", jwtProvider)).build();
> ```
> 
> Oh, we need to change `OrgInstallationCredentialProvider` to `AppInstallationCredentialProvider` (or at least `OrgAppInstallationCredentialProvider`).
> 
> Also, what do you think of changing `CredentialProvider` to `AuthorizationProvider` as that is more accurate to what these are doing now - they are providing the `Authorization` header value.

100% agree on the inlining, it should have a different interface/class class. I will proceed to rename the name to AuthorizationProvider, and all implementations if required (e.g: AppInstallationAuthorizationProvider).
I will also give it another go and see what will be the easiest way to perform the automatic refresh with the bind method, I tried it before but could not understand it fully, will give it another go!
Thanks again!

MarcosCela:
I applied your changes and tested it (successfully!) with the following snippet:


```java
 AuthorizationProvider jwt = new JWTTokenProvider("12345",
            Paths.get("/location/to/my/key"));

        OrgAppInstallationAuthorizationProvider orgAppInstallationAuthorizationProvider = new OrgAppInstallationAuthorizationProvider(
            "my-organization", jwt);

        GitHub gh = new GitHubBuilder().withAuthorizationProvider(orgAppInstallationAuthorizationProvider)
            .build();
        System.out.println("gh.getRateLimit() = " + gh.getRateLimit());
```

Is this the way that you were referring to?

src/main/java/org/kohsuke/github/extras/auth/JWTTokenProvider.java
```diff
@@ -0,0 +1,132 @@
...
+        this.applicationId = applicationId;
+    }
+
+    @Override
+    public String getEncodedAuthorization() throws IOException {
+        synchronized (this) {
+            if (Instant.now().isAfter(validUntil)) {
+                token = refreshJWT();
+            }
+            return String.format("Bearer %s", token);
```
bitwiseman, line 65: @MarcosCela 
Could you add a test that verifies this?

src/test/java/org/kohsuke/github/AbstractGHAppInstallationTest.java
```diff
@@ -63,15 +83,15 @@ private GHAppInstallation getAppInstallationWithToken(String jwtToken) throws IO
...
     }
 
     protected GHAppInstallation getAppInstallationWithTokenApp2() throws IOException {
-        return getAppInstallationWithToken(createJwtToken(PRIVATE_KEY_FILE_APP_2, TEST_APP_ID_2));
+        return getAppInstallationWithToken(JWT_PROVIDER_2.getEncodedAuthorization());
     }
 
     protected GHAppInstallation getAppInstallationWithTokenApp3() throws IOException {
-        return getAppInstallationWithToken(createJwtToken(PRIVATE_KEY_FILE_APP_3, TEST_APP_ID_3));
+        return getAppInstallationWithToken(JWT_PROVIDER_3.getEncodedAuthorization());
```
bitwiseman, line 94: @MarcosCela 
To test JWT provider, maybe just add an assertion to the test that uses these.

src/main/java/org/kohsuke/github/extras/auth/JWTTokenProvider.java
MarcosCela: Will do!

src/test/java/org/kohsuke/github/AbstractGHAppInstallationTest.java
MarcosCela: Tests added, what do you think?

src/test/java/org/kohsuke/github/extras/auth/JWTTokenProviderTest.java
```diff
@@ -0,0 +1,48 @@
...
+     *
+     * <pre>
+     * Authorization: Bearer ey{rest of the header}.{payload}.{signature}
+     * </pre>
+     *
+     * Matched by the regular expression:
+     *
+     * <pre>
+     * ^Bearer (?<JWTHeader>ey\S*)\.(?<JWTPayload>\S*)\.(?<JWTSignature>\S*)$
+     * </pre>
```
bitwiseman: Very nice.

bitwiseman (approved):

src/main/java/org/kohsuke/github/extras/authorization/JWTTokenProvider.java
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Convert a PKCS#8 formatted private key in string format into a java PrivateKey
+     *
+     * @param key
+     *            PCKS#8 string
+     * @return private key
+     * @throws GeneralSecurityException
+     *             if we couldn't parse the string
+     */
+    private static PrivateKey getPrivateKeyFromString(final String key) throws GeneralSecurityException {
```
timja, line 88: any reason this is private? seems like I just have to re-implement this / copy it if I'm working with strings

bitwiseman: Probably better to create a constructor that takes a `String`.

timja: yes or that...

PR #949, opened 2020-09-26
List repositories for a GHAppInstallation & support for items nested in a field

### LINKED ISSUE #948: Implement API method for retrieving repositories under a GHAppInstallation
The API method allowing to retrieve all of the repositories that are assigned to a GHAppInstallation is currently not implemented. The API doc is [here](https://docs.github.com/en/free-pro-team@latest/rest/reference/apps#list-repositories-accessible-to-the-app-installation).

When building GitHub Apps, it can be necessary to list all the installations of the app and then their associated repositories for further processing.

tginiotis-at-work:
# Description 
Implements the API method for listing repositories associated to an app installation. Docs [here](https://docs.github.com/en/free-pro-team@latest/rest/reference/apps#list-repositories-accessible-to-the-app-installation).
Resolves #948.

# Before submitting a PR:

# When creating a PR: 


tginiotis-at-work:
The issue with this API method is that its response structure is unlike other methods. Specifically, it returns the pull request list in a `repositories` field inside the body of the response JSON instead of just putting the list into the root of the response JSON.

I.e. it returns like so:

```json
{
  "total_count": 1,
  "repositories": [
    {...}, {...}, {...}
  ]
}
```
while the usual way the API structures the responses is e.g. (listing commits on a repository):

```json
[
  {...}, {...}, {...}
]
``` 

As far as I looked - this library did not support parsing responses such as the former.

I added handling for parsing responses containing item lists in nested fields. And the API method in question itself.

@bitwiseman This is an initial draft, to see if this is the correct approach to take. I have it working in an internal application, but I am aware that there might be different requirements to make it public. Can you check if the approach here is alright?

If the approach taken is suitable - I would continue with cleaning it up, adding tests & doing the other tasks necessary for final reviews.

PR #966, opened 2020-10-20
Clarify how to configure local credentials.

chids:
⚠️ I'm not fully convinced this is what's intended to be the recommended way but it is _one_ way of doing the setup. Happy to expand on these instructions and include other setups if there are others that are recommended.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,13 +17,14 @@ This the default behavior.
 
 ### Setting up credential
 
-1. Create an OAuth token on github.com
-2. Set the GITHUB_OAUTH environment variable to the value of that token
-3. Set the system property `test.github.useProxy` (usually like "-Dtest.github.useProxy" as a Java VM option)
+1. Create [a personal access token](https://github.com/settings/tokens) on github.com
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -17,13 +17,14 @@ This the default behavior.
 
 ### Setting up credential
 
-1. Create an OAuth token on github.com
-2. Set the GITHUB_OAUTH environment variable to the value of that token
-3. Set the system property `test.github.useProxy` (usually like "-Dtest.github.useProxy" as a Java VM option)
+1. Create [a personal access token](https://github.com/settings/tokens) on github.com
+2. Set the `GITHUB_OAUTH` environment variable to the value of that token
+3. Set the `GITHUB_LOGIN` environment variable to your Github.com username
```
bitwiseman, line 22: This should not be needed.  Have you needed to do this?  
```suggestion
```

bitwiseman (changes requested):
Thanks for the update, but I don't think the new step is needed.  Could you discuss why the change?

bitwiseman:
@chids 
I'm going to close this for now. Happy to discuss further and reopen.

PR #980, opened 2020-11-19
Add ability to update a check run

mrginglymus:
Add functionality to update an existing check run

I have added a test which I fully expect to fail, as I have no idea where to even begin with implementing the wiremock backend - are they all manually crafted files?


timja:
> are they all manually crafted files?

they are recorded files mostly, possibly with small tweaks on top, I've invited you to the test org

mrginglymus:
Hmm, the tests for checks appear to use a repo owned by @jglick (along with a corresponding app).

timja:
> Hmm, the tests for checks appear to use a repo owned by @jglick (along with a corresponding app).

might be better to migrate them to the test org

mrginglymus:
Ok, I'll see what I can do 👍

mrginglymus:
Well that was churny :) All passing now @timja, thanks for sorting me out with the access.

timja (approved):
lgtm

src/test/java/org/kohsuke/github/GHCheckRunBuilderTest.java
```diff
@@ -26,17 +26,22 @@
...
+    protected GitHub getInstallationGithub() throws IOException {
+        return getAppInstallationWithTokenApp3().getRoot();
+    }
 
     @Test
     public void createCheckRun() throws Exception {
-        GHCheckRun checkRun = gitHub.getRepository("jglick/github-api-test")
-                .createCheckRun("foo", "4a929d464a2fae7ee899ce603250f7dab304bc4b")
+        GHCheckRun checkRun = getInstallationGithub().getRepository("hub4j-test-org/test-checks")
+                .createCheckRun("foo", "89a9ae301e35e667756034fdc933b1fc94f63fc1")
```
jglick, line 44: Is there a reason you needed to update existing tests? Once the mocks are created they should continue to work so long as you are not actually changing the behavior of the existing APIs, which AFAICT you are not.

jglick (commented):
WireMock in replay mode is basically unreviewable from my PoV; I find it more maintainable to write mocks programmatically. ([example](https://github.com/jenkinsci/github-branch-source-plugin/blob/72db8334dbb9203215a941c94ea1ef0b12cc5e84/src/test/java/org/jenkinsci/plugins/github_branch_source/GithubAppCredentialsTest.java#L78-L89))

src/test/java/org/kohsuke/github/GHCheckRunBuilderTest.java
timja: It was to make it easier to update them for others by moving the tests to the test org and an app that others can use I believe

mrginglymus:
Hmm, from looking at how this might be used downstream, it could probably do with a `GHRepository#updateCheckRun(long checkId)` - does that seem reasonable? Otherwise it's only accessible from a `GHCheckRun` object.

timja:
> GHRepository

given that create is there https://github.com/hub4j/github-api/blob/master/src/main/java/org/kohsuke/github/GHRepository.java#L1957 I don't see why not

timja (approved):

timja:
@bitwiseman did you want to take a look at this?

bitwiseman (approved):
I'll look at this tody

PR #982, opened 2020-11-20
Feature/add update hook

mronus:
# Description 

* Add support for updating hooks
* Fix the type of parameter in `getRepositoryById`
* Replace `position` with `line` in `createReviewComment` in order to avoid position computation

### DIFF
--- a/src/main/java/org/kohsuke/github/GHHooks.java
+++ b/src/main/java/org/kohsuke/github/GHHooks.java
@@ -77,13 +77,51 @@ class GHHooks {
             }
 
             GHHook hook = root.createRequest()
-                    .method("POST")
+                    .method("PATCH")
                     .with("name", name)
                     .with("active", active)
                     .with("config", config)
                     .with("events", ea)
                     .withUrlPath(collection())
                     .fetch(clazz());
+
+            return wrap(hook);
+        }
+
+        /**
+         * Update hook gh hook.
+         *
+         * @param id
+         *            the ID
+         * @param name
+         *            the name
+         * @param config
+         *            the config
+         * @param events
+         *            the events
+         * @param active
+         *            the active
+         * @return the gh hook
+         * @throws IOException
+         *             the io exception
+         */
+        public GHHook updateHook(long id, String name, Map<String, String> config, Collection<GHEvent> events, boolean active)
+            throws IOException {
+            List<String> ea = null;
+            if (events != null) {
+                ea = new ArrayList<String>();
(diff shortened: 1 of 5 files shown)


src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -558,7 +558,7 @@ public GHRepository getRepository(String name) throws IOException {
      * @throws IOException
      *             the io exception
      */
-    public GHRepository getRepositoryById(String id) throws IOException {
+    public GHRepository getRepositoryById(long id) throws IOException {
```
bitwiseman, line 561: Please make this a separate PR and keep the existing method but make it deprecated. 
Also, add tests for both the old and new methods.  Thanks!

src/main/java/org/kohsuke/github/GHPullRequest.java
```diff
@@ -510,7 +510,7 @@ public GHPullRequestReviewComment createReviewComment(String body, String sha, S
                 .with("body", body)
                 .with("commit_id", sha)
                 .with("path", path)
-                .with("position", position)
+                .with("line", position)
```
bitwiseman: This too needs to be a separate PR, with tests for the change.

The docs indicate that `position` and `line` are not equivalent. You'll probably need to implement a builder to support the various parameter combinations.
https://docs.github.com/en/free-pro-team@latest/rest/reference/pulls#create-a-review-comment-for-a-pull-request

src/main/java/org/kohsuke/github/GHHooks.java
```diff
@@ -77,7 +77,7 @@ public GHHook createHook(String name, Map<String, String> config, Collection<GHE
             }
 
             GHHook hook = root.createRequest()
-                    .method("POST")
+                    .method("PATCH")
```
bitwiseman: Does `POST` not work?

https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#create-a-repository-webhook

bitwiseman (changes requested):
Thank you for choosing to contribute!  
Please split these three independent changes into separate PRs and add tests.  Thanks!

bitwiseman:
@mronus  
Do you have time to finish working on this?

mronus:
@bitwiseman I don't know when I can finish it but will try to do it asap, like in a few weeks at the latest.

mronus:
This PR got closed as it was split into 3 different PRs.

PR #960, opened 2020-10-02
 Implement deployment API support for ant-man and flash previews

marcoferrer:
# Description 
* Implement deployment API support for ant-man and flash previews.
* This pr includes two new constants in the `Previews` class. 
* Getters for newly introduced fields are added to `GHDeployment` and `GHDeploymentStatus`.
* New builder setters are implemented for `GHDeploymentBuilder` and `GHDeploymentStatusBuilder`.
* `GHDeploymentState` has been updated to support `in_progress`, `queued`, and `inactive`. 
  * [Related Gh API Docs](https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#create-a-deployment-status)
* Marked `GHDeploymentStatus.getTargetUrl()` and `GHDeploymentStatusBuilder.targetUrl()` as `@Deprecated`.
  * Method documentation has been updated to advise users to replace usages with `logUrl()` where appropriate

### Github Blog Posts For Previews Implemented
* [flash](https://developer.github.com/changes/2018-10-16-deployments-environments-states-and-auto-inactive-updates/) _(10/16/2018)_
* [ant-man](https://developer.github.com/changes/2016-04-06-deployment-and-deployment-status-enhancements/) _(04/06/2016)_
# Before submitting a PR:
We love getting PRs, but we hate asking people for the same basic changes every time. 

- [X] Push your changes to a branch other than `master`. Create your PR from that branch.    
- [X] Add JavaDocs and other comments
- [x] Write tests that run and pass in CI. See [CONTRIBUTING.md](CONTRIBUTING.md) for details on how to capture snapshot data.
- [x] Run `mvn clean compile` locally. This may reformat your code, commit those changes.
- [x] Run `mvn -D enable-ci clean install site` locally. If this command doesn't succeed, your change will not pass CI.

# When creating a PR: 

- [X] Fill in the "Description" above. 
- [X] Enable "Allow edits from maintainers".

PR #983, opened 2020-11-25
Add arch test for preview API usage

marcoferrer:
# Description 
This PR introduces the concept of architecture tests to the project. The first test implemented verifies that any method or class that is annotated `@Preview` is also annotated with `@Deprecated`.

The library used to implement the tests is [ArchUnit](https://github.com/TNG/ArchUnit)

The following is an example of the output created when this test fails.
image

# Before submitting a PR:

# When creating a PR: 


marcoferrer:
@bitwiseman Based on the discussion we had in this PR https://github.com/hub4j/github-api/pull/960

I decided to implement some architecture tests to cover the `@Deprecated` requirement via CI.

Thoughts?

src/test/java/org/kohsuke/github/ArchTests.java
```diff
@@ -0,0 +1,56 @@
...
+import com.tngtech.archunit.lang.ArchRule;
+import com.tngtech.archunit.lang.ConditionEvents;
+import com.tngtech.archunit.lang.SimpleConditionEvent;
+import org.junit.Test;
+
+import static com.tngtech.archunit.lang.syntax.ArchRuleDefinition.classes;
+
+public class ArchTests {
+
+    private final JavaClasses classFiles = new ClassFileImporter()
```
sullis, line 19: "classFiles" could be a static variable.

Then I would add this:

``
    @BeforeClass
    public static void beforeClass() {
        assertTrue(classFiles.size() > 0);
    }
```

sullis (changes requested):

src/test/java/org/kohsuke/github/ArchTests.java
bitwiseman: @sullis 
Which way do you think it better?

bitwiseman (approved):
I like this. Seems reasonable.  Please address the one comment so far and then I'll merge.

marcoferrer:
@bitwiseman addressed feedback, simplified implementation, added another rule to check for preview APIs in enum fields

PR #984, opened 2020-11-27
Close Okhttp ResponseBody instances when closing the InputStream

gsmet:
Hi there,

I'm opening this PR to have some preliminary discussions about this change. I'm using Okhttp as a backend as the default implementation does not work very well with GraalVM's native images whereas Okhttp works flawlessly.

I put what I'm working on in semi-production and started to see warnings from Okhttp complaining about unclosed ResponseBody instances.

After a quick check of the code, it seems that ResponseBody instances are not closed and only the InputStream instances are.

This PR solves the issue by making sure we close both the ResponseBody and the InputStream when closing the InputStream in the GitHub API. The Javadoc of ResponseBody says closing the stream is enough. But AFAICS, you properly close the stream in the GitHub API and I had quite a lot of warnings.

I just put this on try in production to see how it goes.

/cc @yrodiere


bitwiseman (approved):
@gsmet  This looks reasonable to me.  What has been the result in your semi-production testing?

gsmet:
AFAICS, I don't have the warnings anymore. I will let it run this week and report back.

bitwiseman:
Here's the docs on this subject: https://square.github.io/okhttp/4.x/okhttp/okhttp3/-response-body/#the-response-body-must-be-closed

gsmet:
@bitwiseman I didn't see any warnings related to that this week.

gsmet:
So IMHO, it looks good to merge.

PR #985, opened 2020-11-28
Fixes null commit date

### LINKED ISSUE #883: GHCommit::ShortInfo::GHAuthor.getDate() returns null
**Describe the bug**
After getting commits from github.getRepository(repo).listCommits()
commit.getCommitShortInfo().getAuthor().getDate() returns null

**To Reproduce**
Use the following partial example code:

            Iterator commitIterator = github.getRepository(repo).listCommits().iterator();
            while (commitIterator.hasNext()) {
                GHCommit commit = (GHCommit) commitIterator.next();
                logger.info("Checking commit: {}", commit.getSHA1());
                logger.info("Checking commit date: {}", commit.getCommitShortInfo().getAuthor().getDate());
                }
            }

The date always print out as null

Checking commit: 00e40ec77ffe8d6ecc328719add6577cc2496f85
Checking commit date: null

**Expected behavior**
Commit date should be a actual date object like

Checking commit: 00e40ec77ffe8d6ecc328719add6577cc2496f85
Checking commit date: Tue Jun 30 17:20:50 GMT 2020

 - OS:
CentOS Linux release 7.7.1908 (Core)
KERNEL 3.10.0-1062.12.1.el7.x86_64

 - Version: github-api v1.113

**Additional context**
This trace back to the code comments in https://github.com/hub4j/github-api/blob/3b0216b0234eb093049fdb9afdd4da1aadc4ad8f/src/main/java/org/kohsuke/github/GitUser.java#L41

I believe there is a bug in this code: 
https://github.com/hub4j/github-api/blob/3b0216b0234eb093049fdb9afdd4da1aadc4ad8f/src/main/java/org/kohsuke/github/GHCommit.java#L122

A private member with the same name is already defined in base class
https://github.com/hub4j/github-api/blob/3b0216b0234eb093049fdb9afdd4da1aadc4ad8f/src/main/java/org/kohsuke/github/GitUser.java#L18

When calling the get function from the base class
https://github.com/hub4j/github-api/blob/3b0216b0234eb093049fdb9afdd4da1aadc4ad8f/src/main/java/org/kohsuke/github/GitUser.java#L44

It gets GitUser.date instead of GHAuthor.date and GitUser.date is null

This is a working example:
https://github.com/hub4j/github-api/blob/a42305dd59f0be426f0c9091748e947f60d76bcd/src/main/java/org/kohsuke/github/GHPullRequestCommitDetail.java#L53
Authorship.getDate() works. I believe GHAuthor should not define its own date

lower-case:
# Description 
Fixes #883 

Method `GHAuthor.getDate()` was giving null because the `date` field was overridden in deprecated GHAuthor class but the field was populated only in `GitUser` class.

# Before submitting a PR:

# When creating a PR: 


src/test/resources/org/kohsuke/github/CommitTest/wiremock/commitDateNotNull/mappings/repos_hub4j_github-api_commits_86a2e245aa6d71d54923655066049d9e21a15f01-3.json
```diff
@@ -0,0 +1,41 @@
+{
```
bitwiseman, line 1: You might need to re-record.  I don't see any use of this in the test.

src/test/java/org/kohsuke/github/CommitTest.java
```diff
@@ -120,4 +120,13 @@ public void commitSignatureVerification() throws Exception {
                     commit.getCommitShortInfo().getVerification().getPayload());
         }
     }
+
+    @Test // issue 883
+    public void commitDateNotNull() throws Exception {
+        GHRepository repo = gitHub.getRepository("hub4j/github-api");
+        GHCommit commit = repo.getCommit("ed4f9c8176866977677c99ac9668a8ce10231bc8");
+
+        assertNotNull(commit.getCommitShortInfo().getAuthoredDate());
```
bitwiseman, line 129: Please check for more than just not `null`.  There is a known expected value that isn't going to change over time.

bitwiseman (changes requested):
Thanks for contributing!

Minor change to test data.

Also, there are [other tests that return commit info](https://github.com/hub4j/github-api/search?q=getCommitShortInfo). Please add some checks to other tests to verify this fix works on exiting tests and data.

bitwiseman:
@lower-case 
Do you plan to continue work on this?

PR #1001, opened 2020-12-15
Implement static typing for previews and clean up usage declarations

marcoferrer:
# Description 
This PR introduces the following changes:
- Refactor the `Previews` const class into an enum to prevent string constants from being used at `@Preview` use sites.
- Updates existing usages of `@Preview` to include references to `Previews` enum constants.

# Things to consider
- The PR would introduce a breaking change to consumers that rely directly on the older strings constants.
  - They would be able to remediate this by updating their usage from `Previews.ANTMAN` -> `Previews.ANTMAN.mediaType()` 
- One of the main goals of this PR is to improve transparency to consumers on not just which APIs are previews but what previews exactly they depend on. Allowing consumers to make more informed decisions on their usage.
- We are still not requiring a media type to always be included at use sites be it seems like `@Preview` is also being used to denote sdk specific beta APIs in some cases. I suggest, when appropriate we migrate those usages to use a new annotation as to not conflate the two 'Preview' use cases. Something along the lines of `@Hub4jBetaApi` or `@SdkBetaApi`. After that we should have no issue making preview media types a hard requirement.

# Before submitting a PR:

- [X] Run `mvn clean compile` locally. This may reformat your code, commit those changes.

# When creating a PR: 

### DIFF
--- a/src/main/java/org/kohsuke/github/GHRepository.java
+++ b/src/main/java/org/kohsuke/github/GHRepository.java
@@ -708,7 +708,7 @@ public class GHRepository extends GHObject {
      * @return the boolean
      */
     @Deprecated
-    @Preview
+    @Preview(BAPTISTE)
     public boolean isTemplate() {
         // isTemplate is still in preview, we do not want to retrieve it unless needed.
         if (isTemplate == null) {
@@ -1925,7 +1925,7 @@ public class GHRepository extends GHObject {
      * @see <a href="https://developer.github.com/v3/checks/runs/#list-check-runs-for-a-specific-ref">List check runs
      *      for a specific ref</a>
      */
-    @Preview
+    @Preview(ANTIOPE)
     @Deprecated
     public PagedIterable<GHCheckRun> getCheckRuns(String ref) throws IOException {
         GitHubRequest request = root.createRequest()
@@ -1999,7 +1999,7 @@ public class GHRepository extends GHObject {
      *            the commit hash
      * @return a builder which you should customize, then call {@link GHCheckRunBuilder#create}
      */
-    @Preview
+    @Preview(ANTIOPE)
     @Deprecated
     public @NonNull GHCheckRunBuilder createCheckRun(@NonNull String name, @NonNull String headSHA) {
         return new GHCheckRunBuilder(this, name, headSHA);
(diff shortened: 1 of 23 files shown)


src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -2959,10 +2959,14 @@ void populate() throws IOException {
...
-            root.createRequest().withPreview(BAPTISE).setRawUrlPath(url.toString()).fetchInto(this).wrap(root);
+            root.createRequest().withPreview(BAPTISTE).setRawUrlPath(url.toString()).fetchInto(this).wrap(root);
         } catch (HttpException e) {
             if (e.getCause() instanceof JsonParseException) {
-                root.createRequest().withPreview(BAPTISE).withUrlPath("/repos/" + full_name).fetchInto(this).wrap(root);
+                root.createRequest()
+                        .withPreview(BAPTISTE)
+                        .withUrlPath("/repos/" + full_name)
+                        .fetchInto(this)
+                        .wrap(root);
```
marcoferrer, line 2969: This was a product of the formater

src/main/java/org/kohsuke/github/GitHubRequest.java
```diff
@@ -437,6 +437,10 @@ public B withPreview(String name) {
             return withHeader("Accept", name);
         }
 
+        public B withPreview(Previews preview) {
+            return withPreview(preview.mediaType());
+        }
+
```
marcoferrer, line 443: This overload of the `withPreview` method was defined to prevent us from having to update every usage of the method after `Previews` was converted to an enum

src/main/java/org/kohsuke/github/Previews.java
```diff
@@ -7,121 +7,133 @@
...
+    ANT_MAN("application/vnd.github.ant-man-preview+json"),
 
     /**
      * Create repository from template repository
      *
      * @see <a href="https://developer.github.com/v3/previews/#create-and-use-repository-templates">GitHub API
      *      Previews</a>
      */
-    static final String BAPTISE = "application/vnd.github.baptiste-preview+json";
+    BAPTISTE("application/vnd.github.baptiste-preview+json"),
```
marcoferrer, line 32: Typo fixed throughout the code base

bitwiseman (approved):
Excellent.

PR #1012, opened 2021-01-03
Fix the type of the id parameter of Github#getRepositoryById

mronus:


src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -557,11 +557,28 @@ public GHRepository getRepository(String name) throws IOException {
      * @return the repository by id
      * @throws IOException
      *             the io exception
+     *
+     * @deprecated Do not use this method. It was added due to misunderstanding of the type of parameter. Use the
+     *             alternative getRepositoryById method with long type parameter instead
```
bitwiseman, line 562:
```suggestion
     * @deprecated Do not use this method. It was added due to misunderstanding of the type of parameter. Use the
     *             {@link #getRepositoryById(long)} instead
```

bitwiseman (changes requested):
Thanks for contributing!

Add tests for the new method (and the deprecated one if they are already present).

See CONTRIBUTING.md for instructions on setup and see existing tests for examples.

bitwiseman:
@mronus Do you plan to finish this?

bitwiseman (changes requested):
Great!  Now some tests!

PR #1015, opened 2021-01-05
Fix mocking Previews

seregamorph:
# Description
Make `org.kohsuke.github.Previews` enum public.

Reasoning: fail to Mockito mock classes like `GHPullRequest` with failure:
```
org.mockito.exceptions.base.MockitoException: 
Mockito cannot mock this class: class org.kohsuke.github.GHPullRequest.
Mockito can only mock non-private & non-final classes.
...
Caused by: java.lang.IllegalAccessError: tried to access class org.kohsuke.github.Previews from class com.sun.proxy.$Proxy16
```
It's a regression (test-scoped) since `1.118`, was fine in `1.117`.

Classes like `GHPullRequest` are "read-only" by design, hence we need convenient way to instantiate it in tests. Obvious possible way is to parse json payload, but it may be too verbose for trivial unit-testing.

### DIFF
--- a/src/test/java/org/kohsuke/github/GHPullRequestMockTest.java
+++ b/src/test/java/org/kohsuke/github/GHPullRequestMockTest.java
@@ -1,0 +1,21 @@
+package org.kohsuke.github;
+
+import org.junit.Test;
+
+import java.io.IOException;
+
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class GHPullRequestMockTest {
+
+    @Test
+    public void shouldMockGHPullRequest() throws IOException {
+        GHPullRequest pullRequest = mock(GHPullRequest.class);
+        when(pullRequest.isDraft()).thenReturn(true);
+
+        assertTrue("Mock should return true", pullRequest.isDraft());
+    }
+
+}
--- a/pom.xml
+++ b/pom.xml
@@ -153,7 +153,7 @@
                       <exclude>org.kohsuke.github.example.*</exclude>
 
                       <!-- No methods -->
-                      <exclude>org.kohsuke.github.Previews</exclude>
+                      <exclude>org.kohsuke.github.internal.Previews</exclude>
 
                       <!-- Deprecated -->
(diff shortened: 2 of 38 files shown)


src/test/java/org/kohsuke/github/GHPullRequestMockTest.java
```diff
@@ -0,0 +1,21 @@
...
+
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class GHPullRequestMockTest {
+
+    @Test
+    public void shouldMockGHPullRequest() throws IOException {
+        GHPullRequest pullRequest = mock(GHPullRequest.class);
```
seregamorph, line 15: this test will fail if you revert change in `Previews` class (public modifier)

bitwiseman:
Hm, I thought I'd seen Mockito work on internal classes, but ... oh, well. 

@seregamorph 
Do you think it would make sense to move this to an a new sub-package? `org.kohsuke.github.internal` or `org.kohsuke.github.support` or something?

I suppose it is just one class, but there are other classes that might need to be exposed this same way for mocking or even just convenience. For example, it probably makes sense to move the `Preview` annotation in there as well. There is no reason for any client of this library to use or need to know about that annotation.

bitwiseman (changes requested):
Thanks for the PR!

I understand this blocks adoption of 1.118 for you due to test breakage. 

I'd like to resolve the question I asked about package placement before merging, but once that is done, I'm happy to merge and release this quickly.

seregamorph:
@bitwiseman moved to `internal` sub-package.

bitwiseman (approved):

bitwiseman:
@seregamorph 
Released as v1.119.

seregamorph:
Thanks

youming-lin:
Thanks for the fix! I hit the same issue and eventually realized that the enum not being visible is the cause.

Prior to this change it seemed odd that a function that returns instance of `Previews[]` type is public but `Previews` itself was not. If this type is not meant to be exposed outside the package, then perhaps the function should've also been default instead of public visibility.

PR #1033, opened 2021-02-04
Make sure that `output.text` is set in each checks call

uhafner:
# Description 

If a GitHub checks contains more than 50 annotations, then the
check is split into several calls. This fix makes sure that all
additional calls will not only copy the properties `title` and
`summary` but also the previously missing property `text`.

# Before submitting a PR:

# When creating a PR: 


timja (approved):

bitwiseman (approved):

uhafner:
@bitwiseman It would be nice it you could cut a new release with that fix.

PR #1054, opened 2021-03-09
Fix concurrency issues with GHIssue addLabels and removeLabels

### LINKED ISSUE #1049: Concurrency issues when adding labels to an issue/PR
Hi,

Some context: we are using the Hub4j GitHub API for our Quarkus Bot (it's a GitHub App based on Quarkus and running as a GraalVM native executable).

The bot can automatically add labels based on triage rules but we started seeing concurrency issues when someone adds labels when the bot is also adding some. Believe it or not, this happened in real life.

AFAICS, the issue is that the `GHIssue#addLabels()` method doesn't really add the labels but set them. My understanding is that it should use this method: https://docs.github.com/en/rest/reference/issues#add-labels-to-an-issue and that's not what is done here: https://github.com/hub4j/github-api/blob/6d86cfb4f60642ccfc3ad943eeb472e33e2ff9ec/src/main/java/org/kohsuke/github/GHIssue.java#L362-L374 : you set the labels based on the existing ones + the added ones instead of adding them, which can result in data loss if you're not lucky.

The issue occurs when a label is added manually in the UI between the call at line 365 and the call at line 373: the labels added in the UI are lost, which is very unfortunate given they are critical to our workflow. You can see an example of this issue here: https://github.com/quarkusio/quarkus/pull/15402#event- - Georgios added the `triage/backport?` label in the UI and the bot removed it while our bot is only **adding** labels.

Does it make any sense? Any chance this could be fixed?

Thanks!

gsmet:
# Description 

Fixes concurrency issues when adding and removing labels by using the proper GitHub APIs.

Fixes #1049

# Before submitting a PR:

# When creating a PR: 


src/main/java/org/kohsuke/github/GHIssue.java
```diff
@@ -411,15 +405,9 @@ public void removeLabels(Collection<GHLabel> labels) throws IOException {
...
 
     private void _removeLabels(Collection<String> names) throws IOException {
-        List<String> newLabels = new ArrayList<String>();
-
-        for (GHLabel l : getLabels()) {
-            if (!names.contains(l.getName())) {
-                newLabels.add(l.getName());
-            }
+        for (String name : names) {
+            root.createRequest().method("DELETE").withUrlPath(getIssuesApiRoute() + "/labels", name).send();
```
bitwiseman, line 409: Ouch. I didn't notice this behavior. Not great for rate limit budget... But it is correct.  Good job.

src/test/java/org/kohsuke/github/GHPullRequestTest.java
```diff
@@ -421,6 +421,64 @@ public void setLabels() throws Exception {
...
+        String addedLabel2 = "addLabelsConcurrencyIssue_label_name_2";
+
+        GHPullRequest p1 = getRepository()
+                .createPullRequest("addLabelsConcurrencyIssue", "test/stable", "master", "## test");
+        p1.getLabels();
+
+        GHPullRequest p2 = getRepository().getPullRequest(p1.getNumber());
+        p2.addLabels(addedLabel2);
+
+        p1.addLabels(addedLabel1);
```
bitwiseman, line 454: Also do a case where you add two labels in one method call - which should produce one API call, not two.

bitwiseman (changes requested):
One test change requested. Otherwise, look good to me.

src/main/java/org/kohsuke/github/GHIssue.java
```diff
@@ -360,17 +360,11 @@ public void addLabels(Collection<GHLabel> labels) throws IOException {
...
-            newLabels.add(label.getName());
-        }
-        for (String name : names) {
-            if (!newLabels.contains(name)) {
-                newLabels.add(name);
-            }
-        }
-        setLabels(newLabels.toArray(new String[0]));
+        root.createRequest()
+                .with("labels", names.toArray(new String[0]))
```
bitwiseman, line 364: Wait... this should add a list of labels in one call, not only work for one label. 
https://docs.github.com/en/rest/reference/issues#add-labels-to-an-issue

src/main/java/org/kohsuke/github/GHIssue.java
gsmet: Yeah, I agree it's unfortunate they did it that way. It should have been symmetrical with adding labels IMHO.

src/main/java/org/kohsuke/github/GHIssue.java
gsmet: I think that's what it did anyway but I see you can pass directly the collection so that's what is done now.

gsmet:
@bitwiseman I think I addressed all your concerns and I checked adding two labels only did one API call.

src/test/java/org/kohsuke/github/GHPullRequestTest.java
```diff
@@ -421,6 +421,66 @@ public void setLabels() throws Exception {
...
+    @Test
+    // Requires push access to the test repo to pass
+    public void addLabels() throws Exception {
+        GHPullRequest p = getRepository().createPullRequest("addLabels", "test/stable", "master", "## test");
+        String addedLabel1 = "addLabels_label_name_1";
+        String addedLabel2 = "addLabels_label_name_2";
+        String addedLabel3 = "addLabels_label_name_3";
+
+        p.addLabels(addedLabel1);
+        p.addLabels(addedLabel2, addedLabel3);
```
bitwiseman, line 433:
```suggestion
        int requestCount = mockGitHub.getRequestCount();
        p.addLabels(addedLabel2, addedLabel3);

        // multiple labels can be added with one api call
        assertThat(mockGitHub.getRequestCount(), equalTo(requestCount + 1));
```

src/test/java/org/kohsuke/github/GHPullRequestTest.java
```diff
@@ -421,6 +421,66 @@ public void setLabels() throws Exception {
...
+
+        Collection<GHLabel> labels = getRepository().getPullRequest(p.getNumber()).getLabels();
+        assertEquals(3, labels.size());
+        assertThat(labels,
+                containsInAnyOrder(hasProperty("name", equalTo(addedLabel1)),
+                        hasProperty("name", equalTo(addedLabel2)),
+                        hasProperty("name", equalTo(addedLabel3))));
+    }
```
bitwiseman, line 441:
```suggestion
        // Adding a label which is already present does not throw an error
        p.addLabels(addedLabel1);
    }
```

bitwiseman: Need to rerecord for this change, sorry.

src/test/java/org/kohsuke/github/GHPullRequestTest.java
```diff
@@ -421,6 +421,66 @@ public void setLabels() throws Exception {
...
+
+        Collection<GHLabel> labels = getRepository().getPullRequest(p.getNumber()).getLabels();
+        assertEquals(3, labels.size());
+
+        p.removeLabels(label2, label3);
+
+        labels = getRepository().getPullRequest(p.getNumber()).getLabels();
+        assertEquals(1, labels.size());
+        assertEquals(label1, labels.iterator().next().getName());
+    }
```
bitwiseman, line 482: Hm, if you the following above this line, you'll get a `GHFileNotFoundException`. 

```
        p.removeLabels(label2);
```

Need to test this scenario.  Also, it seems like this could still result in concurrency issues.  Two clients go to remove the same label and one gets an error while the other doesn't.  Would you rather merge this change with this or work on a solution?

bitwiseman (changes requested):
Sorry, a few more changes and questionsI didn't think of on the previous round.

src/test/java/org/kohsuke/github/GHPullRequestTest.java
gsmet: if you... what? There's a word missing :)

I don't see how the case you're talking about is a concurrency issue? For sure, if you have several clients removing the same labels at the same time, you will have errors. But that's how the APIs are designed and that's expected in my opinion.

What's not acceptable is the issue that this PR fixes e.g. your use of the API is perfectly fine and you lose data.

src/test/java/org/kohsuke/github/GHPullRequestTest.java
gsmet: Frankly, I fail to see how this is related to this PR at all.

gsmet: Or is it because the way it was broken before, this case would work?

src/test/java/org/kohsuke/github/GHPullRequestTest.java
bitwiseman: We probably want to have `removeLabels(...)` match the old method behavior and not throw if any of the label is not present. 
And then add a `removeLabel(String)` that does throw if the specified label isn't present.

bitwiseman (approved):
@gsmet 
I updated `removeLabels()` to match the existing non-throwing behavior and added a `removeLabel()` that will behave like the GitHub API and throw.   

This retains your fixes but also stays consistent with past behavior where possible. 

What do you think?

gsmet:
@bitwiseman sorry, I was extremely busy this week. I'll try to have a look next week!

gsmet:
@bitwiseman I find it a bit weird that `removeLabel` and `removeLabels` don't behave the same way but I can live with it and we can adjust if people complain :).

So I think we should merge it!

PR #1063, opened 2021-03-19
Fix CI issues

gsmet:
# Description 

This PR fixes the CI issues with JDK 16+.

The detail is in each commit. There's a good chance things will need proper fixes but at least we can have a green CI until workarounds are found.

For the record, I pass the options for JDK 11+ as they are needed but not required for JDK 11.

For now, I haven't added 17-ea as Findbugs' ASM version doesn't like it yet.

### DIFF
--- a/.github/workflows/maven-build.yml
+++ b/.github/workflows/maven-build.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


gsmet:
@bitwiseman looks like you will need to adjust the Expected statuses if we want to get this in :).

bitwiseman (approved):

PR #1064, opened 2021-03-22
Implement GHWorkflow, GHWorkflowRun and associated payloads

gsmet:
This is a draft for now.

I have several questions and there is at least one big issue with the tests.


src/main/java/org/kohsuke/github/GHWorkflowRun.java
```diff
@@ -0,0 +1,352 @@
...
+    public GHEvent getEvent() {
+        return Enum.valueOf(GHEvent.class, event.toUpperCase(Locale.ROOT));
+    }
+
+    /**
+     * Gets status of the workflow run. It can be one of queued, in_progress, or completed.
+     *
+     * @return status of the workflow run
+     */
+    // TODO GHWorkflowRun should we rather return a String here? That's what GHCheckSuite does.
```
gsmet, line 195: Here, I'm not consistent with GHCheckSuite but I'm not entirely sure it's a good idea to return a String here.

My natural inclination would be for consistency but having an enum is better from a usability POV. Or did you do that to be sure it never fails even if GitHub adds new values?

WDYT?

src/test/java/org/kohsuke/github/GHWorkflowRunTest.java
```diff
@@ -0,0 +1,173 @@
...
+            pollInterval = Duration.ofMillis(1);
+            atMost = Duration.ofMillis(12);
+        }
+    }
+
+    @Test
+    public void testManualRunAndBasicInformation() throws IOException {
+        GHWorkflow workflow = repo.getWorkflow("fast-workflow.yml");
+
+        Date referenceDate = templating.testStartDate;
```
gsmet, line 56: Here, I need to have some recording of the date of the test. I thought that maybe this could be useful but apparently not as it's not recorded AFAICS.

Do you have any idea how I could record the start date to be sure I can test things properly?

gsmet:
@bitwiseman interested with your wisdom on this one. I added some comments for the big remaining issues from my point of view.

pom.xml
```diff
@@ -422,6 +409,12 @@
       <version>4.13.2</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.awaitility</groupId>
+      <artifactId>awaitility</artifactId>
+      <version>4.0.3</version>
+      <scope>test</scope>
+    </dependency>
```
bitwiseman, line 417: Cool.

src/test/java/org/kohsuke/github/GHWorkflowRunTest.java
gsmet: I went for a run and things are clearer now. I think I will just base things on the last id, that will work and be less sensitive to time inconsistencies.

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -2898,6 +2898,74 @@ public GHIssueEvent getIssueEvent(long id) throws IOException {
...
+     * @throws IOException
+     *             the io exception
+     */
+    public GHWorkflowRun getWorkflowRun(long id) throws IOException {
+        return root.createRequest()
+                .withUrlPath(getApiTailUrl("actions/runs"), String.valueOf(id))
+                .fetch(GHWorkflowRun.class)
+                .wrapUp(this);
+    }
+
```
bitwiseman, line 2968: What would you think of having a `workflows()` method that returns a `WorkflowsEndpoint` instance linked to this repository and move all these methods into that class? The idea here is to reduce the size/complexity of `GHRepository` and to more closely match the GraphQL object model. 

I'm not saying you have to do it for this change, I'm more taking notes for v2 design ideas and I'd like your opinion.

gsmet: I think for now it makes sense to have them here given the others are here. But I agree it would be nice to group them into subapis.

gsmet: I could do it now if you would like to do it for new API endpoints, it's not biggie.

src/main/java/org/kohsuke/github/GHWorkflowRun.java
bitwiseman: What you've done here is reasonable.  Store as `String` but return an Enum.  This will result in a runtime exception in the string doesn't match the enum, but the object can still be loaded and interacted with in other ways.  

I'm pretty sure `GHCheckSuite` is the "never fails even if GitHub adds a new value".   We see this problem most strikingly in `GHEvent` where they keep adding new types and it causes hard failures.  

Consistency might also be achieved by changing `GHCheckSuite` to match this. 

In an ideal world, we'd use a "graceful enum" that stores the literal string value and returns the matching enum value or a configurable fallback value when the string doesn't match any of the enum values.  That would solve the issue we see with `GHEvent` and also what we see here.  So, for example, you'd get a `Status.UNKNOWN` if they ever add a new value that doesn't parse, and there'd need to be a way to get the raw string value at that point.

gsmet: But can we break compatibility for `GHCheckSuite`?

If not, it's probably better to have both consistent for now and open an issue for a 2.0 to change them to enum (with a fallback and a `getRaw...` to get the string value).

src/test/java/org/kohsuke/github/GHWorkflowRunTest.java
gsmet: I was able to fix it and get the tests to work (hopefully). They are stable on my laptop, let's see on CI.

src/main/java/org/kohsuke/github/GHRepository.java
bitwiseman: Meh, let's not innovate right now.

pom.xml
gsmet: Yeah, this library is very practical. We use it a lot in Quarkus. It's a bit harder to get things working properly with wiremock recording though.

gsmet:
@bitwiseman ok, the situation is a bit better: I was able to get the tests to run in both modes.

So that lets us with the 2 design decisions:
- enum or not enum for the `getStatus/Completion()`
- isolating the API in `workflows()`

Let me know your take on these two and I'll adjust.

In the meantime, I will work on the payloads.

src/main/java/org/kohsuke/github/GHRepository.java
gsmet: OK :).

src/main/java/org/kohsuke/github/GHWorkflowRun.java
bitwiseman: > But can we break compatibility for `GHCheckSuite`?

We could add a `@BridgeMethod` to keep binary compatibility.  Are they the same status and conclusion enumerations?

bitwiseman:
* isolating the API in workflows()

No, it's fine, leave it. 

* enum or not enum for the getStatus/Completion()

I think you could add a bridge method annotation to bring consistency. And we don't need to  create class - we can do the same as [EnumUtils.getEnum(Class, String, E)](https://commons.apache.org/proper/commons-lang/apidocs/src-html/org/apache/commons/lang3/EnumUtils.html#line.215).  We can't update the version of commons-lang due to downstream limitations, but we can use the same behavior.  

I would even be fine with not having a `getRaw*` method.  We simply don't support other values.  It would make it very clear that any code depending on this method needs to handle the potential case of unknown values.

gsmet:
@bitwiseman I implemented the payloads in my last commit. From my side, this PR is ready to go. I'll let you review it.

There's still something we need to fix with pull requests for check suites and workflow runs but I'll do it in a separate PR once this is in as it's a preexisting issue and a totally different discussion.

gsmet:
Also, is there a way to test bridge methods?

src/main/java/org/kohsuke/github/GHEventPayload.java
```diff
@@ -1326,4 +1327,86 @@ void wrapUp(GitHub root) {
...
+        private GHWorkflowRun workflowRun;
+        private GHWorkflow workflow;
+
+        public GHWorkflowRun getWorkflowRun() {
+            return workflowRun;
+        }
+
+        public GHWorkflow getWorkflow() {
+            return workflow;
+        }
```
jeffmaury, line 1393: There is also (seems undocumented) a head_sha field that is part of the payload. See https://github.com/redhat-developer/intellij-openshift-connector/blob/master/.github/workflows/publish_reports.yml

This is required for my first Quarkus based GitHub app 🥰

gsmet: It's already there, just nested in the `GHWorkflowRun` object: https://github.com/hub4j/github-api/pull/1064/files#diff-87252b4de52f4e0e1317a91eee8a0de1cc885f7b84f872989c8e52ec09d2509dR41 :)

@jeffmaury did you use https://github.com/quarkiverse/quarkus-github-app ?

jeffmaury: That's the plan

gsmet: Cool, let me know how it goes :). (and make sure to use the included Replay UI)

bitwiseman:
> Also, is there a way to test bridge methods?

Not really.  There is a test to verify they are created, but executing them no.

src/main/java/org/kohsuke/github/internal/EnumUtils.java
```diff
@@ -0,0 +1,20 @@
+package org.kohsuke.github.internal;
+
+import java.util.Locale;
+
+public final class EnumUtils {
+
+    public static <E extends Enum<E>> E getEnum(Class<E> enumClass, String value, E defaultEnum) {
```
bitwiseman, line 7: JavaDoc needed.

src/main/java/org/kohsuke/github/internal/EnumUtils.java
```diff
@@ -0,0 +1,20 @@
...
+        if (value == null) {
+            return null;
+        }
```
bitwiseman, line 10: Why return `null`?  Enumerations are value types, I would expect this method to always return a value, never null. 

A comment on you reasoning here would be useful.

bitwiseman: Maybe change the method name to `getEnumOrNull`?  ... I'd kind of expect the name to be `getEnumOrDefault` already, so then it would become `getNullableEnumOrDefault`?   It's "internal" so we can change it later if needed, I guess.

gsmet: null is a valid value for some of the fields that are enums. When the API doesn't return a value, we want it to be null.

I adjusted the method name per your suggestion.

src/main/java/org/kohsuke/github/internal/EnumUtils.java
gsmet: Added.

src/test/java/org/kohsuke/github/GHWorkflowRunTest.java
```diff
@@ -0,0 +1,223 @@
...
+                        && workflowRun.getId() > latestPreexistingWorkflowRunId) {
+                    assertNotNull(workflowRun.getId());
+
+                    workflowRun.cancel();
+                    cancelledWorkflowRunId = workflowRun.getId();
+                    return true;
+                }
+            }
+            return false;
+        });
```
bitwiseman, line 121: A few requests applying to all the `await` sections: 
1. I'd prefer if the `until` only polled for the state change and the asserts are called after, outside the `until`.  
2.  Let's use `getGitHubBeforeAfter()` for polling.  That way the tests don't need to record the polling.  That way when running CI, the test do effectively no waiting.  
3.  With (2) we can remove the special polling interval values in `setup()`.  Maybe use a [fibonacci polling interval](https://github.com/awaitility/awaitility/wiki/Usage#fibonacci-poll-interval) with milliseconds? Recording will burn a few extra requests but they won't be recorded.

bitwiseman (changes requested):
Great work on this!  The only remaining feedback I have is on tests.  Sorry for the multiple rounds, this is the last one and I'll try to do a better job of batching for future PRs.

src/test/java/org/kohsuke/github/GHWorkflowTest.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Test
+    public void testDispatch() throws IOException {
+        GHWorkflow workflow = repo.getWorkflow("test-workflow.yml");
+
+        workflow.dispatch("main");
+        workflow.dispatch("main", Collections.singletonMap("parameter", "value"));
+
+        // if we implement the logs API at some point, it might be a good idea to validate all this
```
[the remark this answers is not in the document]
gsmet: Do you have an example in the code base?

src/test/java/org/kohsuke/github/GHWorkflowRunTest.java
gsmet: 1. yeah that was my inclination as well first and I thought it would be better to reduce the number of API calls. I changed it. I agree it's more readable this way.
2. ah, yes, good idea, I was looking for a way to do exactly that
3. I removed the polling when we don't use a proxy. I tried the fibonacci thingy but it doesn't work very well, don't know why but it was just running forever and I don't really have the cycles to dig more into it. Using a fixed poll interval works so I am sticking with that.

src/test/java/org/kohsuke/github/GHWorkflowTest.java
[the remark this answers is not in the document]
bitwiseman: I haven't used it in this project.
Here's the docs for it: http://wiremock.org/docs/verifying/

src/test/java/org/kohsuke/github/GHWorkflowRunTest.java
bitwiseman: Thanks, looks great. Bummer about the polling interval, but oh well.

src/test/java/org/kohsuke/github/GHWorkflowTest.java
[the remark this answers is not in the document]
gsmet: Thanks for the pointer. Done.

gsmet:
@bitwiseman I think I addressed all your comments. I would be interested in you releasing a new version once this is in so that I can implement this in my Quarkus GitHub App framework for Jeff!

bitwiseman (approved):

bitwiseman:
@gsmet  Great work!  I'll release today.

gsmet:
@bitwiseman thanks for the reviews and release! Glad we could get this one in.

PR #1065, opened 2021-03-26
Make Jacoco work on java 11

bitwiseman:
# Description 
It appears that the change to unblock jacoco on Java 16+ broke jacoco on Java 11.   I noticed this because running `mvn -Denable-ci install site" on Java 11 does not produce the jacoco report in the `target/site` folder.  Java 8 works., but I want to use Java 11 for releases, including site generation.  

@gsmet  Any thoughts on this?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -111,7 +111,7 @@
             <!-- attached to Maven test phase -->
             <execution>
               <id>report</id>
-              <phase>test</phase>
+              <phase>package</phase>
               <goals>
                 <goal>report</goal>
               </goals>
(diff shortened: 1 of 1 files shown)


gsmet (approved):
I thought it was needed for 11, just not required but if it causes issues, you can just make it 16+ (16 included).

bitwiseman:
@gsmet 
Yeah, I guess I'm concerned that since CI didn't detect this break, it might not detect whether 16+ is working either.

bitwiseman:
@gsmet Yup, looks like it isn't actually running on 16. Could you take a look and see if you can fix it.  I only ask because you made the change and you probably know most about it.

gsmet:
I think you can close this one as https://github.com/hub4j/github-api/pull/1079 should fix it.

PR #1067, opened 2021-03-29
Fix GHRepository#listWorkflows() and add a test

gsmet:
I don't know how I missed it but I forgot to add a test for `listWorkflows()` and... it didn't work properly.

I think we probably need a better infrastructure to deal with all these iterables to avoid having to add so many classes but that's something we can discuss another day :).


src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -2904,9 +2905,12 @@ public GHIssueEvent getIssueEvent(long id) throws IOException {
      * @return the paged iterable
      */
     public PagedIterable<GHWorkflow> listWorkflows() {
-        return root.createRequest()
-                .withUrlPath(getApiTailUrl("actions/workflows"))
-                .toIterable(GHWorkflow[].class, item -> item.wrapUp(root));
+        try {
+            GitHubRequest request = root.createRequest().withUrlPath(getApiTailUrl("actions/workflows")).build();
+            return new GHWorkflowsIterable(this, request);
```
bitwiseman, line 2910: Since this iterable only points to "actions/workflows" endpoint, you can late resolve the URL like we do here: 

https://github.com/hub4j/github-api/blob/6caaf9ea4e63650657ca3c6e4e16ada01870b7f8/src/main/java/org/kohsuke/github/GHSearchBuilder.java#L47-L54

You can even pass only `owner` and create the requester in `_iterator()`.

bitwiseman (changes requested):
Minor change suggested.

src/main/java/org/kohsuke/github/GHRepository.java
gsmet: Done!

bitwiseman: Hm, yeah.  Most other `list*` methods throw `IOException`.  Since you made this method not throw `IOException` (which I like) I'd rather have that method _really_ not throw.

It's probably silly, but thanks for being accommodating.

PR #1074, opened 2021-04-01
Add support for repository visibility

### LINKED ISSUE #932: GHRepository's isPrivate() returns true for Internal repositories
We see that isPrivate() method in GHRepository class returns true for Internal repositories too. We tried this with the latest version 1.116 of the API. 
We want to be able to distinguish between private and internal repositories. 
Is this possible with the current API?

nvahren:
# Description 
Adds support for the repository "visibility" field - "public" or "private," plus "internal" for Enterprise
Fixes #932

# Before submitting a PR:

# When creating a PR: 


nvahren:
I wasn't able to add a JUnit test for "internal" because it's exclusive to GitHub Enterprise, but I did do some manual testing against an enterprise instance:

Screen Shot 2021-04-01 at 9 26 55 AM

nvahren:
Also, if I can get access to `hub4j-test-org` I'll update the test snapshots as requested in the contribution instructions if that's preferable.

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -98,6 +98,8 @@
     @JsonProperty("private")
     private boolean _private;
 
+    private GHVisibility visibility;
+
```
bitwiseman, line 102: We were just having a discussion on another thread as to how to handle enums like this.  We settled on a storing them as `String` and converting when requested to avoid throwing exceptions if new values are ever added. 

See https://github.com/hub4j/github-api/pull/1064/commits/b8f00bc699c7de3667b236899fff6ec87d7038a8

src/test/java/org/kohsuke/github/GHRepositoryTest.java
```diff
@@ -249,6 +249,27 @@ public void testUpdateRepository() throws Exception {
         assertThat(redux.getDescription(), equalTo(updatedDescription));
     }
 
+    @Test
+    public void testGetRepositoryWithVisibility() throws IOException {
```
bitwiseman, line 253: Definitely want to test the `INTERNAL` value as well.  If you can create a test for it and then manually modify the data , that would be great.  After recording, have the test use ` snapshotNotAllowed();` so no one accidentally deletes that data.

bitwiseman (changes requested):
Generally looks good. Please make the tweaks requested and re-record using the hub4j-test-org. 
Thanks!

nvahren:
Thanks @bitwiseman, should be ready for another look.

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -704,6 +708,47 @@ public boolean isPrivate() {
...
+    }
+
+    /**
+     * Gets the visibility of the repository.
+     *
+     * @return the visibility
+     */
+    @Deprecated
+    @Preview(NEBULA)
+    @WithBridgeMethods(value = String.class, adapterMethod = "visibilityAsStr")
```
bitwiseman, line 734:
```suggestion
```

Bridge methods are only needed to maintain binary compatibility with previously published API.  You're adding new methods, so they are not needed.

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -704,6 +708,47 @@ public boolean isPrivate() {
...
+            }
+        }
+        return Visibility.from(visibility);
+    }
+
+    @SuppressFBWarnings(value = "UPM_UNCALLED_PRIVATE_METHOD", justification = "Bridge method of getVisibility")
+    private Object visibilityAsStr(Visibility visibility, Class type) {
+        return visibility;
+    }
+
```
bitwiseman, line 751:
```suggestion
```

bitwiseman:
Hm, I have no idea what I did to make this break. I'll look at it tomorrow.

nvahren:
I think this was the issue: https://github.com/hub4j/github-api/pull/1075/files#diff-a57ffc6640b7868f12be28cd63c1587ecbcb58c5551129d51838d8887059139dL61-R67

That change switched from importing all previews to importing specific ones, so I just needed to add NEBULA to the list.

bitwiseman (approved):

PR #1075, opened 2021-04-01
Add support for artifacts and logs of workflow runs

gsmet:
@bitwiseman I need your wisdom. The logs and artifacts are downloaded from another host (`https://pipelines.actions.githubusercontent.com/`) and the way it's done currently, this call is not mocked properly.

See https://github.com/hub4j/github-api/pull/1075/files#diff-87252b4de52f4e0e1317a91eee8a0de1cc885f7b84f872989c8e52ec09d2509dR296 for an example of how it is currently done.

I'm unsure of how we could handle this and thought I might as well ask.

Note: the PR is not completed yet, it's still missing tests for the artifacts. I only test the logs method for now.

```
mvn install -Dtest.github.takeSnapshot -Dtest=GHWorkflowRunTest#testLogs
```
works but running the test after that doesn't work as the actual download call wasn't mocked.


gsmet:
@jeffmaury I'm also interested in your feedback: do my additions cover your needs fully?

bitwiseman:
@gsmet 
You'll need to add another server mapping to the `GitHubWireMockRule`.  https://github.com/hub4j/github-api/commit/472034c95009d0f4a46e572bcedc8141327b9f46#diff-1087a33f8255b3b68f161247f32cb0f53a4457b8b92b4af5305c5f264c543a35 added `codeload`.  Repeat that for your new server.

src/main/java/org/kohsuke/github/GHWorkflowRun.java
```diff
@@ -261,6 +265,55 @@ public void rerun() throws IOException {
         root.createRequest().method("POST").withUrlPath(getApiRoute(), "rerun").fetchHttpStatusCode();
     }
 
+    /**
+     * Gets the artifacts associated with this workflow run.
+     *
+     * @return the paged iterable
+     */
+    public PagedIterable<GHArtifact> getArtifacts() {
```
bitwiseman, line 273:
```suggestion
    public PagedIterable<GHArtifact> listArtifacts() {
```

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -2961,6 +2968,36 @@ public GHWorkflowRun getWorkflowRun(long id) throws IOException {
...
+     *
+     * @return the paged iterable
+     */
+    public PagedIterable<GHArtifact> listArtifacts() {
+        try {
+            GitHubRequest request = root.createRequest().withUrlPath(getApiTailUrl("actions/artifacts")).build();
+            return new GHArtifactsIterable(this, request);
+        } catch (MalformedURLException e) {
+            throw new GHException("Malformed URL", e);
+        }
```
bitwiseman, line 2982: Here's that try-catch again.  I figured out why most other `list*` methods don't need this: 

https://github.com/hub4j/github-api/blob/a68d16d5de6da437fa493f3bd524de03eaaee1ad/src/main/java/org/kohsuke/github/Requester.java#L153-L158

This try catch appears in two places so in this case I definitely want it inside the iterable.

src/main/java/org/kohsuke/github/GHWorkflowRunsIterable.java
```diff
@@ -8,21 +8,22 @@
  * Iterable for workflow runs listing.
  */
 class GHWorkflowRunsIterable extends PagedIterable<GHWorkflowRun> {
-    private final transient GitHub root;
+    private final transient GHRepository owner;
```
bitwiseman, line 11:
```suggestion
    private final GHRepository owner;
```

jeffmaury (approved):
Should cover my needs

src/main/java/org/kohsuke/github/GHWorkflowRun.java
```diff
@@ -261,6 +265,55 @@ public void rerun() throws IOException {
...
+     *             The IO exception.
+     * @return the result of reading the stream.
+     */
+    public <T> T downloadLogs(InputStreamFunction<T> streamFunction) throws IOException {
+        requireNonNull(streamFunction, "Stream function must not be null");
+
+        String logsLocation = root.createRequest()
+                .method("GET")
+                .withUrlPath(getApiRoute(), "logs")
+                .fetchRedirectLocation();
```
bitwiseman, line 302: Interesting.  It doesn't automatically redirect? I think this one does.  
https://github.com/hub4j/github-api/blob/master/src/test/resources/org/kohsuke/github/GHRepositoryTest/wiremock/testZipball/mappings/repos_hub4j-test-org_temp-testzipball_zipball-3.json

 I'm surprised I feel sure we've had other endpoints that redirect. Could you test this with the okhttp3 connector?

gsmet: No, it doesn't. I tried without this magic as I thought it would work and I had an empty content in return. And I was doing the exact same thing than the code you pointed out.

But are you sure the redirect is working in the example you showed? Because there are no recorded request/response after that and I would expect the request to fail if not recorded.

Moreover the test would work even with an empty content.

src/main/java/org/kohsuke/github/GHRepository.java
gsmet: I knew you would say that but the query is different in both cases, thus why I haven't changed it as for the other PR. So should I pass the builder instead of the request and build the request in the iterable?

Frankly, for me the issue is with the initial design decision of throwing low value unchecked exceptions from nearly everywhere.

src/main/java/org/kohsuke/github/GHWorkflowRun.java
gsmet: Apparently, the fact that it didn't follow redirects was a direct consequence of things not being implemented in Wiremock. The issue went away when I implemented the Wiremock server (well, indeed, the tests were broken because suddently the redirect was properly followed).

gsmet:
OK, things look significantly better now. The logs tests are passing. I need to add tests for the artifacts now. I'll probably do that this evening.

@bitwiseman I'll wait for your feedback on the `MalformedURLException` question and I'll fix it accordingly.

gsmet:
@bitwiseman this is now complete - I implemented all the features I wanted and the code coverage should be OK. Let me know what you want me to do for the `MalformedURLException` and I'll adjust the PR.

Feel free to ask other adjustments too :).

src/main/java/org/kohsuke/github/GHRepository.java
bitwiseman: > I knew you would say that but the query is different in both cases, thus why I haven't changed it as for the other PR. So should I pass the builder instead of the request and build the request in the iterable?

Yeah, pass the builder.  Or maybe the right thing is to make a version of the `build()` that wraps those exceptions and use that in all 3 (4?) calls.  I would change the behavior of `build()` but that could result in client that understand how to handle certain exceptions suddenly not getting those anymore. 

> Frankly, for me the issue is with the initial design decision of throwing low value unchecked exceptions from nearly everywhere.

Absolutely.  Please file an issue and describe how it should work in the next version.  I'm interested.  Some of the current behavior is probably cruft from long removed code.  It may also be a result of the underlying implementation of `HttpConnector` and `HttpUrlConnection` not being controlled by this library.  `okhttp3` has notably different behavior from the Java built-in implementation.

gsmet: OK, for now, I passed the builder to the iterable.

I think more involved changes need to be thought out a bit more. I'll think about it and open an issue.

gsmet:
All green!

bitwiseman (approved):

gsmet:
Thanks! As usual, interested by a release when you get the time. We want to put this to good use :).

bitwiseman:
Sure thing, early next week.  Thanks for your contributions!

PR #1079, opened 2021-04-04
Fix JaCoCo execution when we have additional Surefire options

gsmet:
Typically on Java 16.

Note that the build then fails with Javadoc warnings. Apparently, the Java 16 `javadoc` is a bit stricter. Not sure if we want to disable it for Java 16 for now or if someone wants to fix them (but there are a bunch of them and that would mean documenting very obvious stuff).


pom.xml
```diff
@@ -274,7 +278,7 @@
             <id>default-test</id>
             <configuration>
               <excludesFile>src/test/resources/slow-or-flaky-tests.txt</excludesFile>
-              <argLine>${surefire.argLine}</argLine>
+              <argLine>@{jacoco.surefire.argLine} ${surefire.argLine}</argLine>
```
gsmet, line 281: To explain what happened: we are setting our own arg line now and, if set, the JaCoCo agent doesn't inject its options.

The change injects again the JaCoCo options.

PR #1081, opened 2021-04-05
Add support for workflow jobs

gsmet:


src/main/java/org/kohsuke/github/GHWorkflowRun.java
```diff
@@ -285,8 +285,9 @@ public void rerun() throws IOException {
     /**
      * Downloads the logs.
      * <p>
-     * The logs are in the form of a zip archive. The full log file is at the root and called {@code 1_build.txt}. Split
-     * log files are also available in the {@code build} directory.
```
gsmet, line 289: This information was incorrect so I removed it. The structure of the log is more complex than that and depends on job names.

src/main/java/org/kohsuke/github/GHWorkflowRunJob.java
```diff
@@ -0,0 +1,256 @@
...
+import java.util.Objects;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * A workflow run job.
+ *
+ * @author Guillaume Smet
+ */
+public class GHWorkflowRunJob extends GHObject {
```
gsmet, line 23: I thought `GHJob` would be a bit too generic but I can change it if you prefer it.

src/main/java/org/kohsuke/github/GHWorkflowRunJob.java
```diff
@@ -0,0 +1,256 @@
...
+
+    GHWorkflowRunJob wrapUp(GitHub root) {
+        this.root = root;
+        if (owner != null) {
+            owner.wrap(root);
+        }
+        return this;
+    }
+
+    public static class Step {
```
gsmet, line 187: I decided to go for a simple inline class here. Can change it if you prefer something else.

src/main/java/org/kohsuke/github/GHWorkflowRunJobQueryBuilder.java
```diff
@@ -0,0 +1,47 @@
+package org.kohsuke.github;
+
+import java.net.MalformedURLException;
+
+/**
+ * Lists up jobs of a workflow run with some filtering.
+ *
+ * @author Guillaume Smet
+ */
+public class GHWorkflowRunJobQueryBuilder extends GHQueryBuilder<GHWorkflowRunJob> {
```
gsmet, line 10: Even if there's not really search filters, I decided to go with this pattern to be consistent with the work I started (but not yet pushed) on the check runs. They are implemented in the exact same way (so with this weird `filter` thing) but with more criteria.

src/main/java/org/kohsuke/github/GHWorkflowRunJob.java
bitwiseman: The GitHub docs refer to ["Workflow Jobs"](https://docs.github.com/en/rest/reference/actions#workflow-jobs), but I'm also okay with "Workflow Run Jobs".

src/main/java/org/kohsuke/github/GHWorkflowRunJobQueryBuilder.java
gsmet: I kept this class but I pushed an additional commit to hide it from the users as for now it doesn't bring anything valuable.

src/main/java/org/kohsuke/github/GHWorkflowRunJob.java
gsmet: Yeah, I wanted to make it clear they are associated with a run.

gsmet: I thought about it a bit more and `Workflow job` is less of a mouthful so I will switch to it.

src/main/java/org/kohsuke/github/GHWorkflowRunJob.java
bitwiseman: I have no particular objection to inner classes when they make sense.  This seems reasonable.  What was your concern here?

gsmet: Well, it's always hard to predict if at some point they will add more stuff into this and it would have made sense to make it a root level object. But I think we are relatively safe here.

src/main/java/org/kohsuke/github/GHWorkflowRunJobQueryBuilder.java
bitwiseman: I'm all in favor of less API surface area.

gsmet:
@bitwiseman from my side, it's good to go. Looking forward to your review.

(better wait for this one to get in before releasing!)

bitwiseman (commented):
Typo fixes.

gsmet:
Hey @bitwiseman , any chance you could release 1.127 today or tomorrow? I'm currently working with snapshots but I'd like to make my work public so that others could have a look and contribute.

Thanks!

bitwiseman:
Sure, I'll release tonight.

gsmet:
@bitwiseman thanks!

PR #1092, opened 2021-04-12
Properly wrap the check runs with the repository when listing them

gsmet:
Otherwise you have a not so nice NPE when trying to update them:
```
java.lang.NullPointerException
	at org.kohsuke.github.GHCheckRunBuilder.<init>(GHCheckRunBuilder.java:75)
	at org.kohsuke.github.GHCheckRun.update(GHCheckRun.java:339)
```

I added a basic test to check we are able to call the builder as that's the only problem I can foresee there.


gsmet:
The coverage report is a bit suspicious given what has been changed.

bitwiseman:
No worries, I'm just giving this a try.

bitwiseman (approved):

PR #1095, opened 2021-04-14
Log 202 message at the FINE level

gsmet:
Also reintroduce the special case we had before as the info is
important.

@bitwiseman you removed the special cases to improve code coverage but I don't think it's a good idea.

For me, either we keep the logs and make them informative or we remove the log altogether given 202 is a success. It's just that the operation takes time.

In the case of cancel, I dropped the log as really you just want to cancel the thing, you don't really care that it will take time.

### DIFF
--- a/src/main/java/org/kohsuke/github/GitHubClient.java
+++ b/src/main/java/org/kohsuke/github/GitHubClient.java
@@ -438,15 +438,15 @@ abstract class GitHubClient {
             // special case handling for 304 unmodified, as the content will be ""
         } else if (responseInfo.statusCode() == HttpURLConnection.HTTP_ACCEPTED) {
 
-            // Response code 202 means data is being generated.
+            // Response code 202 means data is being generated or an action that can require some time is triggered.
             // This happens in specific cases:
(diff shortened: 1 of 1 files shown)


bitwiseman:
@gsmet 
We should probably implement a 202 handling behavior in GitHubClient.  Currently, forking and statistics both implement their own waiting behavior.  However, now that I look at it [forking returns a `GHRepository` on success](https://docs.github.com/en/rest/reference/repos#create-a-fork), so maybe the waiting behavior isn't correct. 
I think the point of the logging is that if we're getting a 202, it probably needs some special handling.

In the short term, what do you think of keeping the single log call but making it a `FINE`?

gsmet:
Yeah, I'm not sure if we should handle the retry in the GitHub API itself or stay at the level of the REST API. I don't know what the Octokit is doing.

But yeah maybe logging at FINE level is OK for now. I just didn't want to get the warning for cancel operations but we can generalize it.

I'll change my PR accordingly.

gsmet:
Adjusted!

bitwiseman (approved):

PR #1096, opened 2021-04-14
Add SCHEDULE to GHEvent and add UNKNOWN handling to GHEvent

gsmet:
`SCHEDULE` is a valid event for scheduled workflow runs.

Also make `GHEvent` parsing more permissive as per our discussion from the previous weeks.


bitwiseman (changes requested):
Great!  
Could you also update [GHEventInfo#getType()](https://github.com/hub4j/github-api/blob/e66a72387e6fcd5762e0122f8da17d271d27a756/src/main/java/org/kohsuke/github/GHEventInfo.java#L48-L57)?

gsmet:
@bitwiseman I added two commits:

* the first one returns `UNKNOWN` in `GHEventInfo.getType` but I haven't changed the logic as it's doing some manipulation we don't do in `EnumUtils` (and they look a bit too aggressive to be a good fit for a general purpose util)
* I added a log warning if we don't find a value. This way, we can get some feedback about the values that should be added in GitHub API.

src/main/java/org/kohsuke/github/GHEventInfo.java
```diff
@@ -53,7 +53,7 @@ public GHEvent getType() {
             if (e.name().replace("_", "").equalsIgnoreCase(t))
                 return e;
         }
```
bitwiseman, line 55: Ugh, I see the problem now.  All this is to handle the completely different strings used in the REST API and actual event calls. 
https://github.com/hub4j/github-api/blob/master/src/test/resources/org/kohsuke/github/AppTest/wiremock/testEventApi/__files/events-10.json#L283

There's got to be a better way to do this.  I'll file an issue to come back to this.

bitwiseman (approved):

PR #1124, opened 2021-04-21
upgrade commons-io due to CVE-2021-29425

alexanderkjall:
# Description

Upgrade commons-io to 2.8.0, it's a vulnerability in FileNameUtils.normalize that isn't used in this project.

More information about the vulnerability can be found here: https://issues.apache.org/jira/browse/IO-556

I have run the tests locally (which failed to to unrelated javadoc warnings) and done a basic check in my application that uses github-api, but not done any other testing.

### DIFF
--- a/src/test/java/org/kohsuke/github/ArchTests.java
+++ b/src/test/java/org/kohsuke/github/ArchTests.java
@@ -153,7 +153,7 @@ public class ArchTests {
     @Test
     public void testRequireUseOfOnlySpecificApacheCommons() {
 
-        final ArchRule onlyApprovedApacheCommonsLang3Methods = classes()
+        final ArchRule onlyApprovedApacheCommonsMethods = classes()
                 .should(notCallMethodsInPackageUnless("org.apache.commons..",
                         // unless it is one of these methods
                         targetMethodIs(StringUtils.class, "capitalize", String.class),
@@ -187,9 +187,9 @@ public class ArchTests {
                         targetMethodIs(IOUtils.class, "toString", Reader.class),
                         targetMethodIs(IOUtils.class, "toByteArray", InputStream.class)))
                 .because(
-                        "Commons methods must be manually verified to be compatible with commons-io:2.4 or earlier and commons-lang3:3.9 or earlier should be used.");
+                        "Commons methods must be manually verified to be compatible with commons-io:2.4 or earlier and commons-lang3:3.9 or earlier.");
 
-        onlyApprovedApacheCommonsLang3Methods.check(classFiles);
+        onlyApprovedApacheCommonsMethods.check(classFiles);
     }
 
     public static ArchCondition<JavaClass> notCallMethodsInPackageUnless(final String packageIdentifier,
--- a/src/test/java/org/kohsuke/github/ArchTests.java
+++ b/src/test/java/org/kohsuke/github/ArchTests.java
@@ -175,7 +175,7 @@ public class ArchTests {
                         target(HasOwner.Predicates.With.<JavaClass>owner(resideInAPackage("org.apache.commons..")))
                                 .and(DescribedPredicate.not(approvedApacheCommonsMethods)))))
                 .because(
-                        "Only commons methods that have been manually verified to be compatible with v2.4 should be used.");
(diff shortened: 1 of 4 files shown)


bitwiseman (changes requested):
This project is used by other projects that have pinned a lower version of `commons-io`. 
I think it makes sense for use to upgrade the version we use by default but we need to enforce only using methods that are compatible with lower versions.  Please add a test to [ArchTests.java](https://github.com/hub4j/github-api/blob/main/src/test/java/org/kohsuke/github/ArchTests.java) that checks that we are only using a specified approved list for methods from `commons-io`.

bitwiseman:
As noted, this library does not use the method effected by this CVE so it is not urgent to merge and release.

alexanderkjall:
I looked a bit on ArchTests and to be honest I don't understand how to write a unit test that would fail if a function that's not in the library version 2.4 is used?

bitwiseman:
@alexanderkjall 
It's easy! You just... 
Wait, no, I mean, ArchUnit is pretty friendly, you can...

Oh, well, it looks like the test I wrote to enforce using `assertThat()` during tests was wrong. 

Also, checking the methods is not at all intuitive and the documentation for this scenario is not helpful.  

So, thanks for asking. I've implemented a minimal test that does not ensure non-breakage, but will force people to consider when that add commons method calls.

bitwiseman (approved):

bitwiseman:
@gsmet @alexanderkjall 
Could you review these changes?

src/test/java/org/kohsuke/github/ArchTests.java
```diff
@@ -113,14 +138,103 @@ public void testBetaApisAreFlaggedAsDeprecated() {
...
                 .because(reason);
 
-        onlyAssertThatRule.check(tesetClassFiles);
+        onlyAssertThatRule.check(testClassFiles);
+    }
+
+    @Test
+    public void testRequireUseOfOnlySpecificApacheCommons() {
+
+        final ArchRule onlyApprovedApacheCommonsLang3Methods = classes()
```
gsmet, line 156:
```suggestion
        final ArchRule onlyApprovedApacheCommonsMethods = classes()
```
`IOUtils` is in commons-io.

src/test/java/org/kohsuke/github/ArchTests.java
```diff
@@ -113,14 +138,103 @@ public void testBetaApisAreFlaggedAsDeprecated() {
...
+                                Object.class,
+                                Boolean.class),
+                        targetMethodIs(ReflectionToStringBuilder.class, "accept", Field.class),
+                        targetMethodIs(IOUtils.class, "closeQuietly", InputStream.class),
+                        targetMethodIs(IOUtils.class, "closeQuietly", Closeable.class),
+                        targetMethodIs(IOUtils.class, "toString", InputStream.class, Charset.class),
+                        targetMethodIs(IOUtils.class, "toString", Reader.class),
+                        targetMethodIs(IOUtils.class, "toByteArray", InputStream.class)))
+                .because(
+                        "Commons methods must be manually verified to be compatible with commons-io:2.4 or earlier and commons-lang3:3.9 or earlier should be used.");
```
gsmet, line 190:
```suggestion
                        "Commons methods must be manually verified to be compatible with commons-io:2.4 or earlier and commons-lang3:3.9 or earlier.");
```

src/test/java/org/kohsuke/github/ArchTests.java
```diff
@@ -113,14 +138,103 @@ public void testBetaApisAreFlaggedAsDeprecated() {
...
+
+        onlyApprovedApacheCommonsLang3Methods.check(classFiles);
```
gsmet, line 192:
```suggestion
        onlyApprovedApacheCommonsMethods.check(classFiles);
```

gsmet (changes requested):
Spotted a few minor things.

I'm not very familiar with ArchUnit but I suppose you have tested adding another method fails the build?

bitwiseman:
@gsmet  Yes, I've verified that methods not on the list are caught by the test.

jeffalder:
@bitwiseman - Is it possible to get a release cut that includes this PR? I'd be interested in upgrading to make the CVE go away.

bitwiseman:
@jeffalder Done.

Issue #1099, opened 2021-04-14
Streamline GHEvent parsing from GHEventInfo.type 


bitwiseman:
GHEventInfo.type uses Pascal-case names with "Event" on the end instead of snake-case names used elsewhere in the API.  `getType()` does a bunch of processing every time it parses `type` into a `GHEvent`.  We should clean this up.
https://github.com/hub4j/github-api/pull/1096/files#r613635046

akashRindhe:
I'd like to take a crack at this but I'd like to discuss the possible approaches to fix this.

Would a static `Map` from the GitHub defined event type in `Pascal` case to the github-api defined `GHEvent` enum which will be used during the `getType()` call be an acceptable approach?

bitwiseman:
That sounds fine to me.

PR #1126, opened 2021-04-21
(refactor) Replace complex parsing logic from GHEvent.type to GHEvent with static mapping

akashRindhe:
https://github.com/hub4j/github-api/issues/1099

# Description

Replace complex parsing logic from GHEvent.type to GHEvent with static mapping

### DIFF
--- a/src/main/java/org/kohsuke/github/GHEvent.java
+++ b/src/main/java/org/kohsuke/github/GHEvent.java
@@ -2,6 +2,8 @@ package org.kohsuke.github;
 
 import java.util.Locale;
 
+import javax.annotation.Nonnull;
+
 /**
  * Hook event type.
  *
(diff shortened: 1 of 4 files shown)


src/main/java/org/kohsuke/github/GHEvent.java
```diff
@@ -85,4 +87,50 @@ String symbol() {
...
         return name().toLowerCase(Locale.ENGLISH);
     }
+
+    /**
+     * Representation of GitHub Event Type
+     *
+     * @see <a href="https://docs.github.com/en/developers/webhooks-and-events/github-event-types">GitHub event
+     *      types</a>
+     */
+    public enum GitHubEventType {
```
bitwiseman, line 97: I don't think this needs to be public.  It is only used like an private`HashMap<String, GHEvent>`.  
```suggestion
    enum GitHubEventType {
```

src/test/java/org/kohsuke/github/GHEventTest.java
```diff
@@ -0,0 +1,34 @@
...
+        for (GHEvent e : GHEvent.values()) {
+            if (e.name().replace("_", "").equalsIgnoreCase(t)) {
+                return e;
+            }
+        }
+        return GHEvent.UNKNOWN;
+    }
+
+    @Test
+    public void regressionTest() {
```
bitwiseman, line 28: Add a test for the `GHEvent.UNKNOWN` behavior.

bitwiseman (changes requested):
This is not what I was expecting, but it works, is clearer, and it has to be faster than the old way. 

Minor changes needed.

src/main/java/org/kohsuke/github/GHEvent.java
```diff
@@ -85,4 +87,50 @@ String symbol() {
...
+        IssueCommentEvent(ISSUE_COMMENT),
+        IssuesEvent(ISSUES),
+        MemberEvent(MEMBER),
+        PublicEvent(PUBLIC),
+        PullRequestEvent(PULL_REQUEST),
+        PullRequestReviewEvent(PULL_REQUEST_REVIEW),
+        PullRequestReviewCommentEvent(PULL_REQUEST_REVIEW_COMMENT),
+        PushEvent(PUSH),
+        ReleaseEvent(RELEASE),
+        WatchEvent(WATCH);
```
bitwiseman, line 112:
```suggestion
        WatchEvent(WATCH),
        Unknown(UNKNOWN);
```

src/main/java/org/kohsuke/github/GHEvent.java
```diff
@@ -85,4 +87,50 @@ String symbol() {
...
+         * @param event
+         *            the github event as a string to convert to Event enum
+         * @return GHEvent
+         */
+        public static GHEvent transformToGHEvent(@Nonnull String event) {
+            try {
+                GitHubEventType gitHubEventType = GitHubEventType.valueOf(event);
+                return gitHubEventType.event;
+            } catch (IllegalArgumentException ignored) {
+                return UNKNOWN;
```
bitwiseman, line 132:
```suggestion
            // import static org.kohsuke.github.internal.EnumUtils;
            GitHubEventType gitHubEventType = EnumUtils.getNullableEnumOrDefault(GitHubEventType.class, event, Unknown);                     
            return gitHubEventType.event;
```

bitwiseman (commented):
If you use EnumUtils t hat would be better.

akashRindhe:
Thanks for making the changes. I believe everything is resolved now and is good to merge?

bitwiseman:
@akashRindhe 
FYI, I'm sure you could have made the changes yourself, but I felt like getting them done while I was looking at them.   Hope you don't mind.  Thanks for your contributions.

akashRindhe:
> @akashRindhe
> FYI, I'm sure you could have made the changes yourself, but I felt like getting them done while I was looking at them. Hope you don't mind. Thanks for your contributions.

No issues mate. Cheers :)

PR #1136, opened 2021-05-06
GHPullRequest: do not refresh for merge_commit_sha

NicoK:
# Description
The `merge_commit_sha` field of `GHPullRequest` is actually already present and we shouldn't need to refresh the pull request for it.
See https://docs.github.com/en/rest/reference/pulls#list-pull-requests

If you want to be extra-safe on this (I'm not sure this condition holds for all sources of creating `GHPullRequest` instances), we could do a `null`-check for the refresh...


bitwiseman:
https://docs.github.com/en/rest/reference/pulls#get-a-pull-request 

> The value of the `mergeable` attribute can be 'true', 'false', or 'null'. If the value is 'null', then GitHub has started a background job to compute the mergeability. After giving the job time to complete, resubmit the request. When the job finishes, you will see a non-null value for the 'mergeable' attribute in the response. If 'mergeable' is 'true', then 'merge_commit_sha' will be the SHA of the test merge commit.

What doesn't say is if `mergeable` is null, the `merge_commit_sha` value may still be equal to the old value.

It looks like we do not have a test that covers this.  I'll go make one.

bitwiseman:
@NicoK  Okay, the mistake.  We're using `mergeable_state` instead of `mergeable`.  And `mergeable_state` is not `null` in any of or test cases and it basically is never null.  It comes in as `unknown` sometimes, but not `null`.  

I'm not sure whether customers depend on this behavior or not, or if they're rather have it refresh. I would guess many already check for mergeable (as in the code linked above).

NicoK:
Thanks for looking into this. I'm not sure about this `mergeable_state` stuff. I was mainly using this to query pull requests and at least the /pulls resource seems to fill this field already without having to get all the PR details (which is yet another API call that is on a tight rate limit). Maybe we also need to distinguish results from different endpoints here...

bitwiseman:
> Thanks for looking into this. I'm not sure about this mergeable_state stuff. I was mainly using this to query pull requests and at least the /pulls resource seems to fill this field already without having to get all the PR details (which is yet another API call that is on a tight rate limit)

`populate()` checks that the `mergeable_state` is not `null` which will be the case when calling the `pulls` endpoint.  So, refresh() should not be called.

https://github.com/hub4j/github-api/blob/674d7cd05383973e28b0ca304a76bec7410c48ed/src/main/java/org/kohsuke/github/GHPullRequest.java#L369-L373

Are you seeing a different behavior? 

> Maybe we also need to distinguish results from different endpoints here...

That sounds like a really good idea.  We're going to need to do something like this for GraphQL as well, should we ever support it - since GraphQL queries can return a configurable subset of fields.

bitwiseman:
@NicoK  
Did you see my question above?

NicoK:
yes @bitwiseman, sorry - I haven't had time to check it out yet; hopefully late next week

bitwiseman:
@NicoK  
Since you don't have time to work on this PR, I'm going to close it for now.  I'm happy to reopen when you have time.

Issue #1080, opened 2021-04-04
Add method to check if organization has projects enabled


Katsute:
**Describe the bug**

Trying to retrieve organization project boards when they are disabled causes a `GHException`.

**To Reproduce**

1. Disable organization project boards.
2. Use `GHOrganization#listProjects()`

**Expected behavior**

Behavior is expected, however a boolean method would be nice.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Version: 1.126

**Additional context**

Add a new method `GHOrganization#hasProjects()`, similar to `GHRepository#hasProjects()` to check if projects are enabled.

This information can be retrieved from the organization json under the key `has_organization_projects`.

akashRindhe:
@bitwiseman Please help close this issue as the PR has been merged and it is part of the latest release.
Also going forward, is there some way to link the PR to the issue so that the issue will be automatically closed once the PR is merged?

Katsute:
Closed in #1139

bitwiseman:
@akashRindhe 
If you put the text "Fixes #issue-number" in your PR it will automatically close the linked issue.

PR #1139, opened 2021-05-09
(feat) Add method to check if Organization has projects enabled

akashRindhe:
[#1080]

# Description
Adding new method `hasProjects` to `GHOrganization` to allow for checking if a GitHub Org has projects enabled


akashRindhe:
I'd like to get permission to the `hub4j-test-org` in order to write a test for this new method.

src/main/java/org/kohsuke/github/GHOrganization.java
```diff
@@ -18,6 +18,9 @@
  * @author Kohsuke Kawaguchi
  */
 public class GHOrganization extends GHPerson {
+
+    private boolean has_organization_projects;
```
bitwiseman, line 22: https://docs.github.com/en/rest/reference/orgs#get-an-organization 
This shows two different projects fields. Do you know what the difference is? 

```
  "has_organization_projects": true,
  "has_repository_projects": true,
```

src/test/java/org/kohsuke/github/GHOrganizationTest.java
```diff
@@ -231,4 +231,31 @@ public void testCreateAllArgsTeam() throws IOException {
...
+        // Assert that projects are enabled
+        assertThat(result, is(true));
+
+        // Disable projects
+        // Arrange
+        org.root.createRequest()
+                .withUrlPath("/orgs/" + GITHUB_API_TEST_ORG)
+                .method("PATCH")
+                .with("has_organization_projects", false)
+                .fetchInto(org);
```
bitwiseman, line 253: Since this exists you might as well add a public method to `GHOrganization` to update this value and test that as well.

src/main/java/org/kohsuke/github/GHOrganization.java
akashRindhe: image

From the docs.

src/main/java/org/kohsuke/github/GHOrganization.java
```diff
@@ -367,6 +370,35 @@ public void conceal(GHUser u) throws IOException {
...
     }
 
+    /**
+     * Are projects enabled for organization boolean.
+     *
+     * @return the boolean
+     */
+    public boolean areProjectsEnabledForOrganization() {
+        return has_organization_projects;
+    }
```
bitwiseman, line 380: I like having more meaningful names, but it looks like we didn't do that for repository:

https://github.com/hub4j/github-api/blob/b23934a5a15ccc4964a20a7ab422ee894dd9bbfb/src/main/java/org/kohsuke/github/GHRepository.java#L619-L621

I'd suggest the following to keep the same pattern. 

```suggestion
    public boolean hasOrganizationProjects() {
        return has_organization_projects;
    }
```

However, I'd also be okay with the following: 

```suggestion
    public boolean areOrganizationProjectsEnabled() {
        return has_organization_projects;
    }
```

I'm also open to discussion.

src/test/java/org/kohsuke/github/GHOrganizationTest.java
```diff
@@ -231,4 +231,30 @@ public void testCreateAllArgsTeam() throws IOException {
         assertThat(team.getDescription(), equalTo("Team description"));
         assertThat(team.getPrivacy(), equalTo(GHTeam.Privacy.CLOSED));
     }
+
+    @Test
+    public void testAreProjectsEnabledForOrganization() throws IOException {
```
bitwiseman, line 236: Please use [nonRecordingGitHub()](https://github.com/hub4j/github-api/blob/b23934a5a15ccc4964a20a7ab422ee894dd9bbfb/src/test/java/org/kohsuke/github/AbstractGitHubWireMockTest.java#L230) to ensure the setting is `true` before starting.

src/test/java/org/kohsuke/github/GHOrganizationTest.java
```diff
@@ -231,4 +231,30 @@ public void testCreateAllArgsTeam() throws IOException {
...
+
+        // Act
+        boolean result = org.areProjectsEnabledForOrganization();
+
+        // Assert
+        assertThat(result, is(true));
+    }
+
+    @Test
+    public void testSetProjectsEnabledForOrganization() throws IOException {
```
bitwiseman, line 248: Please use [nonRecordingGitHub()](https://github.com/hub4j/github-api/blob/b23934a5a15ccc4964a20a7ab422ee894dd9bbfb/src/test/java/org/kohsuke/github/AbstractGitHubWireMockTest.java#L230) to ensure the setting is `true` before starting, then toggle it to `false`.

bitwiseman (changes requested):
Naming is hard.  Thanks for your patience with multiple reviews. I think I got everything this time.

src/main/java/org/kohsuke/github/GHOrganization.java
akashRindhe: I understand consistency in naming to reduce overhead in remembering API names for the clients. However, the name `hasOrganizationProjects` answers the question whether the `Organization` has projects rather than _are projects enabled for an organization_ . 
Your second suggestion is definitely an improvement over the name I've given and will run with it.

bitwiseman (approved):

bitwiseman:
Thanks!

PR #1159, opened 2021-05-27
Fix issues related to introduction of new values in GHEvent

### LINKED ISSUE #1156: Checks api functionality broken by recent github api change
Since 26/05 previously working calls to `createCheckRun` fail with the following error when trying to parse the github response

```
Cannot deserialize value of type `org.kohsuke.github.GHEvent` from String "discussion": not one of the values accepted for Enum class: [GOLLUM, ALL, INTEGRATION_INSTALLATION_REPOSITORIES, META, DOWNLOAD, FORK_APPLY, CHECK_RUN, MARKETPLACE_PURCHASE, PROJECT_CARD, SCHEDULE, PUSH, PAGE_BUILD, TEAM_ADD, REPOSITORY_VULNERABILITY_ALERT, PACKAGE, UNKNOWN, DELETE, DEPLOYMENT_STATUS, WATCH, FOLLOW, ORG_BLOCK, ORGANIZATION, COMMIT_COMMENT, ISSUE_COMMENT, REPOSITORY_IMPORT, GITHUB_APP_AUTHORIZATION, CODE_SCANNING_ALERT, CHECK_SUITE, FORK, GIST, CONTENT_REFERENCE, REPOSITORY, INSTALLATION_REPOSITORIES, REPOSITORY_DISPATCH, MILESTONE, STAR, LABEL, MEMBERSHIP, SECURITY_ADVISORY, PROJECT_COLUMN, TEAM, PULL_REQUEST_REVIEW, RELEASE, PUBLIC, WORKFLOW_RUN, STATUS, PULL_REQUEST, PROJECT, WORKFLOW_DISPATCH, CREATE, MEMBER, REGISTRY_PACKAGE, INSTALLATION, DEPLOYMENT, PULL_REQUEST_REVIEW_COMMENT, DEPLOY_KEY, PING, ISSUES]
```

It isn't instantly obvious to me how the full github response can be logged, but a little debugging suggests that gh events now include 'DISCUSSION' and 'DISCUSSION_COMMENT'.

Unfortunately I think there have been further changes. Patching the enum to include these values moves things further, but calls to

```
   repo.updateCheckRun(checkRun.getId())
      .add(output)
      .withConclusion(GHCheckRun.Conclusion.SUCCESS)
      .withStatus(GHCheckRun.Status.COMPLETED)
      .create();
```

Now fail with 

`org.kohsuke.github.GHFileNotFoundException: https://api.github.com/repos/ORG/REPO/check-runs/2675563029 {"message":"Not Found","documentation_url":"https://docs.github.com/rest"}`

The check id is presumably correct as it was returned from the successful `createCheckRun` call, and this code previously functioned.

I think authentication is also affected. From memory, I was previously able to run adhoc tests of api calls using a personal access token. I now get a message saying that I must authenticate as a github app in order to use the checks api.

gsmet:
@bitwiseman I decided to apply the same strategy we used for other things: use Strings for serialization/deserialization and only switch to enums for APIs. It's a bit less optimal but less risky. Another better option would be to have custom Jackson deserializer but that's probably something we need to keep for 2.0.

I also fixed an inconsistency in `EnumUtils`: one method was user uppercase values and not the other. This is a recipe for disaster.

Better reviewed commit per commit as I kept them semantic.

I'm not exactly sure if it's worth adding tests as we would have to add tests with unknown values for each enum all over the place.

Fixes #1156


src/main/java/org/kohsuke/github/internal/EnumUtils.java
```diff
@@ -49,7 +49,7 @@
      */
     public static <E extends Enum<E>> E getEnumOrDefault(Class<E> enumClass, String value, E defaultEnum) {
         try {
-            return Enum.valueOf(enumClass, value);
+            return Enum.valueOf(enumClass, value.toUpperCase(Locale.ROOT));
```
bitwiseman, line 52: I actually don't want to do this.  The values are case-sensitive on purpose.  I'm open to discussing changing this but not as part of this PR.

bitwiseman (changes requested):
The case sensitivity is intentional.

gsmet:
Mkay. My big concern is that we have two nearly identical methods and the nullable behavior should be the only difference between the two.

I don't mind how we fix it but I think we need to make the methods consistent in one way or another.

I will remove this commit from the PR.

bitwiseman:
@gsmet 
 You are completely correct in principle, and I'm almost sure you're right in this case.  

I think the reason I chose to do it that way was to match the exiting method's behavior (which was case-sensitive). The events area is not well tested, so I would like to err on the side of caution.  Perhaps not it as an issue and link to https://github.com/hub4j/github-api/pull/1159/commits/38460dc8a9703feac2c6132d49b0d0c160864955 and we can come back to it?

gsmet:
OK, I removed this commit and did the case change at the call sites.

hcoles:
Is there anything I can do to help get this merged?

sathishavm:
I like this change needs to be merged

Issue #1157, opened 2021-05-26
updateCheckRun call fails since 1.128


hcoles:
As originally raised in #1156, calls to update check runs that worked in 1.122 now fail once the enum mapping issue is worked around.

The following code

```
   GitHub github = new GitHubBuilder()
       .withOAuthToken("value supplied by github actions")
       .withEndpoint("value supplied by github actions")
       .build();

   GHCheckRun checkRun = repo.createCheckRun("whatever", "value supplied by github actions").withStatus(GHCheckRun.Status.IN_PROGRESS).create();

   repo.updateCheckRun(checkRun.getId())
      .add(output)
      .withConclusion(GHCheckRun.Conclusion.SUCCESS)
      .withStatus(GHCheckRun.Status.COMPLETED)
      .create();
``` 

Runs without issue in 1.122 to 1.127, but fails during the updateCheckRun call from 1.128 onwards with

```
Caused by: org.kohsuke.github.GHFileNotFoundException: https://api.github.com/repos/XXX/XXX/check-runs/2676567691 {"message":"Not Found","documentation_url":"https://docs.github.com/rest"}
    at org.kohsuke.github.GitHubClient.interpretApiError (GitHubClient.java:482)
    at org.kohsuke.github.GitHubClient.sendRequest (GitHubClient.java:409)
    at org.kohsuke.github.GitHubClient.sendRequest (GitHubClient.java:357)
    at org.kohsuke.github.Requester.fetch (Requester.java:76)
    at org.kohsuke.github.GHCheckRunBuilder.create (GHCheckRunBuilder.java:153)
```

@gsmet has indicated that this represents an authentication failure, however it seems strange that the call to create the checkrun succeeds, while the call to update it fails using the same credentials.

gsmet:
Ah yeah, if the check creation succeeds, then there might be something else going on.

hcoles:
Just had a little dig through the git history, and the change that seems to break things is this

https://github.com/hub4j/github-api/commit/8e20f4d9f5ec8e90a6ce673d47b0049cc4ba26fa

gsmet:
Interesting. That explains why I didn't see it in my GitHub App as I'm using the OkHttp connector.

hcoles:
Yes, looks like I can workaround this with OkHttp.

bitwiseman:
@hcoles 
> It seems that github does not support X-HTTP-Method-Override for this call.

😭🤦 

[...]

PR #1160, opened 2021-05-27
Fix withCredetials() to correctly detect missing creds

### LINKED ISSUE #1155: Problematic regression with ~/.github file authentication
Authenticating with `~/.github` when using `GitHub.connect()` doesn't work anymore in recent versions of the GitHub API.

The client "authenticates" as anonymous instead of using the credentials.

The issue is that the `GitHubBuilder.authorizationProvider` field is never null as it's now initialized to an `ANONYMOUS` instance here: https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/GitHubBuilder.java#L35 .

Thus when arriving here https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/GitHubBuilder.java#L63 , the field is never null and we never get the chance to use the  `~/.github` property file to authenticate a few lines below.

I'm not exactly sure how we want to fix this. A comment in `GitHubBuilder.fromCredentials()` says `@throws IOException If there are no credentials defined in the ~/.github properties file or the process environment.` which is obviously not true anymore.

So:
- we definitely need to test that `authorizationProvider` is not `ANONYMOUS` here: https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/GitHubBuilder.java#L35 .
- I'm not sure if we want to fix the comment and allow anonymous access (but given we explicitly called `fromCredentials()`, I would rather throw an exception) or throw an exception again if we don't have any credentials. The latter would obviously prevent anonymous access.

Note that it can be a bit hard to debug as in my case the GH API was just throwing 404 Not Found errors when trying to do some write operations instead of complaining about the missing credentials.

Also, no idea on how to test this :).

Thoughts?

bitwiseman:
# Description

Fixes #1155

@gsmet @hcoles

If you want to test this fix manually against #1156 or #1157 , that would be helpful. 

# Before submitting a PR:

# When creating a PR: 

### DIFF
--- a/src/test/java/org/kohsuke/github/GitHubConnectionTest.java
+++ b/src/test/java/org/kohsuke/github/GitHubConnectionTest.java
@@ -2,6 +2,7 @@ package org.kohsuke.github;
 
 import org.junit.Assume;
 import org.junit.Test;
+import org.kohsuke.github.authorization.AuthorizationProvider;
 import org.kohsuke.github.authorization.UserAuthorizationProvider;
 
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


src/test/java/org/kohsuke/github/GitHubConnectionTest.java
```diff
@@ -137,6 +138,67 @@ public void testGitHubBuilderFromCustomEnvironment() throws IOException {
...
+        assertThat(builder.authorizationProvider.getEncodedAuthorization(), equalTo("Bearer bogus jwt token string"));
+
+        props.put("password", "bogus weak password");
+        setupEnvironment(props);
+        builder = GitHubBuilder.fromCredentials();
+
+        assertThat(builder.authorizationProvider, instanceOf(UserAuthorizationProvider.class));
+        assertThat(builder.authorizationProvider.getEncodedAuthorization(),
+                equalTo("Basic Ym9ndXMgbG9naW46Ym9ndXMgd2VhayBwYXNzd29yZA=="));
+        assertThat(((UserAuthorizationProvider) builder.authorizationProvider).getLogin(), equalTo("bogus login"));
```
bitwiseman, line 181: TODO: We really should also test for the `~/.github` scenario as well.

hcoles:
@bitwiseman I'm afriad this does not resolve #1157

gsmet (approved):
Perfect, thanks!

gsmet:
Yeah #1157 looks like another issue.

PR #1167, opened 2021-05-28
Add packages, actions and dependabot to meta

shuheiktgw:
# Description
I've added packages, actions, and dependabot to GHMeta class. Thank you for your review! 👍 

# Before submitting a PR:

# When creating a PR: 


bitwiseman:
Could you provide a link to the doc page describing these?

bitwiseman (changes requested):
Thanks for contributing! 

There's no need to change the example classes. Having the extra fields doesn't improve the clarity of the example.  Please revert that part of the change.

shuheiktgw:
@bitwiseman 

> Could you provide a link to the doc page describing these?

Sure, here is the doc page!

https://docs.github.com/en/rest/reference/meta#get-github-meta-information

> There's no need to change the example classes.

OK, I reverted the change related to the example class. Sorry for force-pushing 🙇 (It was much easier simply reset changes on the file)

PR #1175, opened 2021-06-05
Add mapping for from/to fields in issue rename event

t0m4uk1991:
# Description

Issue rename event do not include information about previous(old) name.

According to the [docs](https://docs.github.com/en/developers/webhooks-and-events/events/issue-event-types#renamed)  response  contain rename object with from/to properties for previous/new name. Mappings for this fields were added in scope of this PR.


bitwiseman (approved):
Thank you for submitting a PR.
High quality with tests and comments.

PR #1184, opened 2021-06-14
Remove @Deprecated annotation from Preview and BetaApi elements

### LINKED ISSUE #1003: Remove @Deprecated from @Preview and @BetaApi methods and classes
The `@Deprecated` annotation on `@Preview` and `@BetaApi` methods and class results quite a bit of confusion.  People assume these methods are actually deprecated and going away, when they are not.  Further, there are a growing number of actually deprecated methods in this library this will eventually go away.  We should find a better way to inform users that methods are in preview or beta. 

@marcoferrer 
From your work on #1001 and #1002, perhaps you'd have input here.

Can we use ArchUnit to check that methods with those annotations have some specific text in their JavaDoc titles, for example "(BETA)" or "(PREVIEW)"?  Or perhaps we could write a checkstyle rule?

bitwiseman:
# Description
The use of `@Deprecated` annotations to on `Preview` and `BetaApi` elements is a continuing source of confusion for users.  While it is important to maintain compatibility and communicate the "Preview" state to users, the `@Deprecated` annotation is simply not the right way to do it.  

This change removes all the `@Deprecated` annotations for that scenario. After this change, all methods marked as `@Deprecated` will actually be deprecated and will eventually be removed. 

Fixes #1003

@gsmet @yeikel 

# Before submitting a PR:

# When creating a PR: 


src/main/java/org/kohsuke/github/Preview.java
```diff
@@ -9,9 +9,10 @@
 /**
  * Indicates that the method/class/etc marked maps to GitHub API in the preview period.
  * <p>
- * These APIs are subject to change and not a part of the backward compatibility commitment. Always used in conjunction
- * with 'deprecated' to raise awareness to clients. In addition, it's advised to update the targets documentation to
- * signify that the deprecation is required until preview feature being used is promoted to stable.
+ * These APIs are subject to change and not a part of the backward compatibility commitment.
+ *
+ * It's advised to update the targets documentation with text signify that the preview feature being used is promoted to
+ * stable.
```
timja, line 15: this doesn't make sense to me, I assume you mean that there should be a warning added to the javadoc?

src/test/java/org/kohsuke/github/ArchTests.java
```diff
@@ -68,6 +69,7 @@ public static void beforeClass() {
         assertThat(classFiles.size(), greaterThan(0));
     }
 
+    @Ignore("No longer using deprecated for this purpose")
```
timja, line 72: is there a reason you aren't removing the test?

timja (approved):

src/main/java/org/kohsuke/github/Preview.java
bitwiseman:
```suggestion
 * It is advised to update the target's documentation with text indicating that a preview feature being used.
```

PR #1185, opened 2021-06-20
Fix commit property in GHContentUpdateResponse

takezoe:
# Description

`GHContentUpdateResponse.commit` seems to be different from `GHCommit` according to API reference:
https://docs.github.com/en/rest/reference/repos#create-or-update-file-contents

# Before submitting a PR:

# When creating a PR: 


src/main/java/org/kohsuke/github/GHContentUpdateResponse.java
```diff
@@ -1,11 +1,15 @@
...
+
+import java.net.URL;
+
 /**
  * The response that is returned when updating repository content.
  */
 public class GHContentUpdateResponse {
     private GHContent content;
-    private GHCommit commit;
+    private Commit commit;
```
bitwiseman, line 12: I think this would be pretty close to accurate and save you most of the code in this PR.
```suggestion
    private GHCommit.ShortInfo commit;
```

takezoe: I tried that at first, but found that there are many differences from `GHCommit.ShortInfo`.

For example, `GHCommit.ShortInfo` doesn't have `sha`. Also, it has `comment_count` that doesn't exist in the response of Create / Update Contents API.

bitwiseman: @takezoe
Um, I see a `sha` in there: 


```json
{
  "commit": {
    "sha": "7638417db6d59f3c431d3e1f261cc637155684cd",
    "node_id": "MDY6Q29tbWl0NzYzODQxN2RiNmQ1OWYzYzQzMWQzZTFmMjYxY2M2MzcxNTU2ODRjZA==",
    "url": "https://api.github.com/repos/octocat/Hello-World/git/commits/7638417db6d59f3c431d3e1f261cc637155684cd",
    "html_url": "https://github.com/octocat/Hello-World/git/commit/7638417db6d59f3c431d3e1f261cc637155684cd",
    "author": {
      "date": "2014-11-07T22:01:45Z",
      "name": "Monalisa Octocat",
      "email": "octocat@github.com"
    },
    "committer": {
      "date": "2014-11-07T22:01:45Z",
      "name": "Monalisa Octocat",
      "email": "octocat@github.com"
    },
    "message": "my commit message",
    "tree": {
      "url": "https://api.github.com/repos/octocat/Hello-World/git/trees/691272480426f78a0138979dd3ce63b77f706feb",
      "sha": "691272480426f78a0138979dd3ce63b77f706feb"
    },
    "parents": [
      {
        "url": "https://api.github.com/repos/octocat/Hello-World/git/commits/1acc419d4d6a9ce985db7be48c6349a0475975b5",
        "html_url": "https://github.com/octocat/Hello-World/git/commit/1acc419d4d6a9ce985db7be48c6349a0475975b5",
        "sha": "1acc419d4d6a9ce985db7be48c6349a0475975b5"
      }
    ],
    "verification": {
      "verified": false,
      "reason": "unsigned",
      "signature": null,
      "payload": null
    }
  }
}
```

src/main/java/org/kohsuke/github/GHContentUpdateResponse.java
```diff
@@ -21,7 +25,101 @@ public GHContent getContent() {
      *
      * @return the commit
      */
-    public GHCommit getCommit() {
+    public Commit getCommit() {
```
bitwiseman, line 28: Regardless of other considerations, this is breaking change.  You can work around this with BridgeMethods, but it will still need to be done differently.  (I understand this deeply broken due to lack of testing, to the point that it is virtually impossible that anyone is actually using this API as it stands, but we can't take API breaking change at this time.)

src/main/java/org/kohsuke/github/GHContentUpdateResponse.java
bitwiseman: @takezoe
Wow.  😠  That is annoying. 

At minimum, for `GHCommit`, the [`sha` is outside of the `ShortInfo`/`commit`](https://github.com/hub4j/github-api/blob/81f23e8ae51218b63001c10077a016f07093deae/src/main/java/org/kohsuke/github/GHCommit.java#L260). For `GHContentUpdateResponse`, the `sha` is inside of the `ShortInfo`/`commit`. 

What about having a class that extends `GHCommit.ShortInfo` and adds `sha`? That would allow you to pretty easily use a bridge method to maintain the existing API. 

It looks to me like the current `commit` field will always have an empty record because the mapping is misaligned.  Is that true?  If so, that would mean that the current behavior is completely broken. 

## GHCommit

```json
{
  "url": "https://api.github.com/repos/octocat/Hello-World/commits/6dcb09b5b57875f334f61aebed695e2e4193db5e",
  "sha": "6dcb09b5b57875f334f61aebed695e2e4193db5e",
  "node_id": "MDY6Q29tbWl0NmRjYjA5YjViNTc4NzVmMzM0ZjYxYWViZWQ2OTVlMmU0MTkzZGI1ZQ==",
  "html_url": "https://github.com/octocat/Hello-World/commit/6dcb09b5b57875f334f61aebed695e2e4193db5e",
  "comments_url": "https://api.github.com/repos/octocat/Hello-World/commits/6dcb09b5b57875f334f61aebed695e2e4193db5e/comments",
  "commit": {
    "url": "https://api.github.com/repos/octocat/Hello-World/git/commits/6dcb09b5b57875f334f61aebed695e2e4193db5e",
    "author": {
      "name": "Monalisa Octocat",
      "email": "mona@github.com",
      "date": "2011-04-14T16:00:49Z"
    },
    "committer": {
      "name": "Monalisa Octocat",
      "email": "mona@github.com",
      "date": "2011-04-14T16:00:49Z"
    },
    "message": "Fix all the bugs",
    "tree": {
      "url": "https://api.github.com/repos/octocat/Hello-World/tree/6dcb09b5b57875f334f61aebed695e2e4193db5e",
      "sha": "6dcb09b5b57875f334f61aebed695e2e4193db5e"
    },
    "comment_count": 0,
    "verification": {
      "verified": false,
      "reason": "unsigned",
      "signature": null,
      "payload": null
    }
  },
  "author": {
    "login": "octocat",
    "id": 1,
    "node_id": "MDQ6VXNlcjE=",
    "avatar_url": "https://github.com/images/error/octocat_happy.gif",
    "gravatar_id": "",
    "url": "https://api.github.com/users/octocat",
    "html_url": "https://github.com/octocat",
    "followers_url": "https://api.github.com/users/octocat/followers",
    "following_url": "https://api.github.com/users/octocat/following{/other_user}",
    "gists_url": "https://api.github.com/users/octocat/gists{/gist_id}",
    "starred_url": "https://api.github.com/users/octocat/starred{/owner}{/repo}",
    "subscriptions_url": "https://api.github.com/users/octocat/subscriptions",
    "organizations_url": "https://api.github.com/users/octocat/orgs",
    "repos_url": "https://api.github.com/users/octocat/repos",
    "events_url": "https://api.github.com/users/octocat/events{/privacy}",
    "received_events_url": "https://api.github.com/users/octocat/received_events",
    "type": "User",
    "site_admin": false
  },
  "committer": {
    "login": "octocat",
    "id": 1,
    "node_id": "MDQ6VXNlcjE=",
    "avatar_url": "https://github.com/images/error/octocat_happy.gif",
    "gravatar_id": "",
    "url": "https://api.github.com/users/octocat",
    "html_url": "https://github.com/octocat",
    "followers_url": "https://api.github.com/users/octocat/followers",
    "following_url": "https://api.github.com/users/octocat/following{/other_user}",
    "gists_url": "https://api.github.com/users/octocat/gists{/gist_id}",
    "starred_url": "https://api.github.com/users/octocat/starred{/owner}{/repo}",
    "subscriptions_url": "https://api.github.com/users/octocat/subscriptions",
    "organizations_url": "https://api.github.com/users/octocat/orgs",
    "repos_url": "https://api.github.com/users/octocat/repos",
    "events_url": "https://api.github.com/users/octocat/events{/privacy}",
    "received_events_url": "https://api.github.com/users/octocat/received_events",
    "type": "User",
    "site_admin": false
  },
  "parents": [
    {
      "url": "https://api.github.com/repos/octocat/Hello-World/commits/6dcb09b5b57875f334f61aebed695e2e4193db5e",
      "sha": "6dcb09b5b57875f334f61aebed695e2e4193db5e"
    }
  ],
  "stats": {
    "additions": 104,
    "deletions": 4,
    "total": 108
  },
  "files": [
    {
      "filename": "file1.txt",
      "additions": 10,
      "deletions": 2,
      "changes": 12,
      "status": "modified",
      "raw_url": "https://github.com/octocat/Hello-World/raw/7ca483543807a51b6079e54ac4cc392bc29ae284/file1.txt",
      "blob_url": "https://github.com/octocat/Hello-World/blob/7ca483543807a51b6079e54ac4cc392bc29ae284/file1.txt",
      "patch": "@@ -29,7 +29,7 @@\n....."
    }
  ]
}
```

## GHContentUpdateResponse

```json
{
  "content": {
    "name": "hello.txt",
    "path": "notes/hello.txt",
    "sha": "95b966ae1c166bd92f8ae7d1c313e738c731dfc3",
    "size": 9,
    "url": "https://api.github.com/repos/octocat/Hello-World/contents/notes/hello.txt",
    "html_url": "https://github.com/octocat/Hello-World/blob/master/notes/hello.txt",
    "git_url": "https://api.github.com/repos/octocat/Hello-World/git/blobs/95b966ae1c166bd92f8ae7d1c313e738c731dfc3",
    "download_url": "https://raw.githubusercontent.com/octocat/HelloWorld/master/notes/hello.txt",
    "type": "file",
    "_links": {
      "self": "https://api.github.com/repos/octocat/Hello-World/contents/notes/hello.txt",
      "git": "https://api.github.com/repos/octocat/Hello-World/git/blobs/95b966ae1c166bd92f8ae7d1c313e738c731dfc3",
      "html": "https://github.com/octocat/Hello-World/blob/master/notes/hello.txt"
    }
  },
  "commit": {
    "sha": "7638417db6d59f3c431d3e1f261cc637155684cd",
    "node_id": "MDY6Q29tbWl0NzYzODQxN2RiNmQ1OWYzYzQzMWQzZTFmMjYxY2M2MzcxNTU2ODRjZA==",
    "url": "https://api.github.com/repos/octocat/Hello-World/git/commits/7638417db6d59f3c431d3e1f261cc637155684cd",
    "html_url": "https://github.com/octocat/Hello-World/git/commit/7638417db6d59f3c431d3e1f261cc637155684cd",
    "author": {
      "date": "2014-11-07T22:01:45Z",
      "name": "Monalisa Octocat",
      "email": "octocat@github.com"
    },
    "committer": {
      "date": "2014-11-07T22:01:45Z",
      "name": "Monalisa Octocat",
      "email": "octocat@github.com"
    },
    "message": "my commit message",
    "tree": {
      "url": "https://api.github.com/repos/octocat/Hello-World/git/trees/691272480426f78a0138979dd3ce63b77f706feb",
      "sha": "691272480426f78a0138979dd3ce63b77f706feb"
    },
    "parents": [
      {
        "url": "https://api.github.com/repos/octocat/Hello-World/git/commits/1acc419d4d6a9ce985db7be48c6349a0475975b5",
        "html_url": "https://github.com/octocat/Hello-World/git/commit/1acc419d4d6a9ce985db7be48c6349a0475975b5",
        "sha": "1acc419d4d6a9ce985db7be48c6349a0475975b5"
      }
    ],
    "verification": {
      "verified": false,
      "reason": "unsigned",
      "signature": null,
      "payload": null
    }
  }
}
```

takezoe: > It looks to me like the current commit field will always have an empty record because the mapping is misaligned. Is that true? If so, that would mean that the current behavior is completely broken.

Yes,`commit` field is always null in the current version.

src/main/java/org/kohsuke/github/GHContentUpdateResponse.java
takezoe: Do you mean we should add another method or we should copy data to `GHCommit`?

src/main/java/org/kohsuke/github/GHContentUpdateResponse.java
bitwiseman: > > What about having a class that extends GHCommit.ShortInfo and adds sha? That would allow you to pretty easily use a bridge method to maintain the existing API.
> 
> Don't you mind that `comment_count` becomes null in that case?

Somewhat, but it's not the worst.  You could override `getCommentCount()` to throw an exception. 
 
> > It looks to me like the current commit field will always have an empty record because the mapping is misaligned. Is that true? If so, that would mean that the current behavior is completely broken.
> 
> Yes,`commit` field is always null in the current version.

Wow, that's embarrassing.   And [this has been this way forever](https://github.com/hub4j/github-api/commit/387d4e5bc976f97d3d6c2cbedaa5dba513a9fd70).  😭

src/main/java/org/kohsuke/github/GHContentUpdateResponse.java
bitwiseman: Given that has returns null and [this has been this way forever](https://github.com/hub4j/github-api/commit/387d4e5bc976f97d3d6c2cbedaa5dba513a9fd70), we just need to have a bridge method that returns `null` if someone calls the `GHCommit getCommit()` method.

src/main/java/org/kohsuke/github/GHContentUpdateResponse.java
```diff
@@ -21,7 +25,101 @@ public GHContent getContent() {
...
+        String url;
+
+        /**
+         * Gets sha.
+         *
+         * @return the sha
+         */
+        public String getSha() {
+            return this.sha;
+        }
```
bitwiseman, line 50: This is not committable, but this is what I'm thinking given our discussion so far. 
```suggestion
    public static class Commit extends GHCommit.ShortInfo {
        String sha;

        /**
         * Gets sha.
         *
         * @return the sha
         */
        public String getSha() {
            return this.sha;
        }
        
        
        /**
         * Gets comment count.
         *
         * @return the comment count
         */
        @Override
        public int getCommentCount() {
            throw GHException("Not available on this endpoint.");
        }
```

And then you can delete the rest.

takezoe: Hmm... `tree` is not accessible in `GHCommit.ShortInfo`.

takezoe: Hmm... `tree` in `GHCommit.ShortInfo` is not accessible from a subclass. I think reusing `GHCommit.ShortInfo` may not be a good idea.

src/main/java/org/kohsuke/github/GHContentUpdateResponse.java
takezoe: I tried to add a bridge method for `GHCommit`. Is this correct?

src/main/java/org/kohsuke/github/GHContentUpdateResponse.java
bitwiseman: @takezoe 
It is okay to change internals or to add methods if you believe they make sense.  

You could change `tree` to be `protected` so that it could be accessed here. You'll note though that `ShortInfo.Tree` is not `public`.  Instead, `GHCommit` uses it to request a `GHTree` if the user asks for it.  

I should say here that you've repeated indicated that you don't believe extending like this is the right path.  I'm open to alternatives, but I also don't want to just pile more classes on because GitHub didn't make their API consistent. 

Maybe there should be a common base class for these two classes?  Maybe a `GitCommit` similar to `GitUser`?

takezoe: > You could change tree to be protected so that it could be accessed here. You'll note though that ShortInfo.Tree is not public. Instead, GHCommit uses it to request a GHTree if the user asks for it.

`ShortInfo.Tree` has only `sha` because it seems to be designed to use to create `GHCommit` by an extra API call when asked. But in some other classes, there are `Tree` that have `sha` and `url` and be accessible from users directly.

https://github.com/hub4j/github-api/blob/cf2571858c429967764fdb4e9cd38aae6a18bbe1/src/main/java/org/kohsuke/github/GHCompare.java#L250-L251

https://github.com/hub4j/github-api/blob/cf2571858c429967764fdb4e9cd38aae6a18bbe1/src/main/java/org/kohsuke/github/GHPullRequestCommitDetail.java#L59-L61

I don't know which design is better. If we only need `tree.sha` or `tree.url`, we don't need to make such an extra API call to create `GHTree`. Personally, I prefer this because it allows us to avoid an implicit extra API call.

> I also don't want to just pile more classes

Hmm... If that's a philosophy of this library, why there are duplicated `Tree`s in some classes? Why there are classes that return `Tree` and `GHTree`? It's hard to find the consistency in the existing codebase of this library and what you are saying.

> because GitHub didn't make their API consistent.

Does this mean GitHub modifies API frequently so you want to minimize necessary future effort to follow the changes? Yeah, I totally agree with it!

> Maybe there should be a common base class for these two classes? Maybe a GitCommit similar to GitUser?

On the other hand, looking at use of bridge methods, this library seems to emphasize the backward compatibility, meaning we cannot delete unnecessary properties easily once added. In that case, having a base class could increase the chance of sparse properties (like `commit_count`).

takezoe: To be honest, I cannot find what is the best way to fix this issue from the existing codebase and what you are saying probably because of the lack of my understanding and knowledge about this library. So, can I ask you to fix this issue in a way you want? Obviously, this issue must be fixed, but I don't want to spend more time for this.

bitwiseman: @takezoe 
It is unfortunate that you have no more time.  This project is many years old and has grown organically along with the GitHub REST API.  I understand your frustration.  You figured you'd just drop in fix an obviously broken feature and I've made it much more involved.   Sorry about that.

I agree that all the things you mention above are real problems and I sympathize with your frustration. However, all the problems you mention only highlight the need for cleanup and improvement in this area.  After a bit of digging in the docs, I found an endpoint that [gets a Git Commit](https://docs.github.com/en/rest/reference/git#get-a-commit). So, these various other classes should probably all be inheriting from a single base called `GitCommit` which should be returned by `GHRepository.getGitCommit()`.

I guess I'll file an issue that points the this PR and hopefully someone will have time to fix this eventually.

takezoe: No worries. You a maintainer of this library. Your thought should be respected. I just don't have a time to do try-and-error to find the best way for this.

> However, all the problems you mention only highlight the need for cleanup and improvement in this area.

That may be true. For `GHCommit`, I think it should provide a way to get raw `tree` information without an extra API call. Why we need to make an extra API call when we only need `tree.sha`?

bitwiseman: > > However, all the problems you mention only highlight the need for cleanup and improvement in this area.
> 
> That may be true. For `GHCommit`, I think it should provide a way to get raw `tree` information without an extra API call. Why we need to make an extra API call when we only need `tree.sha`?

Less than optimal design.  There should definitely be a way to do what you are pointing to.  Please feel free to file issues/PRs to address this and any other of the points you mentioned above.

bitwiseman:
@takezoe 
On a related note, it turns out that `getCommit()` doesn't return `null`, it just returns a very sparse `GHCommit`.  I've filed #1190, that verifies the current behavior and makes minor tweaks to populate  the few unguarded fields.  This is still not great, but it is also not complete broken.  The library makes an additional call to populate the missing info and continues on. 

I've also filed issue #1191 to track creating a `GitCommit` class.

takezoe:
> it turns out that getCommit() doesn't return null, it just returns a very sparse GHCommit.

Oh, I see. Then this pull request is invalid anyway, particularly a bridge method that returns null is wrong. I will close this.

bitwiseman:
Thanks for raising this PR.  Please consider contributing further in the future.  I promise most PRs do not drag out the way this one did.

PR #1187, opened 2021-06-21
Add getAppInstallations(/user/installations) method on GHMyself

t0m4uk1991:
# Description

Add  getAppInstallations (under the hood call on /user/installations) method in GHMyself

For more info pls take a look at https://docs.github.com/en/rest/reference/apps#list-app-installations-accessible-to-the-user-access-token


src/main/java/org/kohsuke/github/GHAppInstallations.java
```diff
@@ -0,0 +1,21 @@
+package org.kohsuke.github;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+
+public class GHAppInstallations {
+
+    @JsonProperty("total_count")
+    private int totalCount;
```
bitwiseman, line 10: This is done automatically by Jackson. 
```suggestion
    private int totalCount;
```

bitwiseman (changes requested):
You've tested this for 1 installation and it works fine, but ...

This is actually a paged response.  Similar to search and a number of other endpoints
https://docs.github.com/en/rest/reference/apps#list-app-installations-accessible-to-the-user-access-token--parameters
😭 

You'll need to implement this using a class extending `PagedIterable` in about the same way   https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/PagedSearchIterable.java does.

t0m4uk1991:
Hi @bitwiseman 

Thanks for quick feedback.
I updated PR, could you please take a look again?

src/main/java/org/kohsuke/github/GHAppInstallationsPage.java
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Represents the one page of GHAppInstallations.
+ */
+class GHAppInstallationsPage {
+    private int total_count;
+    private GHAppInstallation[] installations;
+
+    public int getTotalCount() {
+        return total_count;
```
bitwiseman, line 11: Test this.

src/main/java/org/kohsuke/github/GHAppInstallation.java
```diff
@@ -365,4 +365,9 @@ public GHAppCreateTokenBuilder createToken(Map<String, GHPermissionType> permiss
     public GHAppCreateTokenBuilder createToken() {
         return new GHAppCreateTokenBuilder(root, String.format("/app/installations/%d/access_tokens", getId()));
     }
+
+    GHAppInstallation wrap(GitHub root) {
+        this.root = root;
+        return this;
+    }
```
bitwiseman, line 372: Don't think you need this now. `root` gets bound by Jackson automatically.

src/main/java/org/kohsuke/github/GHAppInstallationsPage.java
```diff
@@ -0,0 +1,20 @@
...
+    }
+
+    GHAppInstallation[] getInstallations(GitHub root) {
+        for (GHAppInstallation installation : installations) {
+            installation.wrap(root);
+        }
+        return installations;
+    }
```
bitwiseman, line 19:
```suggestion
    GHAppInstallation[] getInstallations() {
        return installations;
    }
```

src/main/java/org/kohsuke/github/GHAppInstallation.java
bitwiseman, line 372:
```suggestion
```

I don't think you need this now. `root` gets bound by Jackson automatically. 

https://github.com/hub4j/github-api/blob/81f23e8ae51218b63001c10077a016f07093deae/src/main/java/org/kohsuke/github/GitHubInteractiveObject.java#L13-L14

src/main/java/org/kohsuke/github/GHAppInstallationsIterable.java
```diff
@@ -0,0 +1,48 @@
...
+            public boolean hasNext() {
+                return base.hasNext();
+            }
+
+            public GHAppInstallation[] next() {
+                GHAppInstallationsPage v = base.next();
+                if (result == null) {
+                    result = v;
+                }
+                return v.getInstallations(root);
```
bitwiseman, line 44:
```suggestion
                return v.getInstallations();
```

src/main/java/org/kohsuke/github/GHAppInstallation.java
bitwiseman: Also if needed there's already a `wrapUp()` on line 313.

bitwiseman (approved):

PR #1188, opened 2021-06-24
Add ability to search repos by fork and visibility

### LINKED ISSUE #1174: Improvement: Include Forked repositories in GHRepositorySearchBuilder
As a user of your library,
I want to use the GHRepositorySearchBuilder class to find my forked repositories
so that I'm able to automate fetch upstream.

#### Current implementation

The GHRepositorySearchBuilder has a **forks** method. But this is for searching for how many time a repository was forked and not if the repository itself is a fork.

#### Expected implementation

Add a fork method to GHRepositorySearchBuilder so that t can add the parameters true or only to search for repositories that I have forked.

#### References

- https://docs.github.com/en/github/searching-for-information-on-github/searching-on-github/searching-for-repositories

t0m4uk1991:
# Description

Updated GHRepositorySearchBuilder.java, add ability to search repos by fork:(true|only) and visibility:(public|private|internal)

This one issue https://github.com/hub4j/github-api/issues/1174 should be closed after merge.

Fixes #1174


src/main/java/org/kohsuke/github/GHRepositorySearchBuilder.java
```diff
@@ -160,6 +196,33 @@ public GHRepositorySearchBuilder sort(Sort sort) {
...
+     * The enum Visibility.
+     */
+    public enum Visibility {
+        PUBLIC, INTERNAL, PRIVATE;
+
+        public String toString() {
+            return name().toLowerCase(Locale.ENGLISH);
+        }
+    }
+
```
bitwiseman, line 225: Use `GHRepository.Visibility` instead? 

https://github.com/hub4j/github-api/blob/9978678f276619e2bf81e6542a596e74ce1dc0cc/src/main/java/org/kohsuke/github/GHRepository.java#L729-L740

src/main/java/org/kohsuke/github/GHRepositorySearchBuilder.java
```diff
@@ -52,6 +54,40 @@ public GHRepositorySearchBuilder forks(String v) {
...
+     *
+     * @param fork
+     *            search mode for forks
+     *
+     * @return the gh repository search builder
+     *
+     * @see <a href=
+     *      "https://docs.github.com/en/github/searching-for-information-on-github/searching-on-github/searching-in-forks">Searching
+     *      in forks</a>
+     *
```
bitwiseman, line 68: Hm, so the default (if this value isn't supplied is to not search forks, right?  Should that be a value in the enum?  `Forks.DEFAULT` which removes that filter if present?  

Also, you probable want to be explicit about the limitations, both here and in the JavaDoc for the enum: 

> To show forks with more stars than the parent repository in code search results, add `fork:true` or `fork:only` to your query.

bitwiseman (changes requested):
Thanks!  Minor tweaks and questions.  I'm open to discussion if you disagree with my suggestions.

src/main/java/org/kohsuke/github/GHRepositorySearchBuilder.java
t0m4uk1991: @bitwiseman  yeah, I agree, we do not need duplications.
What is expected behavior if we use GHRepository.Visibility.UNKNOWN? Should we just ignore this value in GHRepositorySearchBuilder?

src/main/java/org/kohsuke/github/GHRepositorySearchBuilder.java
t0m4uk1991: @bitwiseman yep, according to docs we ignore forks by default.
Do we have a use case when we need Forks.DEFAULT?

src/main/java/org/kohsuke/github/GHRepositorySearchBuilder.java
bitwiseman: @t0m4uk1991
Doh, `UNKNOWN` should have JavaDoc explaining usage.   

If passed, it would be reasonable to ignore or maybe even throw an exception with a message saying "UKNOWN cannot be passed into this method."

src/main/java/org/kohsuke/github/GHRepositorySearchBuilder.java
bitwiseman: Someone might want to provide a default builder that sets `fork:true` and then change it to default in only a few places in their code.  I see you've made the change.  Thanks!

src/main/java/org/kohsuke/github/GHRepositorySearchBuilder.java
t0m4uk1991: in the current implementation, I just ignore this param if it equals to GHRepository.Visibility.UNKNOWN, also I added some explanation in javadocs

t0m4uk1991:
Hi @bitwiseman 
Could you please take a look at the changes again?
Thanks

bitwiseman:
@t0m4uk1991
In the spirit of collaboration, I pushed https://github.com/hub4j/github-api/pull/1188/commits/d7a19aaeabe082c09a53dcbaa6b3de012e9dc576 making some additional changes. What do you think? 

If you disagree with the changes, please feel free to modify or revert them and we can discuss further.

bitwiseman (approved):

t0m4uk1991:
@bitwiseman 
LGFM. The more documents/explanations we have, the better.

PR #1201, opened 2021-07-30
WIP Add test for searchIssue method

t0m4uk1991:
# Description

Add test for searchIssue method (case for search in  repo)


bitwiseman:
Code coverage shows this really is already covered by #709.  I'm going to pass on adding this.  If you want to update the existing test with a comment about what it is doing and why, that would be fine but not required.

PR #1214, opened 2021-08-09
1209 GHCompare supports per_page and page params

### LINKED ISSUE #1209: GHCompare - support parameters "per_page" and "page"
Hello, does it in the roadmaps to support those parameters?
Github changed the Compare two Commits API, so now it is possible to get over 250 commits:
https://github.blog/changelog/2021-03-22-compare-rest-api-now-supports-pagination/

Best regards, Max

ravingbool:
GHCompare supports per page and page+params

Fixes #1209


ravingbool:
@bitwiseman please run the checks again, I pushed the test

src/test/java/org/kohsuke/github/GHRepositoryTest.java
```diff
@@ -866,4 +866,11 @@ public void getLastCommitStatus() throws Exception {
         assertThat(status.getState(), equalTo(GHCommitState.SUCCESS));
         assertThat(status.getContext(), equalTo("ci/circleci: build"));
     }
+
+    @Test
+    public void getCommitsBetween() throws Exception {
+        GHCompare compare = getRepository().getCompare("e46a9f3f2ac55db96de3c5c4706f2813b3a96465", "8051615eff597f4e49f4f47625e6fc2b49f26bfc", 10, 1);
```
bitwiseman, line 872:
```suggestion
        GHCompare·compare·=·getRepository().getCompare("e46a9f3f2ac55db96de3c5c4706f2813b3a96465",
                "8051615eff597f4e49f4f47625e6fc2b49f26bfc",
                10,
                1);        
```

src/test/java/org/kohsuke/github/GHRepositoryTest.java
```diff
@@ -866,4 +866,14 @@ public void getLastCommitStatus() throws Exception {
         assertThat(status.getState(), equalTo(GHCommitState.SUCCESS));
         assertThat(status.getContext(), equalTo("ci/circleci: build"));
     }
+
+    @Test
+    public void getCommitsBetween() throws Exception {
+        GHCompare compare = getRepository().getCompare("e46a9f3f2ac55db96de3c5c4706f2813b3a96465",
+                "8051615eff597f4e49f4f47625e6fc2b49f26bfc",
+                10,
+                1);
```
bitwiseman, line 875:
```suggestion
                1);
```

ravingbool:
@bitwiseman thank you for changing the test, please push the pull request, seems like the checks are ok

ravingbool:
It is all ok with MR?

bitwiseman (changes requested):
@ravingbool 
Sorry for the delay, I'm currently on vacation.  

This is a harder problem to solve right than it might appear at first.  What you have done is okay and exposes this new feature of the endpoint, but it does so in a way that is inconsistent with the rest of the APIs in this library.

All the other endpoints that support `per_page` and `page` parameters use `PagedIterable` to iterate over items.  However, those endpoints also for the most part return simple lists.  There are also some endpoints that return pages as relatively sparse objects that then contain lists. 

For example `GHRepository.getCheckRuns` returns `PagedIterable<GHCheckRun>`:

https://github.com/hub4j/github-api/blob/4247c8a60d809994836d67fe6ce3ad4d5d115406/src/main/java/org/kohsuke/github/GHRepository.java#L2027-L2033

This actually an instance of [GHCheckRunsIterable](https://github.com/hub4j/github-api/blob/4247c8a60d809994836d67fe6ce3ad4d5d115406/src/main/java/org/kohsuke/github/GHCheckRunsIterable.java#L10) which pages through [GHCheckRunsPage](https://github.com/hub4j/github-api/blob/4247c8a60d809994836d67fe6ce3ad4d5d115406/src/main/java/org/kohsuke/github/GHCheckRunsPage.java#L6) objects.

I think the right way to address this will be to add `GHCompare.listCommits()` method that returns a `PagedIterable<Commit>` and internally implements something similar `GHCheckRunsIterable` that uses `GHCompare` objects as the pages. 

Then change your new method to be `GHRepository.getCompare(String id1, String id2, int default_commits_per_page)`.  The query for that would set `per_page` to 1 and set a new `default_commits_per_page` field on the `GHCompare` instance.  This would result in only one commit being returned as part of the that query, minimizing extra data transmission. 
If `default_commits_per_page` field is set, `GHCompare.listCommits()` would have it's `per_page` set to the that by default and `GHCompare.getCommits()` would call `listCommits().getArray()` instead of it's current behavior.

You'll also need a few new tests for this behavior. Specifically, at least one test that has two pages of results. 

Does this make sense? Do you have time to do this additional work?

ravingbool:
@bitwiseman well, I didn't suppose that it would so a lot of work. Can we merge it as is now and make a follow up for it? Because I would like to have a new version with this functionality as soon as possible. I will refactor it in the future, but now I would like to have it as is, pleeeease @bitwiseman

bitwiseman:
@ravingbool 
Sorry, adding methods that we intend to remove "as soon as possible" is counter to policy for this library. However, I recognize that you've worked on this in good faith and need the functionality badly.  I've pushed a commit that does basically what I described with minor changes. I've modified the test to show how the modified functionality works.  What do you think?

ravingbool:
@bitwiseman 
Thank you very much, it is great!

bitwiseman (approved):

PR #1230, opened 2021-09-09
Streamline GitHub root handling

bitwiseman:
# Description

Streamline GitHub root handling.  
Making this work required change to `GHMyself` and how `GitHub` and `GitHubClient` get login strings. 
I also added tests to cover a number of missing code paths to continue to bring coverage numbers up.


src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -517,7 +517,6 @@ public GHMyself getMyself() throws IOException {
 
     private void setMyself(GHMyself myself) {
         synchronized (this) {
-            myself.wrapUp(this);
             this.myself = myself;
         }
     }
```
bitwiseman, line 522: I feel like this won't work, but might not be properly covered.

PR #1239, opened 2021-09-24
Feature/#1025

### LINKED ISSUE #1025: Add GHIssueQueryBuilder to allow better filtering of issue queries
**Describe the feature**
GitHub API  enables multiple search criteria: https://docs.github.com/en/rest/reference/issues#list-repository-issues (e.g. assignee, creator, labels). 

How, I can't see how one can take advantage of these parameters through the library API?

**Additional context**
I might have crafted manually a call similar to 

```
    /**
     * Lists up all the issues in this repository.
     *
     * @param state
     *            the state
     * @return the paged iterable
     */
    public PagedIterable<GHIssue> listIssues(final GHIssueState state) {
        return root.createRequest()
                .with("state", state)
                .withUrlPath(getApiTailUrl("issues"))
                .toIterable(GHIssue[].class, item -> item.wrap(this));
    }
```

but it seems not trivial to get a reference to the root variable inside GHRepository.

Thanks

bloslo:
# Description

Add `GHIssueQueryBuilder`. Closes #1025.

I didn't use the `GHIssueQueryBuilder` as a base class for the `GHPullRequestQueryBuilder` as suggested in the issue because of the different query parameters the two endpoints list.

I was wondering should the `listIssues` method in `GHRepository` be marked for deprecation? Similar to the `listPullRequests` method in the same class.


bitwiseman:
@bloslo 
 Please write your tests so that they do not reduce existing code coverage (similar to what I did with your previous PR).  
I've invited you to the `hub4j-test-org` so you can re-record test data as needed.

bitwiseman:
> I was wondering should the `listIssues` method in `GHRepository` be marked for deprecation? Similar to the `listPullRequests` method in the same class.

Hm, those endpoints are still valid in the REST API, so deprecating them is probably to strong an action.  Maybe having the JavaDoc note that the query endpoint offers finer-grained filtering and sorting of results.

bitwiseman:
> I was wondering should the `listIssues` method in `GHRepository` be marked for deprecation? Similar to the `listPullRequests` method in the same class.

Yeah, probably.  

I'm also wondering if `GHQueryBuilder` should extend `PagedIterable`.  That would remove the need to call `list()` or end up with `query.list().asList()`.   But that is another issue/PR.

src/main/java/org/kohsuke/github/GHIssueQueryBuilder.java
```diff
@@ -0,0 +1,163 @@
...
+     * Labels gh issue query builder.
+     *
+     * @param label
+     *            the labels
+     * @return the gh issue query builder
+     */
+    public GHIssueQueryBuilder label(String label) {
+        if (label != null) {
+            labels.add(label);
+        }
```
bitwiseman, line 87: No super efficient, but simplifies the `list()` method. 
```suggestion
        if (label != null && label.trim() != "") {
            labels.add(label);
            req.put("labels", labels.stream().collect(Collectors.joining(",")));
        }
```

src/main/java/org/kohsuke/github/GHIssueQueryBuilder.java
```diff
@@ -0,0 +1,163 @@
...
+import java.util.stream.Collectors;
+
+public class GHIssueQueryBuilder extends GHQueryBuilder<GHIssue> {
+    private final GHRepository repo;
+    private List<String> labels = new ArrayList<>();
+
+    GHIssueQueryBuilder(GHRepository repo) {
+        super(repo.root());
+        this.repo = repo;
+    }
```
bitwiseman, line 15: This query builder is specifically querying [issues in a repository](https://docs.github.com/en/rest/reference/issues#list-repository-issues). 

It is also possible to query [issues related to the current user](https://docs.github.com/en/rest/reference/issues#list-issues-assigned-to-the-authenticated-user) or [issues in an organization related to the current user)](https://docs.github.com/en/rest/reference/issues#list-organization-issues-assigned-to-the-authenticated-user).  Even more fun, each of those endpoints supports a different set of query parameters. 😭 

The name of this class needs to reflect that this is the query for a repository.  I think I'd suggest the following change (or something like it - if you have some other design you feel is better, I'm open to it). 

This isn't needed for pull requests, because they can only be queried [in relation to a repo](https://docs.github.com/en/rest/reference/pulls#list-pull-requests).

```suggestion
public abstract class GHIssueQueryBuilder extends GHQueryBuilder<GHIssue> {
    private final List<String> labels = new ArrayList<>();

    GHIssueQueryBuilder(GitHub root) {
        super(root);
    }

    // common query parameters such as "labels", "since", "state", "direction", etc., go in the parent class

    public static class ForRepository extends GHIssueQueryBuilder {
        private final GHRepository repo;

        public ForRepository(GHRepository repo) {
            super(repo.root());
            this.repo = repo;
        }
        
        @Override
        protected abstract String getApiUrl() {
            return repo.getApiTailUrl("issues");
        }
        // repo-specific parameters such as "milestone" and "assignee" go inside this class
    }
        
    // other query builders can be implemented later
   
}
```

src/test/java/org/kohsuke/github/AppTest.java
```diff
@@ -291,6 +291,42 @@ public void testGetIssues() throws Exception {
         String readRepoString = GitHub.getMappingObjectWriter().writeValueAsString(closedIssues.get(0));
     }
 
+    @Test
+    public void testQueryIssues() throws IOException {
+        List<GHIssue> openBugIssues = gitHub.getOrganization("hub4j-test-org")
+                .getRepository("testQueryIssues")
```
bitwiseman, line 297:
```suggestion
        final GHRepository repo = gitHub.getOrganization("hub4j-test-org")
                .getRepository("testQueryIssues");
        List<GHIssue> openBugIssues = repo
```

src/test/java/org/kohsuke/github/AppTest.java
```diff
@@ -291,6 +291,42 @@ public void testGetIssues() throws Exception {
...
+                .queryIssues()
+                .state(GHIssueState.OPEN)
+                .label("bug")
+                .creator(gitHub.getMyself().getLogin())
+                .milestone("1")
+                .pageSize(10)
+                .list()
+                .toList();
+        assertThat(openBugIssues, is(not(empty())));
```
bitwiseman, line 306: Please make a few more meaningful assertions for each of the queries.  At least the number of issues and maybe a couple spot checks for issue values.

src/test/java/org/kohsuke/github/AppTest.java
```diff
@@ -291,6 +291,42 @@ public void testGetIssues() throws Exception {
         String readRepoString = GitHub.getMappingObjectWriter().writeValueAsString(closedIssues.get(0));
     }
 
+    @Test
+    public void testQueryIssues() throws IOException {
```
bitwiseman, line 295: These queries offer a pretty good range of coverage.  
You have one without a call to `label()` and one with a single call, but we also need one with two calls to label
We also need passing `label(null)` and `label("")`.  It's a pain but there's at least that many possible scenarios that need to be checked.

src/test/java/org/kohsuke/github/AppTest.java
```diff
@@ -291,6 +291,42 @@ public void testGetIssues() throws Exception {
...
+                .label("bug")
+                .creator(gitHub.getMyself().getLogin())
+                .milestone("1")
+                .pageSize(10)
+                .list()
+                .toList();
+        assertThat(openBugIssues, is(not(empty())));
+
+        List<GHIssue> openIssuesWithAssignee = gitHub.getOrganization("hub4j-test-org")
+                .getRepository("testQueryIssues")
```
bitwiseman, line 309:
```suggestion
        List<GHIssue> openIssuesWithAssignee = repo
```

src/test/java/org/kohsuke/github/AppTest.java
```diff
@@ -291,6 +291,42 @@ public void testGetIssues() throws Exception {
...
+                .queryIssues()
+                .state(GHIssueState.OPEN)
+                .assignee(gitHub.getMyself().getLogin())
+                .list()
+                .toList();
+        assertThat(openIssuesWithAssignee, is(not(empty())));
+
+        List<GHIssue> allIssuesSince = gitHub.getOrganization("hub4j-test-org")
+                .getRepository("testQueryIssues")
```
bitwiseman, line 318:
```suggestion
        List<GHIssue> allIssuesSince = repo
```

bitwiseman (changes requested):
Sorry for the second round of feedback.  I took a closer look this time around and saw a couple things I missed.  The `label` change is pretty much as written in the suggestions.  The class structure change, the suggested code is for clarity and there are a bunch more changes to be made.  

Thanks for you patience and perseverance.

bitwiseman (approved):
Excellent!

PR #1240, opened 2021-09-24
Added `list-workflow-runs` API

serkan-ozal:
# Description

Add `list-workflow-runs` API support (https://docs.github.com/en/rest/reference/actions#list-workflow-runs) to the `GHWorkflow`.


src/main/java/org/kohsuke/github/GHWorkflowRunsIterable.java
```diff
@@ -18,6 +19,15 @@ public GHWorkflowRunsIterable(GHRepository owner, GitHubRequest request) {
         this.request = request;
     }
 
+    public GHWorkflowRunsIterable(GHRepository owner, GitHubRequest.Builder<?> requestBuilder) {
```
bitwiseman, line 22: Please make [this method](https://github.com/hub4j/github-api/blob/bcb71a3c318d243c92b757310cf153b8afc983d5/src/main/java/org/kohsuke/github/GHWorkflowRunQueryBuilder.java#L94-L99) call this constructor and remove the other constructor.

bitwiseman (changes requested):
Minor change requested. 

Other than that looks great!

src/main/java/org/kohsuke/github/GHWorkflowRunsIterable.java
serkan-ozal: Done :+1:

serkan-ozal:
Hi @bitwiseman, when new version (`1.134`) will be released?

PR #1242, opened 2021-09-28
Using TCK Tested JDK builds of OpenJDK

carldea:
The AdoptOpenJDK has been discontinued since July 2021. When using Zulu 
you get all the latest updated (TCK Tested) builds for all versions of OpenJDK. 

As good practice I added major fixed versions of JDK 8.0.192, 11.0.3. 

An example is when 8.0.192 passes(green) and 8 (latest) fails (red) you know it was the build version not your code that caused the build to fail.

While Temurin is the successor to AdoptOpenJDK it does not support archived versions prior to sept. 2021. Azul has all archived versions and the latest releases.

# Description

# Before submitting a PR:

# When creating a PR: 

### DIFF
--- a/.github/workflows/maven-build.yml
+++ b/.github/workflows/maven-build.yml
@@ -13,14 +13,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [ 16 ]
+        java: [ 16, 17 ]
     steps:
       - uses: actions/checkout@v2
       - name: Set up JDK
         uses: actions/setup-java@v2
         with:
           java-version: ${{ matrix.java }}
-          distribution: 'adopt'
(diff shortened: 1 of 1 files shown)


carldea:
I have to look as to why the fixed version fails in testing.

carldea:
The workflow is all green on my branch 👍🏼. 

I noticed the logic in the workflow on when to use the `--add-opens` option.
-  Do not apply when using JDK `8` or `8.0.192`
-  Do apply when using JDK 11 and greater

The logic will now avoid applying the JVM option to JDK 8 versions (even fixed versions).

carldea:
Thank. Glad to have help contribute!

bitwiseman (approved):

Issue #1262, opened 2021-10-04
ObsoleteUrlFactory overrides OkHttpClient’s interceptors


Typraeurion:
**Describe the bug**
I’ve been troubleshooting why my custom network interceptor that I added to the OkHttpClient was never called.  After tracing through the code, I found that ObsoleteUrlFactory.buildCall is explicitly _removing_ all interceptors set on the OkHttpClient and adding its own:
```
            OkHttpClient.Builder clientBuilder = client.newBuilder();
            clientBuilder.interceptors().clear();
            clientBuilder.interceptors().add(UnexpectedException.INTERCEPTOR);
            clientBuilder.networkInterceptors().clear();
            clientBuilder.networkInterceptors().add(networkInterceptor);
```

**To Reproduce**
```
    public class TimingInterceptor implements Interceptor {

        @Override
        public Response intercept(Chain chain) throws IOException {
            System.out.println("Intercepting network call from OkHttp");
        }

    }

    public GHUser getCurrentUser(String authToken)
            throws ForbiddenException, IOException {

        OkHttpClient.Builder clientBuilder = new OkHttpClient.Builder();
        clientBuilder.addNetworkInterceptor(new TimingInterceptor());
        OkHttpConnector httpConnector = new OkHttpConnector(clientBuilder.build());
        GitHub gitHub = new GitHubBuilder()
                .withConnector(httpConnector)
                .withOAuthToken(authToken)
                .build();
        GHMyself me = gitHub.getMyself();

        return me;

    }
```

**Expected behavior**
Application displays “Intercepting network call from OkHttp” when `getCurrentUser` is called.

**Desktop (please complete the following information):**
 - Java: 1.8.0
 - github-api: 1.333
 - okhttp: 4.9.2 (previously tried 3.14.9, with the same result)

bitwiseman:
Generally speaking, this is a known limitation of `ObsoleteUrlFactory`. This is another reason we need to implement #1085 so we can stop using the deprecated/obsolete code path.

bitwiseman:
@Typraeurion  
We won't be fixing this in `ObsoleteUrlFactory`, in the next release (code already merged to `main`) we've added `OkHttpGitHubConnector` which does not depend on `ObsoleteUrlFactory` allowing for use of interceptors.

Issue #754, opened 2020-03-25
Illegal Reflective Access Operation to field java.net.HttpURLConnection.method


andrewlalis:
**Describe the bug**
I attempt to archive a repository using the following code:
```java
GHRepository repo = this.org.getRepository("repo-name-here");
repo.archive();
```
When checking GitHub itself, the repository is in fact archived, however I receive a warning in the stderr stream consisting of the following:
```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.kohsuke.github.GitHubHttpUrlConnectionClient$HttpURLConnectionResponseInfo (file:/C:/Users/andrew/.m2/repository/org/kohsuke/github-api/1.108/github-api-1.108.jar) to field java.net.HttpURLConnection.method
WARNING: Please consider reporting this to the maintainers of org.kohsuke.github.GitHubHttpUrlConnectionClient$HttpURLConnectionResponseInfo
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

**To Reproduce**
1. Fetch a `GHRepository` that is not yet archived.
2. Call `archive()` on that repository.

**Expected behavior**
I expect that there should not be any illegal reflective access exceptions.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Firefox
 - Version 74

bitwiseman:
This is not an exception, as you noted your action succeeded. 

If you use the JDK version of `HTTPUrlConnection`, it only allows a specific set of verbs, so we must use reflection to override that. 
https://github.com/github-api/github-api/blob/4c30f94355dfca7034f6619f64543ee3e9fcafdf/src/main/java/org/kohsuke/github/GitHubHttpUrlConnectionClient.java#L162-L191

If you switch to another implementation such as OkHttp3, this should go away. 

"All illegal access operations will be denied in a future release".

lotharschulz:
I experience the same behavior using kotlin with 

```kotlin
repository.setDefaultBranch("some-branch")
```

---

> "All illegal access operations will be denied in a future release", but as of Java 13 that has not happened yet.

I experience the same with Java 14 ( jvm target set in _build.gradle.kts_: `kotlinOptions.jvmTarget = "14"`)

bitwiseman:
@lotharschulz 
There is nothing to be done about this as long as this library uses `HttpURLConnection`. 
Have you tried using OkHttp3?

[...]

PR #1287, opened 2021-11-04
Introduce `ResponseConnector` as a superset of `HttpConnector`

### LINKED ISSUE #1085: Deprecate all usages of HttpConnector and HttpURLConnection
[HttpConnector](https://github.com/hub4j/github-api/blob/master/src/main/java/org/kohsuke/github/HttpConnector.java) depends on [HttpURLConnection](https://docs.oracle.com/javase/8/docs/api/java/net/HttpURLConnection.html) an old and very crufty way of working with Http connections.  Some modern clients have even stopped supporting HttpURLConnection. This makes it harder to implement support for different http clients. 

We should remove all dependency on `HttpConnector`.   `GithubClient` will become the new base class to extend in order to support new clients.  Other methods the depend on `HttpURLConnection` should switch to other ways of providing the same information.

nedtwigg:
This PR touches a lot of files, but it's pretty straightforward. The commits tell the story as simply as I can tell it, every commit in the chain passes `mvn test`. It resolves #1085 in a backward-compatible way (along with #1262 and #754).

It introduces a new interface `ResponseConnector`, with just a single method which functions as a delegate to  the `getResponseInfo` method of `GitHubClient` which allows `GitHubClient` to be become final.

https://github.com/hub4j/github-api/blob/7a91e436c6493cc607922c849cc8032284282de8/src/main/java/org/kohsuke/github/GitHubClient.java#L403-L406

`GitHubHttpUrlConnectionClient` now implements `ResponseConnector` rather than `GitHubClient`. This means that all of the old `HttpConnector` stuff can keep working transparently:

https://github.com/hub4j/github-api/blob/7a91e436c6493cc607922c849cc8032284282de8/src/main/java/org/kohsuke/github/GitHubClient.java#L137-L143

In order to make it possible to implement `ResponseConnector`, I had to extract `ResponseInfo` into its own class and make it and some of its members public. Most of the commits in this PR are pure refactors (refactor commits aren't mixed in with new or changed functionality) so you can see exactly which parts had to become public.

The whole point of the PR is to bypass `HttpConnector` and users can now swap the new [`OkHttpResponseConnector`](https://github.com/nedtwigg/github-api/blob/remove-http-connector/src/main/java/org/kohsuke/github/extras/okhttp3/OkHttpResponseConnector.java) instead of ([`OkHttpConnector`](https://github.com/nedtwigg/github-api/blob/remove-http-connector/src/main/java/org/kohsuke/github/extras/okhttp3/OkHttpConnector.java) + [`GitHubHttpUrlConnectionClient`](https://github.com/nedtwigg/github-api/blob/remove-http-connector/src/main/java/org/kohsuke/github/GitHubHttpUrlConnectionClient.java)).

I believe the only breaking change in this PR is to `[Abuse/Rate]LimitHandler`, but it's a transparent change to all users who are using the built-in policies, and an easy change for users who aren't. The new names in the PR aren't great (e.g. `ResponseInfo`, `ResponseConnector`) but I just kept all names to match their name from where they were refactored from. I didn't mark any of the older stuff as deprecated.

One path forward might be to:

1. Merge and release this PR (with whatever feedback you'd like)
2. Let it rack up some mileage with users
3. Deprecate `HttpConnector` and all its usages
4. Remove the deprecations in a breaking change, and possibly rename classes at that point as well


nedtwigg:
Published in `3.33.2`.

bitwiseman:
Okay, this is embarrassing, but I'm working on basically the same change. I haven't looked at your change in any depth, but the description sounds much the same. I'll get you feedback shortly.

nedtwigg:
Lol! Well feel free to dump this PR and just merge yours :) All I care about is that I can do something like `OkHttpResponseConnector`

bitwiseman:
@nedtwigg 
So, yeah, we basically had the same idea. 
I got around the coverage problem by running part of the regular test pass on okhttp.  
Some naming differences of course.  
I also made the change non-breaking for both API and default behavior.

Would you be willing to review #1290 ? It should look very familiar. 😄

nedtwigg:
Absolutely! Review incoming.

src/main/java/org/kohsuke/github/GitHubClient.java
```diff
@@ -409,12 +401,28 @@ private void logRequest(@Nonnull final GitHubRequest request) {
     }
 
     @Nonnull
-    protected abstract GitHubResponse.ResponseInfo getResponseInfo(GitHubRequest request) throws IOException;
-
-    protected abstract void handleLimitingErrors(@Nonnull GitHubResponse.ResponseInfo responseInfo) throws IOException;
+    protected ResponseInfo getResponseInfo(GitHubRequest request) throws IOException {
+        return connector.getResponseInfo(request, authorizationProvider);
+    }
```
nedtwigg, line 406: I chose to treat the request as unmodifiable, whereas your `GitHubClient` has `private GitHubRequest prepareRequest(GitHubRequest request) throws IOException {`. In that method you add the auth headers, and also do some accept/encoding stuff that I don't understand very well. Probably my PR is just plain missing those features.

I think this way of handling authorization by passing the provider vs modifying the request is the only substantive difference between the two PRs, and I like your way better.

nedtwigg:
I like yours better :) Closing as dupe, looking forward to the next release with the new features.

bitwiseman:
> I chose to treat the request as unmodifiable, whereas your `GitHubClient` has `private GitHubRequest prepareRequest(GitHubRequest request) throws IOException {`. In that method you add the auth headers, and also do some accept/encoding stuff that I don't understand very well. Probably my PR is just plain missing those features.

> I think this way of handling authorization by passing the provider vs modifying the request is the only substantive difference between the two PRs, and I like your way better.

Requests are immutable, this creates a new request.  I should change the name of the method.

nedtwigg:
Fwiw I think your name `prepare` is good. I'll restate the case for passing `AuthorizationProvider`: `GitHubRequest` could be a type-level guarantee "this is what the client code requested" vs "this might be what the client code requested, or it might be a mutated copy with auth information and content-negotiation appended on". I was treating `GitHubRequest` as type-enforced pure-intent, but I think your approach is more practical.

Issue #727, opened 2020-03-04
Remove @author javadoc


bitwiseman:
I saw a number of projects making the choice to remove `@author` tags from their JavaDocs (https://github.com/checkstyle/checkstyle/issues/5738 for example).  

I thought it was a good choice - this is a collaborative project that has been going for years with over a hundred contributors and very few classes have only one contributor. 

I removed all the `@author` javadoc tags thinking this was not a controversial choice.  I didn't even send it through PR.  That was wrong to do.   

@PauloMigAlmeida asked that we bring them back and I realized that I had not given the change sufficient consideration.   I reverted the change in https://github.com/github-api/github-api/commit/a42305dd59f0be426f0c9091748e947f60d76bcd  so that we can discuss and do this right.

Paulo, could you talk about reasons to keep them?  Perhaps we can address your concerns with something other that this particular JavaDoc tag?

PauloMigAlmeida:
@bitwiseman first of all, it was really nice of you to revert the commit. I commend you for doing that. 

My 2 cents on whether keep or not the ```@author``` on the source code is ... We should keep them because:

1.  From the performance/software quality point of view, it doesn't impact the way it works.
2.  Keeping them can be interpreted (by some people) as a way of the open-source community express their gratitude for the time and effort that the author put into crafting that piece of code. 
3.  This is a common practice on lots of projects too: For instance, [OpenJDK](https://github.com/openjdk/jdk/blob/6bab0f539fba8fb441697846347597b4a0ade428/src/java.base/share/classes/java/lang/Override.java#L45), [Spring boot](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/Banner.java#L26), [Linux Kernel](https://elixir.bootlin.com/linux/v5.5.7/source/kernel/sched/fair.c#L5)
4.  This is a personal one but I confess I got to know several incredibly brilliant people by reading the ```@author```. For instance:
       1. by reading some of the java classes I learnt about Joshua Bloch, which eventually led me to his legendary book 'Effective Java'. 
       2. by doing a similar thing, I learnt about Ingo Molnar and his astonishing work on the [CFS](https://lkml.org/lkml/2007/4/13/180).
       3. The same goes to this library, I was looking into some of Kohsuke's work while he was working at Sun that eventually led me to this project.  And this led me to become a contributor and also to learn about @bitwiseman and his great work in a project that wasn't being proactively looked after. 

[...]

PR #1290, opened 2021-11-08
Swtich from HttpConnector to GitHubConnector

bitwiseman:
# Description

This change has the same aim as #1287 - remove this library's underlying dependency on `HttpURLConnection` and the limitations it entails.  

This PR introduces a new interface called `GitHubConnector` that replaces `HttpConnector`. It too has a single method, but this one takes `ConnectorRequest` and returns `ConnectorResponse`, interfaces controlled by this library instead of classes tied to any other library or java version. 

Significant effort was made to not change user-visible default behavior with this change.  Only users who change their code to use a non-adapter implementation of `GitHubConnector` (such as `OkHttpGitHubConnetor`) should see any noticeable change and even that should be limited to less used methods such as `GitHub.getConnector()`. 

# Before submitting a PR:

# When creating a PR: 


src/main/java/org/kohsuke/github/RateLimitHandler.java
```diff
@@ -12,6 +14,30 @@
...
+     * implementation needs to wait for some time.
+     *
+     * @param connectorResponse
+     *            Response information for this request.
+     *
+     * @throws IOException
+     *             the io exception
+     * @see <a href="https://developer.github.com/v3/#rate-limiting">API documentation from GitHub</a>
+     */
+    void onError(GitHubConnectorResponse connectorResponse) throws IOException {
```
nedtwigg, line 33: Do you want to `@Deprecated` the `HttpURLConnection` method?

src/main/java/org/kohsuke/github/AbuseLimitHandler.java
```diff
@@ -13,6 +15,33 @@
...
+     *            Response information for this request.
+     * @throws IOException
+     *             on failure
+     * @see <a href="https://developer.github.com/v3/#abuse-rate-limits">API documentation from GitHub</a>
+     * @see <a href=
+     *      "https://developer.github.com/v3/guides/best-practices-for-integrators/#dealing-with-abuse-rate-limits">Dealing
+     *      with abuse rate limits</a>
+     *
+     */
+    void onError(GitHubConnectorResponse connectorResponse) throws IOException {
```
nedtwigg, line 37: Do you want to `@Deprecated` the `HttpURLConnection` method?

nedtwigg (approved):
Looks great to me!

bitwiseman:
@nedtwigg 
I still need to fill in the JavaDoc info. 

Any other suggestions?

nedtwigg:
I love when javadoc has example inputs/outputs (this project scores 10/10 there) but I'm annoyed by rigid compliance with a javadoc linter forcing that gold to get diluted with "returns the value" boilerplate. Other than that personal style preference, no comments from me besides merge and release this beauty ;-)

bitwiseman:
> I love when javadoc has example inputs/outputs (this project scores 10/10 there) but I'm annoyed by rigid compliance with a javadoc linter forcing that gold to get diluted with "returns the value" boilerplate. Other than that personal style preference, no comments from me besides merge and release this beauty ;-)

Yeah, there is a bunch of autogenerated text to get to compliance.  Not great but compliant is still better than non-compliant.

Thanks!

nedtwigg:
> Not great but compliant is still better than non-compliant.

Why? I do a lot of work on Spotless, because I think formatting doesn't matter and I don't want to talk about formatting.

When it comes to documentation, I think it does matter, and so it's worth talking about. Automated tools can't say `obviously getFoo() returns foo` vs `it's interesting that getFoo() returns a String guaranteed to start with 'git://'`. So whether or not the documentation adds information beyond what's already in the name and type signature of the method is a human judgement call (for now, copilot is coming for us!). IMO, automated enforcement of

- does this documentation add information? ->  return true, always document everything, users don't need to read the implementation

isn't much better than

- does this documentation add information? ->  return false, never document anything, users need to read the implementation

If you add `options.addStringOption('Xdoclint:none', '-quiet')` then it will still error for undefined `@link` and things like that, but it lets "(to / not to) document" be the judgement call that imo it should be.

`</unsolicitedopinion>`

bitwiseman:
Re documentation: 
I'm all for streamlining where it makes sense.  
Hm, could you open an issue for this?  I think it is worth discussing on its own. See #727 for similar style discussion that happened a while back.

nedtwigg:
> could you open an issue for this

I've thought a lot about this topic, and I don't have any more to add besides this:

- Train a low-parameter neural net on the dataset `function name+signature -> javadoc` for your project.
- For each javadoc, diff the neural net's prediction against what the javadoc actually is
- Remove all the identical parts of the javadoc

I don't plan to build this, but imo that's the way I would think about it. Happy documenting ;-)

src/main/java/org/kohsuke/github/extras/okhttp3/OkHttpConnector.java
```diff
@@ -21,12 +32,11 @@
  * @author Liam Newman
  * @author Kohsuke Kawaguchi
  */
-public class OkHttpConnector implements HttpConnector {
+public class OkHttpConnector implements GitHubConnector {
```
bitwiseman, line 35: Darn.  `OkHttpConnector` need to continue to use the `HttpConnector` interface because it is may be extended by clients which will depend on further overriding `HttpConnector.connet()`.  

I'll need to create an `OkHttpGitHubConnector` class that contains this one.  I'll also need to retain `ObsoleteUrlFactory` for now as well.  😢

bitwiseman:
Okay, I've got the coverage up and expanded testing to make me confident in this change.

nedtwigg:
Huzzah, thanks a ton! Any blockers to getting this released? I can grab it from JitPack if so.

PR #1301, opened 2021-11-19
Add support for GitHub event branch_protection_rule

### LINKED ISSUE #1297: GHEvent enum is missing branch_protection_rule
**Describe the bug**
We are seeing this issue when Jenkins is receiving a **branch_protection_rule** webhook event, which is rejected with error "Error 400 Hook should contain event type"
Drilling down to [this piece of code](https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/GHEvent.java#L12), seems branch_protection_rule is missing from GHEvents enum.

**Expected behavior**
No error 400 is returned when receiving a GitHub branch_protection_rule event

**Additional context**
I can provide corresponding PR, just would need some guidance reg. scope of the change.
Is it only a matter of adding BRANCH_PROTECTION_RULE into GHEvent?

gjabouley-invn:
# Description
Add support for GitHub event _branch_protection_rule_
Fixes #1297 
Note i had a first look at [GHEventPayload.java](https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/GHEventPayload.java) but i don't know how to fully add the payload for this event (seems it also needs a new event class)

# Before submitting a PR:

- [X] Push your changes to a branch other than `main`. Create your PR from that branch.

# When creating a PR: 

### DIFF
--- a/src/main/java/org/kohsuke/github/GHEvent.java
+++ b/src/main/java/org/kohsuke/github/GHEvent.java
@@ -10,6 +10,7 @@ import java.util.Locale;
  * @see <a href="https://developer.github.com/v3/activity/events/types/">Event type reference</a>
  */
 public enum GHEvent {
+    BRANCH_PROTECTION_RULE,
     CHECK_RUN,
     CHECK_SUITE,
     CODE_SCANNING_ALERT,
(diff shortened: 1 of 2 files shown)


bitwiseman:
@gjabouley-invn 
> Note i had a first look at [GHEventPayload.java](https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/GHEventPayload.java) but i don't know how to fully add the payload for this event (seems it also needs a new event class)

Could you talk a bit more about what you mean here? Each payload gets a new nested class inside `GHEventPayload`.  What other "new event" would be needed?

bitwiseman (approved):

PR #1315, opened 2021-11-21
Refactor GHContentSearchBuilder to allow a Fork type

sahansera:
# Description

- Introduce a new enum `GHFork`.
- Refactor the existing code to use the new `GHFork` enum.
- Add a new unit test to test GitHub code search including results of forks.

# Before submitting a PR:

# When creating a PR: 


src/main/java/org/kohsuke/github/GHFork.java
```diff
@@ -0,0 +1,38 @@
+package org.kohsuke.github;
+
+/**
+ * The enum for Fork search mode
+ */
+public enum GHFork {
```
bitwiseman, line 6: This is copied from https://github.com/hub4j/github-api/blob/bcb71a3c318d243c92b757310cf153b8afc983d5/src/main/java/org/kohsuke/github/GHRepositorySearchBuilder.java#L219-L253 right?

src/main/java/org/kohsuke/github/GHRepositorySearchBuilder.java
```diff
@@ -76,8 +76,8 @@ public GHRepositorySearchBuilder forks(String v) {
      *      in forks</a>
      *
      */
-    public GHRepositorySearchBuilder fork(Fork fork) {
-        if (Fork.PARENT_ONLY.equals(fork)) {
+    public GHRepositorySearchBuilder fork(GHFork fork) {
```
bitwiseman, line 79: This is reasonable but to maintain backward compatibility we need to keep `Fork` and the related method and mark them both `@Deprecated`.

bitwiseman (changes requested):
Generally a good change.  Need adjustment maintain backward compatibility.

src/main/java/org/kohsuke/github/GHRepositorySearchBuilder.java
sahansera: Thanks a lot for your feedback. I have updated PR with the requested changes. Please kindly have another look and kick-off the build checks when you have some time :)

src/main/java/org/kohsuke/github/GHRepositorySearchBuilder.java
```diff
@@ -217,8 +251,9 @@ public GHRepositorySearchBuilder sort(Sort sort) {
     }
 
     /**
-     * The enum for Fork search mode
+     * The enum for Fork search mode. Note: Kept for backward compatibility. Use GHFork instead.
```
bitwiseman, line 254:
```suggestion
     * The enum for Fork search mode.
     *
     * @deprecated Kept for backward compatibility. Use {@link GHFork} instead.
```

src/main/java/org/kohsuke/github/GHRepositorySearchBuilder.java
```diff
@@ -47,9 +57,11 @@ public GHRepositorySearchBuilder size(String v) {
      * @param v
      *            the v
      * @return the gh repository search builder
+     * @deprecated use {@link #fork(GHFork)} instead.
      */
+    @Deprecated
     public GHRepositorySearchBuilder forks(String v) {
-        return q("forks:" + v);
```
bitwiseman, line 52: Well, this is was a bug.

bitwiseman (approved):

src/main/java/org/kohsuke/github/GHRepositorySearchBuilder.java
sahansera: Nice pickup! Thanks for reviewing this PR 🎉

PR #1316, opened 2021-11-21
Fix for Add WEBHOOK_JOB to GHEvent

### LINKED ISSUE #1288: GHEvents enum is missing workflow_job
**Describe the bug**
I’m trying to set up a webhook for a GitHub application which responds to the following events: create, delete, github_app_authorization, ping, workflow_dispatch, workflow_job, and workflow_run.  However, the code failed when it tried to find the GHEvent enum value for “workflow_job”.  I don’t know exactly what the difference between a “job” and a “run” is, but I can’t very well find out without setting up a webhook for both.

**To Reproduce**
```
            String event = "webhook_job";
            boolean found = false;
            for (GHEvent ghEvent : GHEvent.values()) {
                if (ghEvent.toString().equalsIgnoreCase(event)) {
                    found = true;
                    break;
                }
            }
            if (!found)
                throw new IllegalArgumentException("Invalid webhook event name: " + event);
```

**Expected behavior**
`found == true`

 - Java: 1.8
 - Version 1.133

**Additional context**
https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#workflow_job

katieshin:
# Description

Adding fix for https://github.com/hub4j/github-api/issues/1288 - new enum value for WORKFLOW_JOB in GHEvent.

Fixes #1288 

# Before submitting a PR:

# When creating a PR: 

### DIFF
--- a/src/main/java/org/kohsuke/github/GHEvent.java
+++ b/src/main/java/org/kohsuke/github/GHEvent.java
@@ -69,6 +69,7 @@ public enum GHEvent {
     TEAM,
     TEAM_ADD,
     WATCH,
+    WEBHOOK_JOB,
     WORKFLOW_DISPATCH,
     WORKFLOW_RUN,
 
--- a/src/main/java/org/kohsuke/github/GHEvent.java
+++ b/src/main/java/org/kohsuke/github/GHEvent.java
@@ -69,7 +69,7 @@ public enum GHEvent {
     TEAM,
     TEAM_ADD,
     WATCH,
-    WEBHOOK_JOB,
+    WORKFLOW_JOB,
     WORKFLOW_DISPATCH,
     WORKFLOW_RUN,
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/kohsuke/github/GHEvent.java
```diff
@@ -69,6 +69,7 @@
     TEAM,
     TEAM_ADD,
     WATCH,
+    WEBHOOK_JOB,
```
gsmet, line 72: It's not `WEBHOOK_JOB` but `WORKFLOW_JOB`.

gsmet (changes requested):

src/main/java/org/kohsuke/github/GHEvent.java
katieshin: Oh! I'll make the change, thanks.

gsmet: Also, my guess is that you will need to change things here: https://github.com/hub4j/github-api/blob/main/src/test/java/org/kohsuke/github/EnumTest.java#L31 .

Always run the tests after a change, especially in a code base you don't know as it's always full of surprises :).

katieshin: Thank you for the help!

gsmet (approved):
Maybe this should be squashed into one commit for a cleaner history before merging?

PR #1319, opened 2021-11-23
Add support for the Discussion payload

gsmet:
One thing that is a bit specific is that the Discussion object present
in the payload is coming from the GraphQL API and the Discussions API is
only available in the GraphQL API.

Note: I haven't implemented `discussion_comment` for now as the object is more complex and I don't really need it for now.


src/main/java/org/kohsuke/github/GraphQLDiscussion.java
```diff
@@ -0,0 +1,202 @@
...
+ * <p>
+ * This is different from Teams discussions (see {@link GHDiscussion}).
+ * <p>
+ * The discussion event exposes the GraphQL object directly. The new Discussions API is only available through GraphQL.
+ *
+ * @author Guillaume Smet
+ * @see <a href="https://docs.github.com/en/graphql/guides/using-the-graphql-api-for-discussions#discussion">The GraphQL
+ *      API for Discussions</a>
+ */
+public class GraphQLDiscussion {
```
gsmet, line 25: @bitwiseman I'm not sure about this class name. The specificity of it is that it's a GraphQL object injected in the REST API (the GitHub documentation here https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#discussion directly points to the GraphQL documentation). Also the Discussions API is not available as a REST API.

FTR, we already have a GHDiscussion object that is different as related to the Teams Discussions API and it's not the same.

I'm open to any suggestion as I'm not totally happy about this.

src/main/java/org/kohsuke/github/GraphQLDiscussion.java
```diff
@@ -0,0 +1,202 @@
...
+     *
+     * @return the root {@link GitHub} instance
+     */
+    @NonNull
+    GitHub root() {
+        return Objects.requireNonNull(root,
+                "The root GitHub reference for this instance is null. Probably caused by deserializing this class without using a GitHub instance. If you must do this, use the MappingObjectReader from GitHub.getMappingObjectReader().");
+    }
+
+    public static class Category {
```
bitwiseman, line 150: https://docs.github.com/en/graphql/guides/using-the-graphql-api-for-discussions#discussioncategory

bitwiseman (changes requested):
Well done!

This exposes discussions but does not provide any other way to read them because they are GraphQL and we don't support that yet.

bitwiseman:
Side note: 
The objects returned here are NOT the same as the GraphQL Discussion in some key ways - https://docs.github.com/en/graphql/reference/objects#discussion .  

For example, GraphQL has `id` as `ID!` which is a string - https://docs.github.com/en/graphql/reference/scalars#id  but these return `id` as `long` and have `node_id`. 😠 And you can't mix and match - REST API requires numeric ids while GraphQL requires `ID!` ids.  😡 😡

src/main/java/org/kohsuke/github/GraphQLDiscussion.java
gsmet: Did you add this comment for me to add a javadoc? Or do you want me to adjust the name?

gsmet:
@bitwiseman I pushed some adjustments.

src/main/java/org/kohsuke/github/GraphQLDiscussion.java
bitwiseman: Just add JavaDoc, but may also note that GHRepositoryDiscussion (and nested classes) are not exactly the same their GraphQL counterparts.

src/main/java/org/kohsuke/github/GHRepositoryDiscussion.java
```diff
@@ -0,0 +1,210 @@
...
+ * This is different from Teams discussions (see {@link GHDiscussion}).
+ * <p>
+ * The discussion event exposes (more or less) the GraphQL object directly. The new Discussions API is only available
+ * through GraphQL so for now you cannot execute any actions on this object.
+ *
+ * @author Guillaume Smet
+ * @see <a href="https://docs.github.com/en/graphql/guides/using-the-graphql-api-for-discussions#discussion">The GraphQL
+ *      API for Discussions</a>
+ */
+public class GHRepositoryDiscussion {
```
bitwiseman, line 26: Why not this? It should remove a bunch of fields and code from this file.  

```suggestion
public class GHRepositoryDiscussion extends GHObject {
```

At very least you can extend `GitHubInteractiveObject`.

src/main/java/org/kohsuke/github/GHRepositoryDiscussion.java
```diff
@@ -0,0 +1,210 @@
...
+    private String body;
+    private String timelineUrl;
+
+    public GHRepositoryDiscussion() {
+        this.root = null;
+    }
+
+    public GHRepositoryDiscussion(GitHub root) {
+        this.root = root;
+    }
```
bitwiseman, line 62: Not needed if you move to `GHObject` or `GitHubInteractiveObject`

bitwiseman (changes requested):
If you have a solid reason for not extending one of the base classes, I'm open to approving this.  I need to know what the reasoning is though.

src/main/java/org/kohsuke/github/GHRepositoryDiscussion.java
gsmet: Yeah so initially I didn't want that because this object was not really "interactive" given you can't interact with it. But yeah, you're right, it's probably easier to have everything consistent. I'll push something to fix it.

gsmet:
@bitwiseman I think I addressed all your comments. I squashed everything and rebased.

gsmet:
By the way, I know you released 1.300 a few days ago but it would really help us if you could release a 1.301 once this is in. As I would be able to support this in the new version of quarkus-github-app and add the Discussion triaging feature we wanted for the Quarkus GitHub bot.

Thanks for considering it!

bitwiseman (approved):

gsmet:
@bitwiseman thanks for the release, very much appreciated!

Issue #1243, opened 2021-09-28
Missing GHIssueCommentChanges


matusfaro:
Missing `GHIssueCommentChanges` similar to [`GHIssueChanges`](https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/GHIssueChanges.java) as per [`issues_comment`](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#issue_comment) GitHub API.

Thanks!

ReyhanPatria:
Hi! I'd like to work on this if you'd assign me to it @bitwiseman, thanks!

willtsai:
Hello! Full disclosure here: I'm working on this issue as well for my Software Engineering course project, so even if my eventual code changes become redundant, I can at least provide additional perspective on implementing this feature.

Thanks,
-Will

willtsai:
Hi @bitwiseman - I'm trying to use `mvn install -Dtest.github.takeSnapshot -Dtest.github.org=false -Dtest=YourTestClassName` per the contribution guide to run my unit test and take a json snapshot for the stub data so that I may convert it to a Wiremock test before I submit a PR. 

However, I have run `mvn install -Dtest.github.takeSnapshot -Dtest.github.org=false -Dtest=GHEventPayloadTest#issue_comment_edited` several times against the unit test below, but it's never been able to produce stub data stored in the `src/test/resources/org/kohsuhke/github/YourTestClassName/wiremock` path as described in the contributor guide. I'm running this on JDK 11, think that might be the issue? I haven't been able to get JDK 8 installed on my ARM64 Mac so have not tried this on Java 8.
```java
    @Test
    public void issue_comment_edited() throws Exception {
        GHIssueComment comment = getGitHubBuilder().withOAuthToken(System.getenv("GITHUB_OAUTH"))
                .build()
                .getRepository("willtsai/githubapi-testrepo")
                .getIssue(1)
                .getComments()
                .get(0);
        comment.update("comment one");
        String lastCommentBody = comment.getBody();
        comment.update("comment two");
        assertThat(comment.getBody(), is("comment two"));
    }
```

bitwiseman:
Not sure why this would not work for you. I'll invite you to the hub4j test  org. It might be a bug in the non-test-org feature.

willtsai:
ok thanks! I will try another unit test using `getRepository()` targeting the test org repo

willtsai:
[...]

PR #1322, opened 2021-11-24
Implement getChanges() for GHEventPayload.GHIssueComment and GHEventPayload.PullRequestReviewComment

willtsai:
# Description

# Before submitting a PR:

# When creating a PR: 

Fixes https://github.com/hub4j/github-api/issues/1243


willtsai:
Note: I chose to implement with a single `GHCommentChanges.java` wrapper class since the Webhook payload for both issue comments change and pull request review comment changes are identical:

```java
  "changes": {
    "body": {
      "from": "pull request comment BEFORE edit"
    }
  },
```

src/main/java/org/kohsuke/github/GHCommentChanges.java
```diff
@@ -0,0 +1,40 @@
...
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+
+/**
+ * Wrapper for changes on issue and pull request review comments action="edited"
+ *
+ * @see GHEventPayload.IssueComment
+ * @see GHEventPayload.PullRequestReviewComment
+ */
+@SuppressFBWarnings(value = "UWF_UNWRITTEN_FIELD", justification = "JSON API")
+public class GHCommentChanges {
```
gsmet, line 12: I wonder if this class should be a static nested class in `GHEventPayload`. This class is getting big but I think it makes more sense than having it as a top level class.

Let's wait for @bitwiseman 's opinion before changing anything though.

willtsai: @gsmet - I agree with you here actually, especially since `GHCommentChanges` is only applicable to event payloads. But I'll also wait for @bitwiseman's feedback before making any changes.

Perhaps others like `GHIssueChanges` also need to be refactored in this way? (in a separate PR)

bitwiseman: Agreed, this is only used by events, it should go in `GHEventPayload`.

bitwiseman: Also, as a static nested class it can be called `CommentChanges` without the `GH`.

willtsai: Made these changes, please see latest commit. Thanks!

willtsai:
Committed changes to move CommentChanges into static class within GHEventPayload. Please review again, thanks!

bitwiseman (approved):

Issue #1191, opened 2021-06-28
Fix GHContentUpdatedResponse to return GitCommit instead of GHCommit


bitwiseman:
In #1185, @takezoe discovered that the date returned for `GHContentUpdatedResponse.commit` does not match up with `GHCommit`.  It most closely matches `GHCommit.ShortInfo` but even that is not truly accurate.  

In #1190 , I figured out that while this a bad the functionality is not completely broken. The `sha` and the `url` match up and that is enough for the `GHCommit` to populate the missing data and continue on.  This is still not great because it results in additional requests where there probably shouldn't be any. 

As discussed in #1185, what we should do is create a new class call `GitCommit` , matching the structure of the `repo/commit` endpoint and then have other classes like `GHCommit.ShortInfo` extend that class.

ecxia:
hi @bitwiseman - I'm taking a look at this one. Can I get access to the organization for testing? Thanks!

bitwiseman:
@ecxia  This change is not marked as a good first issue.  Are you sure you want to take this on?

ecxia:
@bitwiseman i have noticed that it is not straightforward but I will give it a shot！

bitwiseman:
@ecxia 
Okay!  Feel free to create a draft PR of your changes earlier your development process if you want early feedback to ensure you're going in a good direction.

PR #1329, opened 2021-11-30
Add gitcommit class

ecxia:
# Description

Directional draft PR for #1191, very messy! Is this in line with what you were thinking? In particular, I note that the `GitCommit` class is based off of [the endpoint you found that gets a git commit](https://docs.github.com/en/rest/reference/git#get-a-commit), but currently the code changes I've made don't actually contemplate calling that endpoint - just using it for structure.

I'm also a bit concerned about the amount of info that ends up duplicated between `GitCommit` and `GHCommit` (outside of `commit`), with `parents` / `getParents` being the trickiest.

Thanks for your patience!

# Before submitting a PR:

# When creating a PR: 


src/main/java/org/kohsuke/github/GHCommit.java
```diff
@@ -236,6 +170,7 @@ public URL getBlobUrl() {
         public String getSha() {
             return sha;
         }
+
     }
 
     /**
```
bitwiseman, line 176: Ugh, why is `Parent` (three lines below this one) public?  Not your fault.  Very odd.

src/main/java/org/kohsuke/github/GitCommit.java
```diff
@@ -0,0 +1,224 @@
...
+
+    public List<GHCommit> getParents() throws IOException {
+        List<GHCommit> r = new ArrayList<GHCommit>();
+        for (String sha1 : getParentSHA1s())
+            r.add(owner.getCommit(sha1));
+        return r;
+    }
+
+    @SuppressFBWarnings(value = "EI_EXPOSE_REP", justification = "acceptable")
+    public List<GHCommit.Parent> getRawParents() {
```
bitwiseman, line 191: This should probably be package private.
```suggestion
    List<GHCommit.Parent> getRawParents() {
```

src/main/java/org/kohsuke/github/GHCommit.java
```diff
@@ -34,94 +33,29 @@
...
-         *
-         * @return the Verification status
-         */
-        public GHVerification getVerification() {
-            return verification;
+        public ShortInfo() {
+        };
+        public ShortInfo(GitCommit commit) {
+            super(commit);
+            comment_count = -1;
```
bitwiseman, line 56: Is this needed? 

You can default `comment_count` when declared: `private int comment_count = -1;`

Also, you'll want to override `ShortInfo.getParentSHA1s()` to either throw a `GHException` or call `GHCommit.getParentSHA1s()`.   Otherwise it will return empty.

bitwiseman:
> I'm also a bit concerned about the amount of info that ends up duplicated between `GitCommit` and `GHCommit` (outside of `commit`), with `parents` / `getParents` being the trickiest.

There's only so much you can do when the structures returned have a bunch of duplicate data in them.  What alternative do you have?

src/main/java/org/kohsuke/github/GitCommit.java
```diff
@@ -0,0 +1,224 @@
...
+        this.node_id = commit.getNodeId();
+        this.html_url = commit.getHtmlUrl();
+        this.url = commit.getUrl();
+        this.author = commit.getAuthor();
+        this.committer = commit.getCommitter();
+        this.message = commit.getMessage();
+        this.verification = commit.getVerification();
+        this.tree = commit.getTree();
+        this.parents = commit.getRawParents();
+    }
```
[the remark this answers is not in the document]
ecxia: this is for the bridge method conversion to `GHCommit.ShortInfo`, so I think it's actually needed (Jackson uses the default constructor a few lines up).

ecxia:
Thank you for the feedback!

src/main/java/org/kohsuke/github/GitCommit.java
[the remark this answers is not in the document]
bitwiseman: Ah, I see.  Okay.

ecxia:
Ah. Failing on coverage checks - apparently the GHContent integration test is insufficient here. I'll add a wiremock test for the new class.

ecxia:
I'm actually pretty confused by this - it doesn't make any sense to me that jacoco is returning *0* test coverage for the GitCommit method, given that a good chunk of the content is checked in `GHContentIntegrationTest`. Would you have any advice on what the reason could be?

bitwiseman:
The message is: 
```
Warning:  Rule violated for class org.kohsuke.github.GitCommit.1: methods covered ratio is 0.00, but expected minimum is 0.50
Warning:  Rule violated for class org.kohsuke.github.GHCommit.GHAuthor: methods covered ratio is 0.00, but expected minimum is 0.50
```

src/main/java/org/kohsuke/github/GitCommit.java
```diff
@@ -0,0 +1,211 @@
...
+            @Override
+            public String get(int index) {
+                return parents.get(index).sha;
+            }
+
+            @Override
+            public int size() {
+                return parents.size();
+            }
+        };
```
bitwiseman, line 203: This is `org.kohsuke.github.GitCommit.1`.

src/main/java/org/kohsuke/github/GitCommit.java
```diff
@@ -0,0 +1,211 @@
...
+ * @author Emily Xia-Reinert
+ * @see GHContentUpdateResponse#getCommit() GHContentUpdateResponse#getCommit()
+ */
+
+@SuppressFBWarnings(value = { "NP_UNWRITTEN_FIELD", "UWF_UNWRITTEN_FIELD" }, justification = "JSON API")
+public class GitCommit {
+    private GHRepository owner;
+    private String sha, node_id, url, html_url;
+    private GitUser author;
+    private GitUser committer;
```
bitwiseman, line 21: These are stored as `GHAuthor` in the old code. 

```
        private GHAuthor author;
        private GHAuthor committer;
```

bitwiseman: I think the reason the coverage went to 0 is that these were changed to `GitUser` and the bridge method that returns `GHAuthor` is never called (requires reflection to be called).

src/test/java/org/kohsuke/github/GHContentIntegrationTest.java
```diff
@@ -85,6 +89,17 @@ public void testGetDirectoryContentTrailingSlash() throws Exception {
         assertThat(entries.get(0).getUrl(), endsWith("?ref=main"));
     }
 
+    GHCommit getGHCommit(GHContentUpdateResponse resp)
+            throws GHException, IllegalAccessException, IllegalArgumentException, InvocationTargetException {
+        for (Method method : resp.getClass().getMethods()) {
+            if (method.getName() == "getCommit" && method.getReturnType() == GHCommit.class) {
+                return (GHCommit) method.invoke(resp);
+            }
```
bitwiseman, line 97: To get coverage on `GHCommit.GHAuthor` you'll need to call `GHAuthor commit.getAuthor()` using reflection as well I think.

src/test/java/org/kohsuke/github/GHContentIntegrationTest.java
```diff
@@ -217,6 +173,110 @@ public void testCRUDContent() throws Exception {
...
+        assertThat("GHCommit resolving GHTree is not cached",
+                mockGitHub.getRequestCount(),
+                equalTo(expectedRequestCount += 2));
+
+        return expectedRequestCount;
+    }
+
+    int checkCommitParents(GitCommit gitCommit, GHCommit ghCommit, int expectedRequestCount) throws IOException {
+        assertThat(gitCommit.getParentSHA1s(), hasSize(greaterThan(0)));
+        assertThat(ghCommit.getParentSHA1s(), hasSize(greaterThan(0)));
```
bitwiseman, line 276: Hm... maybe `hasSize()` calls something other than `size()`?  Try changing this test to 
```suggestion
        assertThat(gitCommit.getParentSHA1s().size(), is(greaterThan(0)));
        assertThat(ghCommit.getParentSHA1s().size(), is(greaterThan(0)));

        assertThat(gitCommit.getParentSHA1s().get(0), notNullValue());
        assertThat(ghCommit.getParentSHA1s().get(0), notNullValue());
```

src/main/java/org/kohsuke/github/GHContentUpdateResponse.java
```diff
@@ -25,7 +26,14 @@ public GHContent getContent() {
...
      */
     @SuppressFBWarnings(value = { "EI_EXPOSE_REP" }, justification = "Expected behavior")
-    public GHCommit getCommit() {
+    @WithBridgeMethods(value = GHCommit.class, adapterMethod = "gitCommitToGHCommit")
+    public GitCommit getCommit() {
         return commit;
     }
+
+    @SuppressFBWarnings(value = "UPM_UNCALLED_PRIVATE_METHOD", justification = "bridge method of getCommit")
+    public Object gitCommitToGHCommit(GitCommit commit, Class targetType) {
```
bitwiseman, line 35:
```suggestion
    private Object gitCommitToGHCommit(GitCommit commit, Class targetType) {
```

bitwiseman, line 35:
```suggestion
    private Object gitCommitToGHCommit(GitCommit commit, Class targetType) {
```

src/main/java/org/kohsuke/github/GitCommit.java
bitwiseman: You might have to change these back to `GHCommit.GHAuthor` and add clear comments that they should be changed for v2. 

Add `@Deprecated` annotation here (not just the JavaDoc):
https://github.com/hub4j/github-api/blob/eaef661255199d5b0c022e15d70096de0a39b3cd/src/main/java/org/kohsuke/github/GHCommit.java#L75

src/main/java/org/kohsuke/github/GitCommit.java
ecxia: oh I see - the message is pointing to the method, not the whole class. That makes a lot more sense to me, thank you!

ecxia:
I tried calling the existing `GHAuthor` bridge method with reflection and it actually failed, so I tried using the same kind of adapter method (inherited copy constructors) as with `GitCommit`/`GHCommit`.

ecxia:
Oh my goodness, I'm sorry, I clicked the "Commit suggestion" button and then didn't make sure the changes had propagated correctly.

ecxia:
Oh, method to call commit parents wasn't being called. I'm sorry there's so much back and forth on this one!

bitwiseman:
> Oh, method to call commit parents wasn't being called. I'm sorry there's so much back and forth on this one!

You're doing great.  Thanks for working through this!

src/test/java/org/kohsuke/github/GHContentIntegrationTest.java
```diff
@@ -260,6 +354,7 @@ public void testGetFileContentWithNonAsciiPath() throws Exception {
     }
 
     @Test
+    @Ignore
```
bitwiseman, line 357: Why ignored?

src/main/java/org/kohsuke/github/GitCommit.java
```diff
@@ -0,0 +1,220 @@
...
+        return tree.getUrl();
+    }
+
+    @SuppressFBWarnings(value = "EI_EXPOSE_REP", justification = "acceptable")
+    List<GHCommit.Parent> getParents() {
+        return parents;
+    }
+
+    public List<String> getParentSHA1s() {
+        if (parents == null)
```
bitwiseman, line 200:
```suggestion
        if (parents == null || parents.size() == 0)
```

src/test/java/org/kohsuke/github/GHContentIntegrationTest.java
ecxia: That's a mistake left over from when I was messing around with tests - sorry about that!

ecxia:
hi - I think after my last commit (and pending rebase) this should be ready for review.

bitwiseman:
@ecxia 
Thanks for your patience on this.   Great work!

ecxia:
Thank you!

Issue #1265, opened 2021-10-15
GHWorkflow.dispatch does not throw an exception on a 422 HTTP status


Typraeurion:
**Describe the bug**
I’ve been tracing the code to see why my application apparently succeeds in sending a workflow dispatch request yet the workflow never starts.  Eventually I found that after sending a POST request to /repos/{owner}/{repo}/actions/workflows/{workflow_id}/dispatches (no body, because I had passed `null` as the ref), GitHub responds with error 422 Unprocessable Entity.  GitHubClient isn’t throwing an exception for this code; instead, the GitHubClient.sendRequest method returns a GitHubResponse,Requester.fetchHttpStatusCode returns 422, and GHWorkflow.dispatch ignores it, so my application thinks the request succeeded when it actually failed.

**To Reproduce**
```
        // authToken, organization, repository, and workflowId are set;
        // branch = null; inputs = Collections.emptyMap();

        GHWorkflow workflow = getConnection(authToken).getRepository(
                String.format("%s/%s", organization, repository))
        .getWorkflow(workflowId);

        workflow.dispatch(branch, Optional.ofNullable(inputs)
                .orElse(Collections.emptyMap()));
```

**Expected behavior**
The method should either have thrown a GHIOException (or appropriate subclass for HTTP status 422), or should see that `branch` is null and throw an IllegalArgumentException.

**Desktop (please complete the following information):**
 - OS: MacOS 11.4
 - Java: 1.8.0_162
 - github-api Version: 1.133
- okhttp Version: 4.9.2

**Additional context**
Per https://docs.github.com/en/rest/reference/actions#create-a-workflow-dispatch-event, a `ref` parameter in the request body is *required*.

bitwiseman:
@Typraeurion 
The problem is in the `dispatch` method as you would expect:
https://github.com/hub4j/github-api/blob/a28a726b9c78b007a2b45e60a1bc73a6a3bf1804/src/main/java/org/kohsuke/github/GHWorkflow.java#L126-L136

Either than method should throw when it gets `null` for the ref, or the last line should be changed to call `send()` instead of `fetchStatusCode()`.

PR #1337, opened 2021-12-05
fix: fixed issue #1265

yixingz3:
# Description

Fixes issue #1265. 
- added null and empty check for 'ref' and throw IllegalArgumentException if null or empty is passed in. I'm not sure what would be the most appropriate type of exception to throw and would be very happy to change it if you see fits. 
- added tests to verify the issue is fixed

# Before submitting a PR:

# When creating a PR: 

### DIFF
--- a/src/main/java/org/kohsuke/github/GHWorkflow.java
+++ b/src/main/java/org/kohsuke/github/GHWorkflow.java
@@ -124,6 +124,10 @@ public class GHWorkflow extends GHObject {
      *             the io exception
      */
     public void dispatch(String ref, Map<String, Object> inputs) throws IOException {
+        if (ref == null || ref.length() == 0) {
+            throw new GHIOException("ref is null or empty.");
+        }
+        
         Requester requester = root().createRequest()
                 .method("POST")
                 .withUrlPath(getApiRoute(), "dispatches")
--- a/src/main/java/org/kohsuke/github/GHWorkflow.java
+++ b/src/main/java/org/kohsuke/github/GHWorkflow.java
@@ -122,12 +122,14 @@ public class GHWorkflow extends GHObject {
      *            default properties configured in the workflow file will be used when inputs are omitted.
      * @throws IOException
      *             the io exception
+     * @throws IllegalArgumentException
+     *             the illegal argument exception
      */
-    public void dispatch(String ref, Map<String, Object> inputs) throws IOException {
+    public void dispatch(String ref, Map<String, Object> inputs) throws IOException, IllegalArgumentException {
         if (ref == null || ref.length() == 0) {
-            throw new GHIOException("ref is null or empty.");
+            throw new IllegalArgumentException("ref cannot be null or empty");
         }
-        
+
         Requester requester = root().createRequest()
                 .method("POST")
                 .withUrlPath(getApiRoute(), "dispatches")
--- a/src/test/java/org/kohsuke/github/GHWorkflowTest.java
+++ b/src/test/java/org/kohsuke/github/GHWorkflowTest.java
@@ -78,16 +78,16 @@ public class GHWorkflowTest extends AbstractGitHubWireMockTest {
     }
 
     @Test
-    public void testDispatch() throws IOException {
+    public void testDispatch() throws IOException, IllegalArgumentException {
         GHWorkflow workflow = repo.getWorkflow("test-workflow.yml");
 
-        String expectedMessage = "ref is null or empty.";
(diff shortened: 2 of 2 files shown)


src/main/java/org/kohsuke/github/GHWorkflow.java
```diff
@@ -122,8 +122,14 @@ public void dispatch(String ref) throws IOException {
...
      *             the io exception
+     * @throws IllegalArgumentException
+     *             the illegal argument exception
      */
-    public void dispatch(String ref, Map<String, Object> inputs) throws IOException {
+    public void dispatch(String ref, Map<String, Object> inputs) throws IOException, IllegalArgumentException {
+        if (ref == null || ref.length() == 0) {
+            throw new IllegalArgumentException("ref cannot be null or empty");
+        }
+
```
bitwiseman, line 132: I think you'll also need to update Line 142 to use `send()`. 

What happens if `ref` is an invalid/missing ref?

yixingz3: Yes, that makes sense. Thank you for spotting it! Fixed.

src/main/java/org/kohsuke/github/GHWorkflow.java
```diff
@@ -122,8 +122,14 @@ public void dispatch(String ref) throws IOException {
      *            default properties configured in the workflow file will be used when inputs are omitted.
      * @throws IOException
      *             the io exception
+     * @throws IllegalArgumentException
+     *             the illegal argument exception
```
bitwiseman, line 126:
```suggestion
```

src/main/java/org/kohsuke/github/GHWorkflow.java
```diff
@@ -122,8 +122,14 @@ public void dispatch(String ref) throws IOException {
...
      */
-    public void dispatch(String ref, Map<String, Object> inputs) throws IOException {
+    public void dispatch(String ref, Map<String, Object> inputs) throws IOException, IllegalArgumentException {
```
bitwiseman, line 128: Adding an exception is technically an API change.  Also, `IllegalArgumentException` is a `RuntimeException` so it doesn't have to be part of the signature. 

```suggestion
    public void dispatch(String ref, Map<String, Object> inputs) throws IOException {
```

src/test/java/org/kohsuke/github/GHWorkflowTest.java
```diff
@@ -77,9 +78,20 @@ public void testDisableEnable() throws IOException {
...
+        IllegalArgumentException nullException = assertThrows(IllegalArgumentException.class, () -> {
+            workflow.dispatch(null);
+        });
+        assertThat(nullException.getMessage(), equalTo(expectedMessage));
+
+        IllegalArgumentException emptyException = assertThrows(IllegalArgumentException.class, () -> {
+            workflow.dispatch("");
+        });
+        assertThat(emptyException.getMessage(), equalTo(expectedMessage));
+
```
bitwiseman, line 94:
```java
        // What happens in this case? 
        Exception emptyException = assertThrows(Exception.class, () -> {
            workflow.dispatch("invalidRef");
        });
```

bitwiseman (changes requested):
A couple more things. Sorry about the multiple review cycles.

src/test/java/org/kohsuke/github/GHWorkflowTest.java
yixingz3: @bitwiseman I'm not super sure what would happen in this case... As of now, 'ref' as a string, is not null, and has a length so it can't be empty or missing. Then if the value of the string 'ref' is somehow invalid (e.g. a value that's not in the list of branch name or tag name as that's the value it represents). Let's say ref = 'invalidRef', the current tests cases with requester.send() would return the json file - _repos-hub4j-test-org_ghworkflowtest_actions_workflows_6817859_dispatches-5.json_ and since it has the status code 422, the current test cases will fail at the last case with singletonMap as inputs.
Should I change to status code to 204, same as _repos-hub4j-test-org_ghworkflowtest_actions_workflows_6817859_dispatches-4.json_, which doesn't have the inputs, to fix the test? And for the invalid ref, what would you say would be best way to fix it?

bitwiseman: @yixingz3 
Please try testing manually and recording _new_ data. See CONTRIBUTING.md for how to do this.

Issue #1325, opened 2021-11-25
Accessing Gists via getMyself().listGists() returns null data for Gist files


EasyG0ing1:
Once I authenticate to GitHub with a valid access token, but then attempt to pull the Map of Gists using this method:

`gitHub.getMyself().listGists();`

It will give me a map of GHGist objects and their gistId's, but when I then use the GHGist object to pull the files out of Gist, the file data comes back null.

The only way I am able to access files successfully, is if I first iterate through that map and build a list of Gist Ids like this:

```
 List<String> gistIDs = new ArrayList<>();
 for (GHGist gist : gitHub.getMyself().listGists()) {
     String gistID = gist.getGistId();
     gistIDs.add(gistID);
 }
```

Then iterate through that list and specifically call for the Gist object from the parent connection like this:

```
for (String gistID : gistIDs) {
    GHGist gist = gitHub.getGist(gistID);
    for (String filename : gist.getFiles().keySet()) {
        GHGistFile file = gist.getFiles().get(filename);
        System.out.println("Contents: " + file.getContent());
    }
}

```
ONLY by using this two-step process, can I access the files in the Gist. I believe this is a bug in the API.

Thank you,

Mike

EasyG0ing1:
After switching to version 1.300 of the API, this seems to be working correctly.

SCHJonathan:
@bitwiseman Hi Bitwiseman! May I request the access to `hub4j-test-org` as I managed to have a fix for this issue and want to submit my PR? I found this issue is caused by the file content field is not returned by the API and we need to do a lazy download from the `raw_url` field in case the file content is needed but has a `null` value.

bitwiseman:
@SCHJonathan 
Invitation sent.

PR #1339, opened 2021-12-07
fix: fix issue #1325 and add missing `truncated` field in the `GHGist` object

SCHJonathan:
# Description

1. Fixes [#1325](https://github.com/hub4j/github-api/issues/1325) by adding a lazy download from the `raw_url` field of `GHGistFile` object if the gist file `content` is uninitialized.
2. From the [list public gist](https://docs.github.com/en/rest/reference/gists#list-public-gists) documentation, I found out we are missing the `truncated` field in the `GHGist` object and I added that field in this PR.

# Before submitting a PR:

# When creating a PR: 


src/main/java/org/kohsuke/github/GHGistFile.java
```diff
@@ -60,11 +64,16 @@ public String getLanguage() {
...
+     * Content of this file. Lazy download the content in case the size is nonzero but the content is empty. Function
+     * modified for <a href="https://github.com/hub4j/github-api/issues/1325">issue #1325</a>.
      *
      * @return the content
      */
-    public String getContent() {
+    public String getContent() throws IOException {
+        if (this.content == null && this.size > 0) {
+            InputStream inputStream = new URL(this.raw_url).openStream();
+            this.content = new String(inputStream.readAllBytes());
```
bitwiseman, line 75: This works while testing, but this circumvents the `GitHubClient`.  If the user has configured proxying or expects to use a different http client library, this will not function the way they want.  Please use the the `Requester` class like all other parts of the library.

bitwiseman: https://github.com/hub4j/github-api/blob/06b031278894f3a1a5ea804c2720e0e7e0baaf63/src/main/java/org/kohsuke/github/Requester.java#L116

bitwiseman: Also, if the size is 0, we should return `""` right?

src/main/java/org/kohsuke/github/GHGistFile.java
```diff
@@ -60,11 +64,16 @@ public String getLanguage() {
     }
 
     /**
-     * Content of this file.
+     * Content of this file. Lazy download the content in case the size is nonzero but the content is empty. Function
+     * modified for <a href="https://github.com/hub4j/github-api/issues/1325">issue #1325</a>.
```
bitwiseman, line 68:
```suggestion
     * Content of this file.
     * 
     * Content is lazy downloaded and cached locally the first time this method is called.
```

src/main/java/org/kohsuke/github/GHGistFile.java
```diff
@@ -60,11 +64,16 @@ public String getLanguage() {
...
      *
      * @return the content
      */
-    public String getContent() {
+    public String getContent() throws IOException {
```
bitwiseman, line 72: Sorry, adding an exception is considered an API change. You'll need to wrap the `IOException` in a `GHException` (which is a `RuntimeException` that can be thrown without being declared.  Add comments in the JavaDoc Explaining the situation.   
```suggestion
    public String getContent() {
```

You might add a new method  `public String getContentString() throws IOException {` and then have the existing `getContent()` wrap the new method.

bitwiseman (changes requested):
Thanks for contributing.  

Sorry about all the comments.  This is basically a good PR with test coverage an everything, but you ran it in a bunch of the gotchas of working in the library all at once.  

I'm happy to answer any questions you might have about the comments.

SCHJonathan:
> 

@bitwiseman. Thank you so much for your help and detailed comment, bitwiseman! All of those suggestions are super helpful and I have made revisions based on your suggestions!

src/main/java/org/kohsuke/github/GHGistFile.java
```diff
@@ -62,10 +69,27 @@ public String getLanguage() {
     /**
      * Content of this file.
      *
+     * Content is lazy downloaded and cached locally the first time this method is called.
+     * Function modified for <a href="https://github.com/hub4j/github-api/issues/1325">issue #1325</a>.
```
bitwiseman, line 73: Users a year from now won't care about this. 
```suggestion
```

src/main/java/org/kohsuke/github/GHGistFile.java
```diff
@@ -62,10 +69,27 @@ public String getLanguage() {
...
+            try {
+                InputStream inputStream = this.owner.root().createRequest()
+                        .setRawUrlPath(raw_url)
+                        .fetchStream(Requester::copyInputStream);
+                this.content = IOUtils.toString(inputStream, Charset.defaultCharset());
+            } catch (IOException e) {
+                throw new GHException("Failed to retrieve the gist file content", e);
+            }
+        }
+        return this.content;
```
bitwiseman, line 92: The way you have it written, if the first call to `getContent()` throws, the next call will always return `""`.  Is that what you think should happen?  

```suggestion
        if (this.content == null) {
            if (this.size == 0) {
                this.content = "";
            } else {
                try {
                    InputStream inputStream = this.owner.root().createRequest()
                            .setRawUrlPath(raw_url)
                            .fetchStream(Requester::copyInputStream);
                    this.content = IOUtils.toString(inputStream, Charset.defaultCharset());
                } catch (IOException e) {
                    throw new GHException("Failed to retrieve the gist file content", e);
                }
            }
        }
        return this.content;
```

bitwiseman (changes requested):
A few more bits of feedback.

src/main/java/org/kohsuke/github/GHGistFile.java
SCHJonathan: Ohhh!! Nice catch! Thank you so much for the suggestions! I will make my revision right away

src/main/java/org/kohsuke/github/GHGistFile.java
```diff
@@ -76,4 +100,14 @@ public String getContent() {
...
+
+
+    /**
+     * set the owner object of the GHGistFile.
+     *
+     * @param owner the object stores owner information
+     */
+    public void setOwner(GHUser owner) {
+        this.owner = owner;
+    }
```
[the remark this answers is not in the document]
SCHJonathan: > Sorry, why add this?

Sorry, the reason I am adding this setter is that I want to pass the `owner` object from the `GHGist` object to each `GHGistFile` object so that get content can use `owner.root()` to create request. At a second thought, probably defining a `wrapUp` method to set the owner as the `GHGist` object owning this current `GHGistFile` object would make more sense. I have make this revision accordingly. Thank you again for your suggestions.

bitwiseman:
@SCHJonathan 
You're very close on this.  Do you time to finish this PR in the next few days?  I plan to do a release soon.

src/main/java/org/kohsuke/github/GHGistFile.java
```diff
@@ -76,4 +99,14 @@ public String getContent() {
...
         return truncated;
     }
+
+    /**
+     * set the owner object of the GHGistFile.
+     *
+     * @param owner
+     *            the object stores owner information
+     */
+    public void wrapUp(GHGist owner) {
```
bitwiseman, line 109:
```suggestion
    void wrapUp(GHGist owner) {
```

bitwiseman (changes requested):
It looks like some test is failing.

PR #1371, opened 2022-02-02
Fix installation payload handling

gsmet:
Hey there,

Long time no see :).

The problem was initially reported in my Quarkus GitHub App project: https://github.com/quarkiverse/quarkus-github-app/issues/253. 

Parsing `Installation` event payloads with a connected `GitHub` instance (i.e. not an offline one) leads to:
```
Exception: java.lang.NullPointerException: Missing instance URL!
	at java.base/java.util.Objects.requireNonNull(Objects.java:247)
	at org.kohsuke.github.GHRepository.populate(GHRepository.java:3260)
	at org.kohsuke.github.GHEventPayload$Installation.lateBind(GHEventPayload.java:276)
	at org.kohsuke.github.GitHub.parseEventPayload(GitHub.java:944)
```

Installation payloads do not contain the URL of the repository so we cannot populate the data from there. You can see an example in the tests here:
https://github.com/hub4j/github-api/blob/main/src/test/resources/org/kohsuke/github/GHEventPayloadTest/installation.json#L45-L53

I think what I did is safe but better check the URL thing as I decided to always use a forged URL rather than the one provided in the JSON. I can do it in another way if you prefer and keep the JSON URL when it's provided but, given we already had a try/catch for a corner case, I thought we might as well always forge the URL.

The tests are not failing because we are testing all the payloads with an offline `GitHub` instance thus not populating the repositories. Maybe we should change that? But I'm not totally sure of the consequences of this change. So I thought I might as well ask.

Very motivated to get this change in as one of my users is blocked by this so eager to hear your feedback :).


src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -2978,7 +2977,7 @@ String getApiTailUrl(String tail) {
         if (tail.length() > 0 && !tail.startsWith("/")) {
             tail = '/' + tail;
         }
-        return "/repos/" + getOwnerName() + "/" + name + tail;
+        return "/repos/" + full_name + tail;
```
gsmet, line 2980: FWIW, the owner is not provided either in the installation token, the only info we have is the full name.

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -3257,29 +3256,13 @@ void populate() throws IOException {
...
-                root().createRequest()
-                        .withPreview(BAPTISTE)
-                        .withPreview(NEBULA)
-                        .withUrlPath("/repos/" + full_name)
-                        .fetchInto(this);
-            } else {
-                throw e;
-            }
-        }
+        root().createRequest().withPreview(BAPTISTE).withPreview(NEBULA).withUrlPath(getApiTailUrl("")).fetchInto(this);
```
gsmet, line 3265: Decided to simplify that but I can add a bunch of ifs if you prefer using the URL coming from the JSON when it is around.

gsmachado:
I can confirm that this PR fixes the issue on my GitHub app. 👍  💪  💯

src/test/java/org/kohsuke/github/GHEventPayloadTest.java
```diff
@@ -630,7 +630,7 @@ public void checkRunEvent() throws Exception {
                 GHEventPayload.CheckRun.class);
         final GHCheckRun checkRun2 = verifyBasicCheckRunEvent(event2);
 
-        int expectedRequestCount = mockGitHub.isUseProxy() ? 3 : 2;
+        int expectedRequestCount = 2;
```
gsmet, line 633: I refreshed the snapshots and this is no longer needed.

bitwiseman:
Sorry for the slow response. Looking now.

src/main/java/org/kohsuke/github/GHRepository.java
bitwiseman: No, this is functional and clean. 💯

bitwiseman (approved):
Looking good. 
Is there anyway to raise the code coverage?

src/main/java/org/kohsuke/github/GHEventPayload.java
```diff
@@ -267,12 +267,15 @@ void lateBind() {
...
-                        "Expected check_suite payload, but got something else. Maybe we've got another type of event?");
+                        "Expected installation payload, but got something else. Maybe we've got another type of event?");
             }
             super.lateBind();
             if (repositories != null && !repositories.isEmpty()) {
                 try {
-                    for (GHRepository singleRepo : repositories) { // warp each of the repository
+                    for (GHRepository singleRepo : repositories) {
+                        // populate each repository
+                        // the repository information provided here is so limited
```
bitwiseman, line 277:
```suggestion
                        // the repository information provided here is so limited
```

gsmet:
Hey @bitwiseman , any chance we could have a release with this fix? Thanks!

gsmet:
@bitwiseman thanks!

Issue #1377, opened 2022-02-22
GHOrganization.Permission does not support custom roles


jgangemi:
github added the ability to define custom roles, however the `Permission` object is an enum and this not extensible.

i was thinking of converting the `Permission` enum into a class and creating constants for the static roles, however that would be a breaking change. i could probably introduce a new class to work around this but don't have a lot of great names, `RolePerm` perhaps?

thoughts?

bitwiseman:
`RolePermission` seems fine.  Go for it and we can adjust it when it reaches PR if it seems needed.

jgangemi:
resolved by #1403

PR #1394, opened 2022-03-03
- converted GHOrganization.Permission to a class

jgangemi:
potential fix for #1377.

i just converted the existing `enum` to a `class`. this only required minimal changes, so hopefully we can just roll w/ this change.

i'll fix the ci issues once it's decided if this is acceptable :)


jgangemi:
@bitwiseman

src/main/java/org/kohsuke/github/GHOrganization.java
```diff
@@ -448,10 +448,26 @@ public GHProject createProject(String name, String body) throws IOException {
     }
 
     /**
-     * The enum Permission.
+     * Permissions (roles) for teams and collaborators.
      */
-    public enum Permission {
-        ADMIN, MAINTAIN, PUSH, TRIAGE, PULL
+    public static class Permission {
+        // pre-canned permissions/roles
```
bitwiseman, line 454: @jgangemi 

Sorry, no. We can't change the existing email to a class as that is a binary breaking change. We can create a new class but the existing one has to stay. 

We can add an `UNKOWN` value to the enum and add a `getPermissionString()` method to allow people to get the raw value. 

But we're not ready to make a breaking change at this time.

bitwiseman (changes requested):
Can't make a breaking change like this.

src/main/java/org/kohsuke/github/GHOrganization.java
jgangemi: mm - right. forgot about that. all good, i'll do something else here.

Issue #1386, opened 2022-03-01
Update GHTeam APIs to not use deprecated urls - 15th March effective


AmyShields-EN0085:
GHTeam APIs updated to not use deprecated urls - comes into effect 22nd March

Communication from GitHub -

https://github.blog/changelog/2022-02-22-sunset-notice-deprecated-teams-api-endpoints/

In [early 2020](https://developer.github.com/changes/2020-01-21-moving-the-team-api-endpoints/), we moved the Teams API from a top-level path under /teams/:team_id to a scoped path under the organization that owns the team like /organizations/:org_id/team/:team_id and added support to make the API available under a named path like /orgs/:org/teams/:team_slug.

Over the next few weeks, we will be sunsetting the API endpoints under /teams/:team_id.

**Removal
March 15, 2022 at 14:00 UTC**

**Affected endpoints**
GET /teams/:team_id
PATCH /teams/:team_id
DELETE /teams/:team_id
GET /teams/:team_id/teams
GET /teams/:team_id/repos
GET /teams/:team_id/repos/:owner/:repo
PUT /teams/:team_id/repos/:owner/:repo
DELETE /teams/:team_id/repos/:owner/:repo
GET /teams/:team_id/projects
GET /teams/:team_id/projects/:project_id
PUT /teams/:team_id/projects/:project_id
DELETE /teams/:team_id/projects/:project_id

**Steps to take**
The easiest way to update your code is to switch to the /organizations/:org_id/team/:team_id path. If you're unsure of what org_id to use, you can GET the current team under /teams/:team_id and find the id field inside the organization hash in the response.

gsmet:
I'll have a look at this one tonight.

Thanks @AmyShields-EN0085 for opening the issue.

AmyShields-EN0085:
This comes into effect 15th march - not 22nd! Apologies!

gsmet:
Even better :).

gsmet:
@bitwiseman I started working on this one but I don't have the permissions to properly regenerate the required snapshots. Most of the Teams API requires me to be an admin.

bitwiseman:
@gsmet I'll get you those tonight.

bitwiseman:
@gsmet You are already an owner in `hub4j-test-org`. 

I'm a little concerned.  If there are versions of GitHub Enterprise still using these old endpoints, they will stop working with this change.  But that is lower priority bug that we can address if it comes up.

gsmet:
[...]

PR #1395, opened 2022-03-04
Switch to the new Teams API

gsmet:
https://developer.github.com/changes/2020-01-21-moving-the-team-api-endpoints/
https://github.blog/changelog/2022-02-22-sunset-notice-deprecated-teams-api-endpoints/

Fixes https://github.com/hub4j/github-api/issues/1386


src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -873,23 +873,6 @@ public GHLicense getLicense(String key) throws IOException {
...
-     * @throws IOException
-     *             the io exception
-     *
-     * @deprecated Use {@link GHOrganization#getTeam(long)}
-     * @see <a href= "https://developer.github.com/v3/teams/#get-team-legacy">deprecation notice</a>
-     */
-    @Deprecated
-    public GHTeam getTeam(int id) throws IOException {
-        return createRequest().withUrlPath("/teams/" + id).fetch(GHTeam.class).wrapUp(this);
-    }
```
gsmet, line 891: I decided to remove this method as even if it could work for old GH Entreprise instances, you would end up with a dead end and an object that can't do anything useful.

I can adjust if you prefer I keep it and throw an exception or if you prefer keeping it altogether.

gsmet (commented):
@bitwiseman this PR is time sensitive as GitHub will do another brownout on March 8th (see https://github.blog/changelog/2022-02-22-sunset-notice-deprecated-teams-api-endpoints/).

It's probably a good idea to get something released before that.

src/main/java/org/kohsuke/github/GitHub.java
bitwiseman: We must preserve API compatibility.  Feel free to have it throw if needed.

bitwiseman:
@gsmet 
We need to know if there are "supported" (by GitHub) versions of GHE that still use the old endpoints.  If so, then we need to add fallback behavior that tries the new endpoints and then falls back to the old endpoints if they get an exception. 

First step is to determine what is the oldest version of GHE that supports the new endpoints.

gsmet:
My assumption was that they would continue to use older versions of the GitHub API.
I doubt we will be able to get this information, except if you have contacts at GitHub but my guess is it will take time and time is not our friend in this case.

I can try to maintain the old behavior but I won't be able to test it though.

bitwiseman:
Wow. It looks like they have gotten aggressive about deprecation. 3.1 is the oldest currently supported version. 

https://docs.github.com/en/enterprise-server@3.4/admin/all-releases

gsmet:
AFAICS, they already removed the old teams endpoint in 3.1.

bitwiseman:
So, in short, all we need is to revert the method removal and we're good.

gsmet:
Yeah I'm doing that as we speak.

gsmet:
@bitwiseman done!

bitwiseman (approved):

gsmet:
@bitwiseman do you plan a new release before the next brownout of March 8th?

Personally I don't use this API but some users asked for it.

PR #1423, opened 2022-04-09
Switch to the new delete reaction API

gsmet:
The old delete reaction API has been deprecated for two years and
removed in February 2022.

See https://github.blog/changelog/2022-02-11-legacy-delete-reactions-rest-api-removed/

@bitwiseman  It's kinda in the way of some of my work so if you have a release planned, it would be nice to have this one included.


gsmet:
Forgot `forRemoval` was JDK 11+.

gsmet:
So the issue with code coverage is that reaction methods are now specific to each `Reactable` and reactions were not tested for some of them.
Also the `GHReaction` method is obviously not tested anymore because it's not working anymore.

gsmet:
I pushed some additional tests to improve coverage. The only thing that should be tested less now is `GHReaction` where the non working method is.

bitwiseman:
@gsmet 

> The old delete reaction API has been deprecated for two years and removed in February 2022.
> 
> See https://github.blog/changelog/2022-02-11-legacy-delete-reactions-rest-api-removed/
> 
> @bitwiseman It's kinda in the way of some of my work so if you have a release planned, it would be nice to have this one included.

If it really doesn't work, then maybe we should make the old method throw an exception? Then you can add a test that ensures it throws the exception. 

We follow semver, which to my understanding guarantees binary compatibility. We certainly don't pledge to maintain functionality of long deprecated endpoints.

gsmet:
@bitwiseman this one should be ready to go now. I throw an exception in the old method and it's tested.

bitwiseman (approved):

PR #1433, opened 2022-04-18
Add GHRepository#hasPermission()

gsmet:
@bitwiseman this is a proposal to add a `GHRepository#hasPermission(user, permission)` method.

The rationale of it is that most often than not you want to check that someone has **at least** a given level and not a specific permission.
The current API makes it a bit cumbersome to do it.

The added test explains it all.


gsmet:
BTW, if you can include it in 1.304, that would be awesome!

PR #1437, opened 2022-04-20
Update GHTeam.java

qc12su:
# Description
The remove function is using the wrong path
https://docs.github.com/en/rest/teams/members#remove-team-membership-for-a-user-legacy

# Before submitting a PR:

- [x] Changes must not break binary backwards compatibility. If you are unclear on how to make the change you think is needed while maintaining backward compatibility, [CONTRIBUTING.md](CONTRIBUTING.md) for details.
- [x] Add JavaDocs and other comments as appropriate. Consider including links in comments to relevant documentation on https://docs.github.com/en/rest .  
- [x] Add tests that cover any added or changed code. This generally requires capturing snapshot test data. See [CONTRIBUTING.md](CONTRIBUTING.md) for details.
- [x] Run `mvn -D enable-ci clean install site` locally. If this command doesn't succeed, your change will not pass CI.
- [x] Push your changes to a branch other than `main`. You will create your PR from that branch.

# When creating a PR: 

- [x] Fill in the "Description" above with clear summary of the changes. This includes:
  - [x] If this PR fixes one or more issues, include "Fixes #<issue number>" lines for each issue. 
  - [x] Provide links to relevant documentation on https://docs.github.com/en/rest where possible.
- [x] All lines of new code should be covered by tests as reported by code coverage. Any lines that are not covered must have PR comments explaining why they cannot be covered. For example, "Reaching this particular exception is hard and is not a particular common scenario."
- [x] Enable "Allow edits from maintainers".

PR #1438, opened 2022-04-21
Fix GHTeam#remove(GHUser) and add tests

gsmet:
I included the original commit of https://github.com/hub4j/github-api/pull/1437 with the original author (just reworded the commit message for more clarity).

My advice would be to release something right away as this could be considered as a security issue given we are not removing people from teams appropriately.

Note that it was broken anyway since GH retired the old Teams API.


src/main/java/org/kohsuke/github/GHTeam.java
```diff
@@ -151,6 +152,19 @@ public PagedIterable<GHUser> listMembers(String role) throws IOException {
...
+     *
+     * @param role
+     *            the role
+     * @return the paged iterable
+     * @throws IOException
+     *             the io exception
+     */
+    public PagedIterable<GHUser> listMembers(Role role) throws IOException {
+        return listMembers(role.name().toLowerCase(Locale.ROOT));
+    }
```
gsmet, line 166: Added this in passing as I wanted to test this and it's more practical with the enum rather than the string.

bitwiseman: Modifed to use `transformEnum()`, but otherwise 👍 .

gsmet:
Thanks for the quick release @bitwiseman !

PR #1455, opened 2022-05-12
Use send() to execute requests instead of fetchHttpStatusCode()

### LINKED ISSUE #1380: Determining result of Workflow Dispatch
### Unable to identify the cause of failure for workflow dispatch event

image

As we known, `dispatch` doesn't return anything, making it extremely difficult to identify the cause of failure for a workflow trigger.
As an example, assume a workflow requires a specific input which is not provided in the call - the method has no way of providing us with such information on failure. 

Do we have a way to identify at minimum the status code returned from the dispatch?

gsmet:
Fixes #1380

Note: I had to take a new snapshot for GHWorkflowTest as the dispatch
test was silently failing before (we were using an input that was not
declared in the workflow, I fixed that in the test workflow).


gsmet:
No idea how the code coverage can be lower given the changes I made :).

bitwiseman (approved):

PR #1462, opened 2022-05-30
Add support for workflow_job event payload

### LINKED ISSUE #1331: Implement workflow job event payload
**Describe the bug**
#1316 added `WORKFLOW_JOB` to the GHEvent enum, but receiving workflow_job events is still not fully supported because it can’t be mapped to any GHEventPayload subclass.  An attempt to map it to the GHEventPayload superclass results in:
> com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `org.kohsuke.github.GHEventPayload` (no Creators, like default constructor, exist)

I’ve captured the request body for a workflow_job event in order to check its structure, and it looks like implementing this is very straightforward.  I used the WorkflowRun subclass as a model.
```
diff --git a/src/main/java/org/kohsuke/github/GHEventPayload.java b/src/main/java/org/kohsuke/github/GHEventPayload.java
index 4d47264aa..ca3b75036 100644
--- a/src/main/java/org/kohsuke/github/GHEventPayload.java
+++ b/src/main/java/org/kohsuke/github/GHEventPayload.java
@@ -1446,6 +1446,42 @@ public abstract class GHEventPayload extends GitHubInteractiveObject {
         }
     }
 
+    /**
+     * A workflow job has been queued, is in progress, or has been completed.
+     *
+     * @see <a href=
+     *      "https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#workflow_job">
+     *      workflow job event</a>
+     * @see <a href="https://docs.github.com/en/rest/reference/actions#workflow-jobs">Actions Workflow Jobs</a>
+     */
+    public static class WorkflowJob extends GHEventPayload {
+        private GHWorkflowJob workflowJob;
+
+        /**
+         * Gets the workflow job.
+         *
+         * @return the workflow job
+         */
+        @SuppressFBWarnings(value = { "EI_EXPOSE_REP" }, justification = "Expected")
+        public GHWorkflowJob getWorkflowJob() {
+            return workflowJob;
+        }
+
+        @Override
+        void lateBind() {
+            if (workflowJob == null) {
+                throw new IllegalStateException(
+                        "Expected workflow_job payload, but got something else.  Maybe we've got another type of event?");
+            }
+            super.lateBind();
+            GHRepository repository = getRepository();
+            if (repository == null) {
+                throw new IllegalStateException("Repository must not be null");
+            }
+            workflowJob.wrapUp(repository);
+        }
+    }
+
     /**
      * A label was created, edited or deleted.
      *
```

I’ve tested this locally by posting a copy of the workflow_job event to my application’s webhook endpoint, where the application is doing:
```
        Class<? extends GHEventPayload> eventClass;
        switch (eventName.toLowerCase()) {

            …

            case "workflow_job":
                eventClass = GHEventPayload.WorkflowJob.class; break;

            default:
                throw new UnsupportedOperationException(
                        "Event \"" + eventName + "\" is not handled");

        }

        GHEventPayload eventPayload = GitHub.offline().parseEventPayload(
                new StringReader(payload), eventClass);
```

gsmet:
Fixes #1331


gsmet:
What's not tested in `GHEventPayload` is the IllegalStateException branches. The rest is properly covered.

gsmet:
@bitwiseman any chance we could get this one in? The loss in coverage is just caused by the edge cases of `IllegalStateException`.

PR #1490, opened 2022-07-18
Added listGists method

### LINKED ISSUE #1325: Accessing Gists via getMyself().listGists() returns null data for Gist files

EasyG0ing1:
# Description

Currently, when pulling down a list of GHGist objects, using
```Java
gitHub.getMyself().listGists().toList()
```
The GHGist objects returned in the list do not contain the contents of the files within each Gist, and there is no other method from which to pull down all gists with complete files in a single list. The reason for this seems to be that the GitHub API URL path that is used to pull the list from, intentionally does not include the contents of the file. The API docs seem to indicate a conservative data yield for that particular path, requiring a developer to then use another URL path from which to pull the individual Gists, which would then include the contents of the file.

I added this method to the GitHub class, which I have tested and verified that the list it returns is a complete list of the user's GHGist objects, where each file within the Gists has all of the data associated with the file, including content.

Fixes #1325 

The only issue that came from ```mvn -D enable-ci clean install site``` was a warning about inconsistent synchronization, but it didn't just flag my code, it flagged other code in the class as well. It also indicates that the code might be just fine and that it could be a false positive due to the nature of the test. 

The method is relatively simple, I cannot see where it would cause any issues. However, I'm willing to make whatever changes are necessary.

# Before submitting a PR:

- [X] Add JavaDocs and other comments as appropriate. Consider including links in comments to relevant documentation on https://docs.github.com/en/rest .  

# When creating a PR: 


src/main/java/org/kohsuke/github/GitHub.java
```diff
@@ -919,6 +920,22 @@ public GHGist getGist(String id) throws IOException {
...
+     * Gets a List of all GHGist objects
+     *
+     * @return the list
+     * @throws IOException
+     *             the io exception
+     */
+    public List<GHGist> listGists() throws IOException {
+        CopyOnWriteArrayList<GHGist> list = new CopyOnWriteArrayList<>();
+        assert myself != null;
+        for (GHGist ghGist : myself.listGists().toList()) {
```
bitwiseman, line 933:
```suggestion
        for (GHGist ghGist : getMyself().listGists().toList()) {
```

bitwiseman (changes requested):
Please add at least one test for your new method.

EasyG0ing1:
> Please add at least one test for your new method.

The test will need to use an account that has Gists and files in the Gists in order to work. I wrote a test that authenticates with my own token and it works, but I don't know how to write the test so that it will always work regardless of who is running it ... since it needs a valid account that has Gists in it.

bitwiseman:
@EasyG0ing1 

You can record a snapshot of the tests and then add the `snapshotNotAllowed();` method to prevent it from being overwritten later. 

However, I'm going o make another suggestion.  Instead of adding a new method to get the list Gists, I suggest you add a `populate()` method to `GHGist` that calls the URL that returns the full contents similar to other `populate()` methods:

https://github.com/hub4j/github-api/blob/bcb71a3c318d243c92b757310cf153b8afc983d5/src/main/java/org/kohsuke/github/GHLicense.java#L196-L19

That will allow people to get the file contents if needed while preserving the incremental loading.   The test will also be simpler. You can add a step at the end of the existing test for listing Gists.

PR #1492, opened 2022-07-27
Add support for repositories body parameter when creating installation access tokens

dwnusbaum:
# Description

This PR adds support for the `repositories` body parameter as specified in https://docs.github.com/en/rest/apps/apps#create-an-installation-access-token-for-an-app to `GHAppCreateTokenBuilder`.

Originally, if you wanted to generate an installation access token limited to a specific repository, the `/app/installations/{installation_id}/access_tokens` endpoint required you to pass a `repository_ids` body parameter. That parameter is a bit awkward because generally you will need to make an additional API call to retrieve the ID of a repository given its name.

At some point (I could not find a changelog) GitHub added a `repositories` body parameter that works like `repository_ids`, but accepts repository names rather than IDs, which is a nice improvement.

# Before submitting a PR:

# When creating a PR: 


src/main/java/org/kohsuke/github/GHAppCreateTokenBuilder.java
```diff
@@ -46,6 +46,20 @@ public GHAppCreateTokenBuilder repositoryIds(List<Long> repositoryIds) {
...
 
+    /**
+     * By default the installation token has access to all repositories that the installation can access. To restrict
+     * the access to specific repositories, you can provide repository names when creating the token.
+     *
+     * @param repositories
+     *            Array containing the repository names
+     * @return a GHAppCreateTokenBuilder
+     */
+    @BetaApi
```
dwnusbaum, line 57: From a bit of reading I think this API is out of beta now, but I added this annotation anyways to match the other methods in this class. Let me know if I should remove the annotation or anything.

src/test/java/org/kohsuke/github/GHAppTest.java
```diff
@@ -146,6 +146,27 @@ public void createToken() throws IOException {
         assertThat(installationToken2.getRepositories(), nullValue());;
     }
 
+    @Test
+    public void createTokenWithRepositories() throws IOException {
+        GHApp app = gitHub.getApp();
+        GHAppInstallation installation = app.getInstallationByUser("bogus");
```
dwnusbaum, line 152: IDK if this is the preferred approach, but it matches the above `createToken` test (I ran it against a real app and everything though when generating the stubs and then just changed the names to match the other test).

bitwiseman (approved):

bitwiseman (approved):

PR #1502, opened 2022-08-10
Add ProjectsV2Item event payload

gsmet:
Frankly, this thing is underdocumented and IMHO not very usable but I
suppose it will improve over time.
Someone asked for it in my Quarkus GitHub App project so I thought I
might as well implement it.

@bitwiseman I think you relaxed the rules about documenting the simple getters but I'll add javadoc if required.

### DIFF
--- a/src/main/java/org/kohsuke/github/GHEventPayload.java
+++ b/src/main/java/org/kohsuke/github/GHEventPayload.java
@@ -1630,4 +1630,26 @@ public abstract class GHEventPayload extends GitHubInteractiveObject {
             return GitHubClient.parseDate(starredAt);
         }
     }
+
+    /**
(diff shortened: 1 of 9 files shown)


bitwiseman:
@gsmet 
> Frankly, this thing is underdocumented and IMHO not very usable but I suppose it will improve over time. 

I suppose you mean the underlying API, not this PR, right?

bitwiseman (approved):

gsmet:
Yes. The PR is fine given what’s in the payload. We can’t do much better. 

There might be some things that are not covered as I had to experiment to understand what were in the changes. But it should be resilient enough to not break in this case.

bitwiseman (approved):

PR #1522, opened 2022-09-18
Add polymorphic deserialization for event account.

### LINKED ISSUE #1497: GHUser even when event comes from Organization type
**Describe the bug**
A `GHUser` type is returned even when in the payload the account type is `Organization`.

More precisely in the event `installation.created` when the `installation.account.type` is set to `Organization` (app installed in organization) the `installation.account` in the `@Installation.Created` payload returns a `GHUser`.

```yaml
{
  "action": "created",
  "installation": {
    "id": 12345,
    "account": {
      "login": "org-name",
      <...>
      "type": "Organization",
      "site_admin": false
    },
<...>
```

```kotlin
fun onInstall(@Installation.Created installationPayload: GHEventPayload.Installation) {
  val tenantGHOrganization = installationPayload.installation.account // <-- GHUser instance
}
```

Also tried to use the `installationPayload.organization` object but it always returned null (probably it's used in other events):

```kotlin
fun onInstall(@Installation.Created installationPayload: GHEventPayload.Installation) {
  LOG.info("org: " + installationPayload.organization.name) // <--  Cannot invoke "org.kohsuke.github.GHOrganization.getName()" because the return value of "org.kohsuke.github.GHEventPayload$Installation.getOrganization()" is null
}
```

**To Reproduce**
1. Create an event with `installation.created` in an organization (install app in an organization)
2. Catch the event
3. Retrieve the `installationPayload.installation.account` object
4. Or retrieve the `installationPayload.organization` object

**Expected behavior**
If the payload has the `installation.account.type` set to `Organization` the `installationPayload.installation.account`should return a GHOrganization object or the `installationPayload.organization` should not return null.

 - OS: macOS Monterey v12.4 (Apple M1)

**Additional context**
Using Kotlin and [Quarkus GitHub App](https://github.com/quarkiverse/quarkus-github-app)

bloslo:
# Description

Add polymorphic deserialization for the `GHAppInstallation` `account` field.
Fixes #1497

# Before submitting a PR:

# When creating a PR: 


bloslo:
@bitwiseman Can you review my changes? I know checks are breaking, but this is because I change a method signature. I don't see how this can be avoided.

P.S. I used the StackOverflow link you referenced in the issue as a source to implement the changes.

bitwiseman:
@bloslo 
To address API breaks, you'll need to use `WithBridgeMethods` annotations. They preserve binary compatibility while letting you update the source API. Usage is described in the [source file](https://github.com/infradna/bridge-method-injector/blob/master/annotation/src/main/java/com/infradna/tool/bridge_method_injector/WithBridgeMethods.java) and there are plenty of examples in this project.

bitwiseman (changes requested):
Changes look good, nicely done! 

Add bridge methods to fix API breaks. The bridge will need to be a custom conversion method. And you'll need to add a test that makes sure the scenario where someone calls the old `GHUser` method when the instance is now a`GHOrganization` still works the same as before. You don't have to improve the behavior, just make sure it isn't any more broken.

Haarolean:
@bitwiseman is there a chance to get this merged in?

bitwiseman:
@Haarolean 
See my previous comment.   There are API breaking changes that need to be addressed (using bridge methods) before this can be merged.

anujhydrabadi:
@bitwiseman any chance we can have this in the 2.x release? I'm happy to recover this PR and re-test or re-implement if needed, but looks complete otherwise, as we do not need to worry about backward compatibility.

bitwiseman:
@anujhydrabadi 
Please proceed.

PR #1599, opened 2023-01-11
Add logs in GitHubClient response. fixes #1594

### LINKED ISSUE #1594: Improve logs
It would be great for debugging purpose if we can have logs about the GHResponse payload.
Currently we have the GHRequest URL (without params 😢) but nothing about the response.

c3p0-maif:
# Description

This PR adds a log in order to debug the GitHub response.
I add the hash of the request to the begginning of the log, in order to identify which response corresponds to which request.


bitwiseman (approved):
@c3p0-maif
We generally require tests coverage of new code. In this case, there's no test covering the existing code... But there are tests that use lockers to check the output of certain actions. Would you be willing to write a test for this?

PR #1625, opened 2023-03-08
GHRepository: allows creating deploy keys being read-only

### LINKED ISSUE #1592: Enhancement: GHRepository addDeployKey - Possibility to add a key as read only
Hi,
it would be nice to be able to add deploy keys as read only. This would require sending the boolean `read_only = true` to the github api in addition to the title and ssh key.

[GitHub API Doc - Create a deploy key](https://docs.github.com/de/rest/deploy-keys?apiVersion=2022-11-28)

MarcelHB:
# Description

See request in #1592, this PR adds the ability to create a deploy key set to read-only. I've left the previous `addDeployKey` as-is and added an override.

Related API Documentation, but it's trivial: https://docs.github.com/en/rest/deploy-keys?apiVersion=2022-11-28#create-a-deploy-key

fixes #1592


bitwiseman (approved):
Solid work, thanks! 
I used your test data to also test the old `addDeloyKey` method.

Issue #1613, opened 2023-02-02
Feature Request - GET /marketplace_listing/accounts/{account_id}


blacelle:
Feature Request: GET /marketplace_listing/accounts/{account_id}

https://docs.github.com/en/rest/apps/marketplace?apiVersion=2022-11-28#get-a-subscription-plan-for-an-account

I'd like to request this API which seems to be missing in the latest version.

Would this be a relevant contribution for a new-comer ?

blacelle:
By the way, I get some exception is existing related API:

com.fasterxml.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type `org.kohsuke.github.GHMarketplacePriceModel` from String "PER_UNIT": not one of the values accepted for Enum class: [flat-rate, per-unit, free]

``` json
[
  {
    "url": "https://api.github.com/marketplace_listing/plans/5538",
    "accounts_url": "https://api.github.com/marketplace_listing/plans/5538/accounts",
    "id": 5538,
    "number": 1,
    "name": "Free",
    "description": "Free account",
    "monthly_price_in_cents": 0,
    "yearly_price_in_cents": 0,
    "price_model": "FREE",
    "has_free_trial": false,
    "unit_name": null,
    "state": "published",
    "bullets": [
      "Free for OpenSource repositories"
    ]
  },
  {
    "url": "https://api.github.com/marketplace_listing/plans/6696",
    "accounts_url": "https://api.github.com/marketplace_listing/plans/6696/accounts",
    "id": 6696,
    "number": 3,
    "name": "Private Repositories",
    "description": "CleanThat for private repositories",
    "monthly_price_in_cents": 1,
    "yearly_price_in_cents": 10,
    "price_model": "PER_UNIT",
    "has_free_trial": true,
    "unit_name": "user",
    "state": "retired",
    "bullets": []
  },
  {
    "url": "https://api.github.com/marketplace_listing/plans/8362",
    "accounts_url": "https://api.github.com/marketplace_listing/plans/8362/accounts",
    "id": 8362,
    "number": 4,
    "name": "Personal Public Repositories",
    "description": "Small Fee for personal public repositories",
    "monthly_price_in_cents": 1,
    "yearly_price_in_cents": 10,
    "price_model": "FLAT_RATE",
    "has_free_trial": false,
    "unit_name": null,
    "state": "published",
    "bullets": [
      "Side projects: Keep it clean",
      "Student: Learn faster"
    ]
  }
]
```

[...]

PR #1630, opened 2023-03-19
Enable fetch plan for current app and account

blacelle:
# Description

Fixes https://github.com/hub4j/github-api/issues/1613

This enables accessing the plan with current GithubApp on current GHInstallation.

# Before submitting a PR:

# When creating a PR: 


pom.xml
```diff
@@ -600,6 +600,12 @@
       <version>2.0.3</version>
       <scope>test</scope>
     </dependency>
+	<dependency>
```
blacelle, line 603: Used to facilitate loading a custom private JWK for a GHApp

src/test/java/org/kohsuke/github/AbstractGHAppInstallationTest.java
```diff
@@ -13,17 +16,25 @@
...
+import java.util.List;
+import java.util.Set;
 
 // TODO: Auto-generated Javadoc
 /**
  * The Class AbstractGHAppInstallationTest.
  */
 public class AbstractGHAppInstallationTest extends AbstractGitHubWireMockTest {
 
+    private static String ENV_GITHUB_APP_ID = "GITHUB_APP_ID";
```
blacelle, line 33: The default GHApps have no marketplace listing. Hence, I had to rely on my own app listing.

bitwiseman:
It looks like you're still working on this.  Setting to `draft` for now.

pom.xml
blacelle: I considered relying on jjwt, but the feature does not exist yet into there, and people points back to nimbus-jose: https://github.com/jwtk/jjwt/issues/236#issuecomment-640715042

I would have to go through the JWK to PEM procedure, which is a burden given my own setup (i.e. my app using this lib relies on a JWK, not a PEM). Also, it would be more error-prone to rely on a given PEM content through environment variable (which is again my setup, and how I suggest to manage custom privateKey here).

I tried copying some classes from Nimbus to fill the gap, but it seems to require quite a bunch of classes. I'm having another try but focusing on the minimal set of methods to get a `PrivateKey`.

My only justification for this additional lib is existing tests for GitHub App APIs has not dynamic mechanisms for now (neither from local file, or from Env). I may then just drop this code, and be potentially discouraged to contribute here (this is a weak argument as I would be happy to fit into any existing mechanism, the current option would be to copy my own privateKey as a test resources, with a risk of pushing it publicly).

blacelle: I'm dropping nimbus, with a mechanism to load a PEM from a `Path` provided as environment variable (similarly to the pushed code). However, I will not be a user of this mechanism. It will pave the way for next contribution requiring a custom JWK.

src/test/java/org/kohsuke/github/AbstractGHAppInstallationTest.java
```diff
@@ -89,17 +114,28 @@ private String createJwtToken(String keyFileResouceName, String appId) {
...
-                .listInstallations()
-                .toList()
-                .stream()
-                .filter(it -> it.getAccount().login.equals("hub4j-test-org"))
-                .findFirst()
-                .get();
+        GHApp app = gitHub.getApp();
+
+        GHAppInstallation appInstallation;
+        if (Set.of(TEST_APP_ID_1, TEST_APP_ID_2, TEST_APP_ID_3).contains(Long.toString(app.getId()))) {
```
blacelle, line 128: This switch leads to invalid recordings: I've done it to prevent recording the N installations of my real-world app. However, when re-running the tests without my custom JWK, the test will list of installations, which will match no stub (as we recorded the other branch of the `if`, which was focusing on a dedicated repo.)

This remain useful, however one has to edit the recordings manually (e.g. copy-pasting the installations listing from another test).

@bitwiseman Would you mind granting me access to the test repo/app ? It may be useful on related matters not requiring markplace-listing.

blacelle:
@bitwiseman This is ready for review/merging.

.java-version
```diff
@@ -0,0 +1 @@
+11.0
```
bitwiseman, line 1: What's this?

src/main/java/org/kohsuke/github/GHMarketplaceAccount.java
```diff
@@ -72,4 +73,25 @@ public GHMarketplaceAccountType getType() {
...
+     *
+     * @return a GHMarketplaceListAccountBuilder instance
+     * @throws IOException
+     *             in case of {@link IOException}
+     * @see <a href=
+     *      "https://docs.github.com/en/rest/apps/marketplace?apiVersion=2022-11-28#get-a-subscription-plan-for-an-account">Get
+     *      a subscription plan for an account</a>
+     */
+    public GHMarketplaceAccountPlan getPlan() throws IOException {
+        return new GHMarketplacePlanForAccountBuilder(root(), this.id).createRequest();
```
bitwiseman, line 94: @blacelle  Code coverage doesn't believe this line is covered by tests.  Please update to have it covered.  If it is easier to manually create data to cover this, that is fine too.  It just needs to be exercised.

bitwiseman (changes requested):
Missing coverage for one method. Fix that and we can merge.

.java-version
blacelle: Sorry, it is a file added by standard macos tooling (`jenv`) to manage jdk by directory. I'll remove it.

src/main/java/org/kohsuke/github/GHMarketplaceAccount.java
blacelle: Latest commit should fix this. No additional data seemed require, supposedly because the underlying API call was done elsewhere.

bitwiseman (approved):

bitwiseman:
Thanks!

PR #1638, opened 2023-04-06
Applied Refactoring 

BadhriNadh:
# Description

1. Variable 'v' does not have any meaning. So, renamed as 'includeReadNotifications',  'restrictedToParticipatingNotifications' and  ‘pollIntervalHeaderValue’.
2. Extracted a block to new method named 'getLatestUnReturnedThread',
3. Extracted a block to new method named 'waitUntilNextPollingTime'
4. Extracted a block to new method named 'updateThreadsAndMetadataFromGitHub'
5. Moved condition to a function that returns a boolean.
6. The ‘hashcode’ method in this class hierarchy [GHUser -> GHRepository] is overridden in a subclass and is only utilized in that subclass without being used in any other classes in the hierarchy. Therefore, the implementation of the hashcode method has been moved down to the subclass where it is actually used to simplify the parent class and make the code more modular.
7. The ‘getApiRoute’ method was present in multiple subclasses, but with slight implementation differences. To address this, I used the Template Method pattern and pulled-up the method to the parent class. This eliminated code duplication, simplified the codebase, and made it easier to maintain and update the common code. <--This is fixing a potential issue-->
8. Method ‘getApiTail’ has almost the same implementation in four classes: GHBlobBuilder, GHCommitBuilder, GHCommitComment, and GHTreeBuilder. It's also causing feature envy smell with GHRepository so moved these methods to GHRepository into one method. <--This is fixing a potential issue-->

# Before submitting a PR:

# When creating a PR: 

### DIFF
--- a/src/main/java/org/kohsuke/github/GHArtifact.java
+++ b/src/main/java/org/kohsuke/github/GHArtifact.java
@@ -105,7 +105,7 @@ public class GHArtifact extends GHObject {
      *             the io exception
      */
     public void delete() throws IOException {
-        root().createRequest().method("DELETE").withUrlPath(getApiRoute()).send();
+        root().createRequest().method("DELETE").withUrlPath(getApiRoute(owner)).send();
     }
 
     /**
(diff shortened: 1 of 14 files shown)


src/main/java/org/kohsuke/github/GHObject.java
```diff
@@ -190,4 +192,16 @@ public void append(StringBuffer buffer, String fieldName, Object value, Boolean
             super.append(buffer, fieldName, value, fullDetail);
         }
     };
+
+    public String getApiRoute(GHRepository owner) {
```
bitwiseman, line 196:
```suggestion
    protected String getApiRoute(GHRepository owner) {
```

src/main/java/org/kohsuke/github/GHArtifact.java
```diff
@@ -122,15 +122,11 @@ public void delete() throws IOException {
...
     }
 
-    private String getApiRoute() {
-        if (owner == null) {
-            // Workflow runs returned from search to do not have an owner. Attempt to use url.
-            final URL url = Objects.requireNonNull(getUrl(), "Missing instance URL!");
-            return StringUtils.prependIfMissing(url.toString().replace(root().getApiUrl(), ""), "/");
-        }
+    @Override
+    protected String getApiRouteImpl() {
```
bitwiseman, line 129: I don't love this `*Impl` pattern, but it makes sense here.

bitwiseman (changes requested):
Generally approve of these refactors.  Some questions/requests.

BadhriNadh:
Need a different approach for this api

PR #1647, opened 2023-04-27
implemented create from manifest action

### LINKED ISSUE #1641: Feature Request: Get An App by Slug, Create App from Manifest
Feature Request: 

GET /apps/{app_slug}: https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#get-an-app
POST /app-manifests/{code}/conversions: https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#create-a-github-app-from-a-manifest

The basic implementation for an GHApp already exists, however only the endpoint retrieving the currently authorized App is supported.

If desired I could provide the implementation for both endpoints.

dbaur:
# Description

This pull requests implements the following API methods:

GET /apps/{app_slug}: https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#get-an-app
POST /app-manifests/{code}/conversions: https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#create-a-github-app-from-a-manifest

Fixes #1641

# Before submitting a PR:

- [ x Add JavaDocs and other comments as appropriate. Consider including links in comments to relevant documentation on https://docs.github.com/en/rest .  

# When creating a PR: 


dbaur:
Currently I have tested the given implementation manually. Before the implementation of automatic tests I would like to clarify two questions with respect to the implementation of the test cases:

* the POST /app-manifests/{code}/conversions is hard to test automatically as it requires a manual interaction that approves the App installation (see documentation for Manifest Flow linked above). While I can record the API interaction for automated testing, the {code} part of the GET request is only short-lived, so if anyone runs the tests against the actual GitHub API, the tests will fail. Is this ok?
* while I would like to implement the tests for GET /apps/{app_slug} in the existing GHAppTest class, the current class and the abstract implementation it inherits from focus solely on tests with respect to authorization as App Installation, where the implementation of the above method does not fit. Should I create a new Test class for those methods?

Best,

Daniel

bitwiseman:
> Currently I have tested the given implementation manually. Before the implementation of automatic tests I would like to clarify two questions with respect to the implementation of the test cases:
> 
> * the POST /app-manifests/{code}/conversions is hard to test automatically as it requires a manual interaction that approves the App installation (see documentation for Manifest Flow linked above). While I can record the API interaction for automated testing, the {code} part of the POST request is only short-lived, so if anyone runs the tests against the actual GitHub API, the tests will fail. Is this ok?

Yes, we have some tests that are recorded and never get run against live site for this same reason. Record your data, modify it as needed then add [snapshotNotAllowed()](https://github.com/hub4j/github-api/blob/5380f59f380747deeb0553ca0a67ee27812218da/src/test/java/org/kohsuke/github/AbstractGitHubWireMockTest.java#L160-L163) to your test method to make it clear.

> * while I would like to implement the tests for GET /apps/{app_slug} in the existing GHAppTest class, the current class and the abstract implementation it inherits from focus solely on tests with respect to authorization as App Installation, where the implementation of the above method does not fit. Should I create a new Test class for those methods?

Do what you think is best, just include comments to the code explaining the choice. I'm not that picky about how the testing gets done.

Thanks for contributing.

dbaur:
Hi,

I have implemented the Tests and stored the API snapshots and removed the draft flag from the PR.

Since the manifest flow test validates the client secret and the pem of the GitHub App, they expose the secrets and the pem to the public internet. I have therefore rotated those secrets in the App itself, however, external scanners will still detect those secrets and flag the contents as insecure.

We can either accept this as it poses no threat (secrets are rotated, app has no installations), or we modify the stored snapshots to some random string that is no longer detected by the secret scanners.

Best,

Daniel

PR #1649, opened 2023-04-30
Allowed creation of PR review comments using line number. Fixes #1645

### LINKED ISSUE #1645: Allow creation of comments using line number, rather than position
#1463 allowed you to list PR review comments with `line`. It did not allow you to create a comment using `line`. This is much easier than using position as it avoids a complex calculation.

AlkahtaniHisham:
# Description
Fixes #1645

Updated org.kohsuke.github.GHPullRequestReviewBuilder to allow creating pull request review comments by line, instead of the deprecated parameter position.

There were some mistakes in my previous PR so I redid it.

# Before submitting a PR:

# When creating a PR: 


alexec (commented):
I think you’ll need to support both line number and position, for backwards compatibility.

They can both be nullable. So change `position` to be `Integer` and overload the `comment` method with a version that takes both `position` and `line`. 

That said, it might be good to support `start_line` and `side` too. This allows for richer multi-line comments.


https://docs.github.com/en/rest/pulls/comments?apiVersion=2022-11-28#create-a-review-comment-for-a-pull-request

AlkahtaniHisham:
Github api says
> If you use position, the line, side, start_line, and start_side parameters are not required.

AlkahtaniHisham:
github docs say
> If you use position, the line, side, start_line, and start_side parameters are not required. 
so would it work if I keep the old method (with `position`) unchanged, and add a method that accepts `line`, `start_line`, and `side`?

alexec:
yes, that'd work

alexec (approved):
LGTM

src/main/java/org/kohsuke/github/GHPullRequestReviewBuilder.java
```diff
@@ -75,19 +80,44 @@ public GHPullRequestReviewBuilder event(GHPullRequestReviewEvent event) {
...
      * @param body
-     *            The relative path to the file that necessitates a review comment.
+     *            Text of the review comment.
      * @param path
-     *            The position in the diff where you want to add a review comment. Note this value is not the same as
-     *            the line number in the file. For help finding the position value, read the note below.
+     *            The relative path to the file that necessitates a review comment.
      * @param position
-     *            Text of the review comment.
+     *            The position in the diff where you want to add a review comment. Note this value is not the same as
```
bitwiseman, line 87: Instead of saying what this is not, describe what it is.

src/main/java/org/kohsuke/github/GHPullRequestReviewBuilder.java
```diff
@@ -103,15 +133,35 @@ public GHPullRequestReview create() throws IOException {
...
         private String body;
         private String path;
-        private int position;
+        private Integer position;
+        private Integer line;
+        private Integer start_line;
+        private String side;
 
-        DraftReviewComment(String body, String path, int position) {
+        DraftReviewComment(String body, String path, Integer position) {
```
bitwiseman, line 145: For this constructor, the `position` is required and not allowed to be `null`, correct? 
```suggestion
        DraftReviewComment(String body, String path, int position) {
```

bitwiseman (changes requested):
Looks reasonable, minor suggestions. 
Add tests to exercise the new method.

src/test/java/org/kohsuke/github/GHPullRequestTest.java
```diff
@@ -207,7 +207,8 @@ public void pullRequestReviews() throws Exception {
         GHPullRequest p = getRepository().createPullRequest(name, "test/stable", "main", "## test");
         GHPullRequestReview draftReview = p.createReview()
                 .body("Some draft review")
-                .comment("Some niggle", "README.md", 1)
+                .comment("Some niggle", "README.md", 2)
+                .comment("other niggle", "README.md", 1, null, null)
```
bitwiseman, line 211: You'll need to update the test data for these new interactions.

alexec (approved):
Was there some kind of merge conflict with all the `.json` files?

I never said this, but the use case I have is making multi line suggestions.

bitwiseman:
@alexec @AlkahtaniHisham 
Could you record new test data?

alexec:
LGTM

bitwiseman:
@AlkahtaniHisham We need new test data.

gilday:
@bitwiseman I can help record the test data if needed. I understand I would need access to hub4j-test-org to record and check-in interactions with that account.

bitwiseman:
@gilday Invitation sent for the test org.

PR #1654, opened 2023-05-01
Ignoring Unnecessarily generated japicmp direcory

java-codehunger:
# Description

In our analysis, we observe that this target/japicmp/ directory is generated but not later used during the CI build. Hence, we propose to disable generating this directory to save runtime.

The generation of this directory can be disabled by simply adding -Djapicmp.skip to the mvn command.

### DIFF
--- a/.github/workflows/maven-build.yml
+++ b/.github/workflows/maven-build.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Maven Install (skipTests)
         env:
           MAVEN_OPTS: ${{ env.JAVA_11_PLUS_MAVEN_OPTS }}
-        run: mvn -B clean install -DskipTests --file pom.xml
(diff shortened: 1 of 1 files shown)


bitwiseman:
We do use `japicmp`, so no we do not want to skip that, sorry. 
Also, generation of extra folders or files under the `target` directory is not really a problem, is it?

java-codehunger:
Because generating the unnecessary directories/files takes time to finish. So, if we remove the unnecessary directory, that will save run-time. One more thing, in our analysis, we observe that this directory is not uploaded once the CI build finishes its run. Are you really using this directory then?

bitwiseman:
`japicmp` is used to prevent API breaking changes.  

https://github.com/hub4j/github-api/blob/dc64592a8c06c8bacbdeddac4c2311b2968fe47b/pom.xml#L396-L412

We will not be disabling to prevent the creation of one directory under `target`.

PR #1655, opened 2023-05-03
Ignoring unnecessarily generated surefire-report

java-codehunger:
In our analysis, we observe that this target/surefire-report/ directory is generated but not later used during the CI build. Hence, we propose to disable generating this directory to save runtime.
The generation of this directory can be disabled by simply adding -Dsurefire.useFile=false -DdisableXmlReport=true to the mvn command.

### DIFF
--- a/.github/workflows/maven-build.yml
+++ b/.github/workflows/maven-build.yml
@@ -74,7 +74,7 @@ jobs:
         cache: 'maven'
     # JDK 8
     - name: Maven Install with Code Coverage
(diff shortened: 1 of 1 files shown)


bitwiseman:
@java-codehunger 
Why do we care about this folder being created?

bitwiseman:
Adding complexity to the command-line call is not an acceptable solution.  Updating the `pom.xml` would be the way to go for this.  However, even then tweaking build settings to achieve minor improvements in build time is a non-goal at this time, sorry. 

I hope you will consider submitting issues and PRs that improve the customer-facing features of this library.

PR #1660, opened 2023-05-15
New method to retrieve repo action variable

garridobarrera:
# Description
PR to include a new method to retrieve repo action variable

closes #1659

PR #1668, opened 2023-06-01
CRUD repository variable

garridobarrera:
# Description

Methods to create and update repository variables:

* Create repository variable: https://docs.github.com/en/rest/actions/variables?apiVersion=2022-11-28#create-a-repository-variable
* Update repository variable: https://docs.github.com/en/rest/actions/variables?apiVersion=2022-11-28#update-a-repository-variable 

# Before submitting a PR:

# When creating a PR: 


src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -2729,6 +2729,23 @@ public GHContent getReadme() throws IOException {
...
+     *             the io exception
+     */
+    public void createRepoVariable(final GHRepositoryVariable name) throws IOException {
+        root().createRequest()
+                .withUrlPath(getApiTailUrl("actions/variables"))
+                .method("POST")
+                .with("name", name.getName())
+                .with("value", name.getValue())
+                .send();
+    }
```
bitwiseman, line 2747: Please move the implementation into a static method on `GHRepositoryVariable`. This method can remain here but the meat should be on the `GHRepositoryVariable` class.  See https://github.com/hub4j/github-api/blob/c357507750f9dd28b3732a2349ca8a1db4759e42/src/main/java/org/kohsuke/github/GHRepository.java#L2383C1-L2385 for an example.  

Also, probably rename to `createVariable` since the `Repo` is implied - if there are other `variable`s to be created on `GHRepository` I'm open to discussion here.  ... Ugh. I see the problem: https://github.com/hub4j/github-api/pull/1660 . You're quite reasonably following that naming convention. 😭  Hmm. 

```suggestion
    public void createVariable(final name, value) throws IOException {
      GHRepositoryVariable.create(this).name(name).value(value).done();
    }
```

The thing that particularly makes me think the `GHLabel` pattern is the way to go is your update method below that takes a separate `name` field to allow updating the name of an existing `GHRepositoryVariable`.

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -2744,6 +2761,26 @@ public GHRepositoryVariable getRepoVariable(String name) throws IOException {
...
+    public void updateRepoVariable(final String name, final GHRepositoryVariable variableToUpdate) throws IOException {
+
+        root().createRequest()
+                .withUrlPath(getApiTailUrl("actions/variables"), name)
+                .method("PATCH")
+                .with("name", variableToUpdate.getName())
+                .with("value", variableToUpdate.getValue())
+                .send();
+    }
+
```
bitwiseman, line 2783: As noted above I think this should be moved to `GHRepositoryVariable`.   Even if you don't want to use the pattern that `GHLabel` uses, the update should be done from an instance method on `GHRepositoryVariable`.

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -2729,6 +2729,23 @@ public GHContent getReadme() throws IOException {
...
+     */
+    public void createRepoVariable(final GHRepositoryVariable name) throws IOException {
+        root().createRequest()
+                .withUrlPath(getApiTailUrl("actions/variables"))
+                .method("POST")
+                .with("name", name.getName())
+                .with("value", name.getValue())
+                .send();
+    }
+
```
bitwiseman, line 2748: Sigh, and while you're at it please deprecate the `getRepoVariable()` method?  It should either be named `getRepositoryVariable` or `getVariable`. :sob:

src/test/java/org/kohsuke/github/GHRepositoryTest.java
```diff
@@ -1594,4 +1594,34 @@ public void testRepoActionVariable() throws Exception {
...
+     * @throws IOException
+     *             the exception
+     */
+    @Test
+    public void testCreateRepoActionVariable() throws IOException {
+        GHRepository repository = getRepository();
+        GHRepositoryVariable newVariable = new GHRepositoryVariable();
+        newVariable.setName("mynewvariable");
+        newVariable.setValue("mynewvalue");
+        repository.createRepoVariable(newVariable);
```
bitwiseman, line 1610: Since both `name` and `value` are required the method signature should probably look like: 
```suggestion
        repository.createRepoVariable("mynewvariable", "mynewvalue");
```

bitwiseman (changes requested):
I'm sorry about this review.  You've done solid work here, including providing tests.  

I didn't properly think through the PR the introduced `GHRepositoryVariables`. My bad.  But now that I see there's more to them, I'd like to do this API better.   Are you up for it? 

If you're strongly opposed, I'm open to alternatives.

garridobarrera:
> I'm sorry about this review. You've done solid work here, including providing tests.
> 
> I didn't properly think through the PR the introduced `GHRepositoryVariables`. My bad. But now that I see there's more to them, I'd like to do this API better. Are you up for it?
> 
> If you're strongly opposed, I'm open to alternatives.

no worries @bitwiseman. I agree you !!

src/main/java/org/kohsuke/github/GHRepository.java
garridobarrera: I think that getVariable is better option

src/main/java/org/kohsuke/github/GHRepository.java
garridobarrera: Of course, the GHLabel pattern sounds good.

src/main/java/org/kohsuke/github/GHRepository.java
garridobarrera: Really inside GHRepository class, name a method with the word repository is redundant.

garridobarrera:
@bitwiseman, I have applied the same pattern as GHLabel.

PR #1672, opened 2023-06-05
Add MAINTAIN and TRIAGE permissions

MarkEWaite:
## Add MAINTAIN and TRIAGE permissions to permission type

- Shows #1671 Create a test demoing the issue
- Fixes #1671 by adding MAINTAIN and TRIAGE permissions to permission type

I've not included links to the GitHub REST API documentation describing those pemrissions because I've been unable to find them listed in the REST API documentation.  I'm confident that they are in the documentation, but my search and navigation skills are insufficient to find them.

Failures for MAINTAIN permission on ci.jenkins.io are documented in:

* https://github.com/jenkins-infra/helpdesk/issues/3617

Failures for TRIAGE permission on other systems are documented in:

* [Community post](https://community.jenkins.io/t/multibranch-pipline-fails-because-triage-enum-doesnt-exist/7800) - Multibranch pipline fails because TRIAGE enum doesn't exist

Special thanks to @PierreBietz for his investigation work and the test case that shows the issue.

Requesting review from @jglick and @basil in case I've missed something important in the implementation.

# Before submitting a PR:

- [x] Changes must not break binary backwards compatibility. If you are unclear on how to make the change you think is needed while maintaining backward compatibility, [CONTRIBUTING.md](CONTRIBUTING.md) for details.
- [x] Add JavaDocs and other comments explaining the behavior. 
- [x] When adding or updating methods that fetch entities, add `@link` JavaDoc entries to the relevant documentation on https://docs.github.com/en/rest . 
- [x] Add tests that cover any added or changed code. This generally requires capturing snapshot test data. See [CONTRIBUTING.md](CONTRIBUTING.md) for details.
- [x] Run `mvn -D enable-ci clean install site` locally. If this command doesn't succeed, your change will not pass CI.
- [x] Push your changes to a branch other than `main`. You will create your PR from that branch.

# When creating a PR: 

- [x] Fill in the "Description" above with clear summary of the changes. This includes:
  - [x] If this PR fixes one or more issues, include "Fixes #<issue number>" lines for each issue. 
  - [ ] Provide links to relevant documentation on https://docs.github.com/en/rest where possible. If not including links, explain why not.
- [x] All lines of new code should be covered by tests as reported by code coverage. Any lines that are not covered must have PR comments explaining why they cannot be covered. For example, "Reaching this particular exception is hard and is not a particular common scenario."
- [x] Enable "Allow edits from maintainers".

Issue #1671, opened 2023-06-05
GHRepository#getPermission fails with a user having the maintain role


PierreBtz:
**Describe the bug**
When trying to retrieve the permission of a user with the maintain role on a repository, an exception is raised:

```
java.lang.IllegalArgumentException: No enum constant org.kohsuke.github.GHPermissionType.MAINTAIN
	at java.base/java.lang.Enum.valueOf(Enum.java:240)
	at org.kohsuke.github.GHPermission.getPermissionType(GHPermission.java:55)
	at org.kohsuke.github.GHRepository.getPermission(GHRepository.java:1096)
```

**To Reproduce**

A test reproducing the issue against the `hub4j-test` organization has been created:

https://github.com/hub4j/github-api/commit/ff26f68286ec9066114a045b45ae34747bb4cb57

**Expected behavior**

The permission is properly retrieved.

**Additional context**

The naive approach would be to add `MAINTAIN` value in the `GHPermissionType` enum but it doesn't seem right to confuse roles and permissions and [the github documentation](https://docs.github.com/en/rest/collaborators/collaborators?apiVersion=2022-11-28#get-repository-permissions-for-a-user) clearly lists only `admin, write, read, and none` as valid permissions.

So I think the issue is with https://github.com/hub4j/github-api/blob/29799cde73a32f923e90a6dc3f8f139c2cd56b47/src/main/java/org/kohsuke/github/GHPermission.java#L54-L56 but I'd need help figuring things out.

stratiocommit-test:
Hi! We are facing the same error from Jenkins pipelines:
![image](https://github.com/hub4j/github-api/assets/36296135/633d5483-2bbb-492f-be67-062794ccb04a)

lpgutierrez:
Hi! We are facing the same issue from Jenkins pipelines:
![image](https://github.com/hub4j/github-api/assets/92737780/9b1a2aac-797f-4778-bfa7-36e131dd1ec9)

PierreBtz:
Yeah, fun fact I extracted this use case out of a failing Jenkins build :)

jtnord:
> The naive approach would be to add MAINTAIN value in the GHPermissionType enum but it doesn't seem right to confuse roles and permissions and [the github documentation](https://docs.github.com/en/rest/collaborators/collaborators?apiVersion=2022-11-28#get-repository-permissions-for-a-user) clearly lists only admin, write, read, and none as valid permissions.

this is the approach?

[...]

PR #1674, opened 2023-06-07
#1671 Remove unreleased MAINTAIN and TRIAGE permissions added during GitHub outage

PierreBtz:
# Description

Permissions added to the enum in #1672 do not make sense anymore since the Github Rest API change was roll-backed with no intention of re-introducing this change (as per: https://github.com/hub4j/github-api/issues/1671#issuecomment-1577515662).

In order to clean things up, this is a partial revert of #1672 only keeping the fallback logic.

I intentionally deleted the enum members instead of deprecating them, I don't really see a use case where somebody would have used the new enum values since the library was released around the same time Github rollbacked their changes.

# Before submitting a PR:

- [ ] Changes must not break binary backwards compatibility. If you are unclear on how to make the change you think is needed while maintaining backward compatibility, [CONTRIBUTING.md](CONTRIBUTING.md) for details. -> *SEE MY COMMENT above!*

# When creating a PR: 

### DIFF
--- a/src/main/java/org/kohsuke/github/GHPermissionType.java
+++ b/src/main/java/org/kohsuke/github/GHPermissionType.java
@@ -10,12 +10,8 @@ public enum GHPermissionType {
 
     /** The admin. */
     ADMIN(30),
-    /** The maintain. */
-    MAINTAIN(25),
     /** The write. */
     WRITE(20),
-    /** The triage. */
-    TRIAGE(15),
     /** The read. */
     READ(10),
     /** The none. */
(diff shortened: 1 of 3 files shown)


jglick (approved):
CC @MarkEWaite

MarkEWaite (approved):
This change looks very good to me.  

The change to add `MAINTAIN` and `TRIAGE` permissions was never released.  The `github-api-1.315` tag is placed on commit 5be032edd218791b2bf945c96e77927d995da070 May 31, 2023, while the addition of `MAINTAIN` and `TRIAGE` permissions was done on June 5, 2023 in commit 49a93bfb4a94356bbd3b794704a0437b34c42208 .  No compatibility impact because the new enumerations were never released.

MarkEWaite:
It might help the readers of the changelog if the title of the pull request specifically mentioned the permissions being removed so that they see the "add" line and then the "remove" line in the same changelog.

== Remove `MAINTAIN` and `TRIAGE` permissions added during GitHub outage

PierreBtz:
Wasn't aware that the change didn't even get released, it definitely answers my question about backward compatibility :)

I followed your suggestion and updated the title of this PR to be more helpful.

PR #1676, opened 2023-06-09
Support make_latest when creating/updating a release

### LINKED ISSUE #1595: Feature Request: GHReleaseBuilder able to uptade realese to latest
It would be great to be able to get update a release to latest. According to the GitHub REST API (https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#update-a-release) it only needs a make_latest="true" in the body of the PATCH request.

nikclayton:
# Description

Add support for the `make_latest` option when creating or updating a release (https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#create-a-release).

Fixes #1595.

# Before submitting a PR:

# When creating a PR: 


src/main/java/org/kohsuke/github/GHReleaseBuilder.java
```diff
@@ -103,6 +104,38 @@ public GHReleaseBuilder categoryName(String categoryName) {
...
+    }
+
+    /**
+     * Optional.
+     *
+     * @param latest
+     *            Whether to make this the latest release. Default is {@code TRUE}
+     * @return the gh release builder
+     */
+    public GHReleaseBuilder makeLatest(MakeLatest latest) {
```
nikclayton, line 134: Unsure how you would like to test this.

A release (as returned by the GH API) does not have a `latest` property, or anything like that, to describe whether or not it is the latest release. So I don't think the existing wiremock tests are appropriate.

You, could, perhaps, add a test that verifies that if this is option is set then the appropriate field is included in the JSON request (for create and update), but a cursory inspection didn't show other tests that did that, so I haven't done that here either.

bitwiseman (changes requested):
It looks like there is an api to the the latest release, example: `https://api.github.com/repos/hub4j/github-api/releases/latest` .

I've invited you to https://github.com/hub4j-test-org to make creating the test easier.

So, you could write a test that:
* Creates a temp repository (built-in utility)
* Creates a release
* Gets latest (should match the created release)
* Creates a new release with `make_latest` set to `false`
* Gets latest (should match the first release)
* Creates a new release with `make_latest` set to `true`
* Gets latest (should match the last created release)

src/main/java/org/kohsuke/github/GHReleaseBuilder.java
```diff
@@ -103,6 +104,38 @@ public GHReleaseBuilder categoryName(String categoryName) {
         return this;
     }
 
+    public static enum MakeLatest {
```
bitwiseman, line 107:
```suggestion
    /**
     * Values for whether this release should be the latest.
     */
    public static enum MakeLatest {
```

PR #1678, opened 2023-06-20
Add support for deleting files from tree

### LINKED ISSUE #1484: Cannot delete blob from tree
**Describe the bug**
While there is a way to create a tree (`GHTreeBuilder`) and to add an object to the tree (`GHTreeBuilder#add`), there is no clear way to remove an object (i.e. remove a file upon committing the tree + updating a ref). According to the [GitHub REST api docs](https://docs.github.com/en/rest/git/trees#create-a-tree), one should be able to remove a file by passing in a null sha. This is not possible with `GHTreeBuilder#add`, as it requires a non-null content (not sha). 

Is there something I'm missing here? Is that a separate method for this in the api?
Note: `GHRepository#getContent#delete` will not suffice. That requires me to do a commit for every single file I want to delete, which is not ideal.

**To Reproduce**
1. Create a new `GHTreeBuilder`, with a `baseTree`
2. Try to pass in a null value for content in `GHTreeBuilder#add`
3. Be sad at the compile errors

**Expected behavior**
Upon committing the tree and updating a ref, the file with a null sha passed should be deleted.

davseitsev:
Fixes #1484
Github API allows deleting files from tree by passing `null` value in `sha` field. Existing implementation skips all null fields during serialization due to `@JsonInclude(Include.NON_NULL)` annotation. This PR adds separate object for delete operation with `sha` field which is always serialized.


davseitsev:
@bitwiseman could you please take a look?

src/main/java/org/kohsuke/github/GHTreeBuilder.java
```diff
@@ -162,6 +171,21 @@ public GHTreeBuilder add(String path, String content, boolean executable) {
...
+     * Removes an entry with the given path from base tree.
+     *
+     * @param path
+     *            the file path in the tree
+     * @param executable
+     *            true, if the file should be executable
+     * @return this GHTreeBuilder
+     */
+    public GHTreeBuilder delete(String path, boolean executable) {
+        TreeEntry entry = new DeleteTreeEntry(path, executable ? "100755" : "100644", "blob");
```
bitwiseman, line 184: Is `executable` required for deletion?  Does it need to match the tree object that is being deleted?

bitwiseman: I also wonder what happens if the `type` doesn't match the item at that `path`. 
https://docs.github.com/en/rest/git/trees?apiVersion=2022-11-28#create-a-tree under `properties of tree`, shows type can be `blob`, `tree`, or `commit`.

This isn't necessarily a blocker, but it would be good to know if there are scenarios that the current `delete()` method doesn't handle.

bitwiseman (changes requested):
A few questions, but otherwise looks great.

src/main/java/org/kohsuke/github/GHTreeBuilder.java
davseitsev: > Is `executable` required for deletion? Does it need to match the tree object that is being deleted? If not (if delete works regardless of this value), we should remove the `executable` parameter here.

Yes, it's required for deletion as well and it should be not null.
Reference doc says nothing about what value for `mode` should be used for deletion but my test shows it works with any valid non-null value. So we can use some hadcoded value. I will change this

davseitsev: > I also wonder what happens if the `type` doesn't match the item at that `path`.

I tested `type=commit` while deleting a `blob` and it works fine.

davseitsev:
I've added comments and hardcoded `mode` value for delete method.
@bitwiseman let me know if I need to fix something else or squash commits in the branch.

src/main/java/org/kohsuke/github/GHTreeBuilder.java
```diff
@@ -37,6 +37,19 @@ private TreeEntry(String path, String mode, String type) {
...
+        /**
+        According to reference doc https://docs.github.com/en/rest/git/trees?apiVersion=2022-11-28#create-a-tree:
+        if sha value is null then the file will be deleted. That's why in this DTO sha is always {@literal null} and is included to json.
+         */
+        @JsonInclude
+        private final String sha = null;
+
+        private DeleteTreeEntry(String path, String mode, String type) {
+            super(path, mode, type);
+        }
```
bitwiseman, line 50:
```suggestion
        private DeleteTreeEntry(String path) {
            // The `mode` and `type` parameters are required by the API, but their values are ignored during delete.
            // Supply reasonable placeholders.
            super(path, "100644", "blob");
        }
```

bitwiseman (approved):

src/main/java/org/kohsuke/github/GHTreeBuilder.java
```diff
@@ -37,6 +37,21 @@ private TreeEntry(String path, String mode, String type) {
         }
     }
 
+    private static class DeleteTreeEntry extends TreeEntry {
+        /**
+        According to reference doc https://docs.github.com/en/rest/git/trees?apiVersion=2022-11-28#create-a-tree:
+        if sha value is null then the file will be deleted. That's why in this DTO sha is always {@literal null} and is included to json.
+         */
```
bitwiseman, line 44:
```suggestion
        /**
         * According to reference doc https://docs.github.com/en/rest/git/trees?apiVersion=2022-11-28#create-a-tree: if
         * sha value is null then the file will be deleted. That's why in this DTO sha is always {@literal null} and is
         * included to json.
         */
```

PR #1679, opened 2023-06-23
Return all files from a commit

### LINKED ISSUE #1669: GHCommit.getFiles() limited to 300 files, does not support pagination
**Describe the bug**
https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#get-a-commit

> Note: If there are more than 300 files in the commit diff, the response will include pagination link headers for the remaining files, up to a limit of 3000 files. Each page contains the static commit information, and the only changes are to the file listing.

Except `hub4j` doesn't support pagination at this time: 
https://github.com/hub4j/github-api/blob/9801b86b8aff903aaee4b7b0983a616ee646db0d/src/main/java/org/kohsuke/github/GHCommit.java#L418-L421

**To Reproduce**
1. Use `GHCommit.getFiles()` on a commit with more than 300 files changed. 

**Expected behavior**
Suggest deprecating `getFiles()` and adding a `listFile()` that returns a [`PagedIterable`](https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/PagedIterable.java).  However, I think that `PagedIterable` will need to be backed by a custom class similar to  [`GHWorkflowsIterable`](https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/GHWorkflowsIterable.java) with the `GHCommit` as the container.  It's going to take some trial-and-error to figure out how this works.

frink182:
# Description

Fixes #1669
If there are more than 300 files in the commit diff, the GitHub API response will include pagination link headers for the remaining files, up to a limit of 3000 files. Each page contains the static commit information, and the only changes are to the file listing. This PR adds `GHCommit.listFiles()` which returns the list of files in a commit via paginated responses. It deprecates the previous method to get the files: `GHCommit.getFiles()`.

The PR also populates the `message` in a `GHCommit` which was previously null.

# Before submitting a PR:

# When creating a PR: 


src/main/java/org/kohsuke/github/GHCommitIterable.java
```diff
@@ -0,0 +1,72 @@
...
+     *
+     * @param owner
+     *            the owner
+     * @param sha
+     *            the SHA of the commit
+     */
+    public GHCommitIterable(GHRepository owner, String sha) {
+        this.owner = owner;
+        this.sha = sha;
+    }
```
bitwiseman, line 29: Need to add an internal constructor for this path.
```suggestion
    public GHCommitIterable(GHRepository owner, String sha, List<File> files) {
        this.owner = owner;
        this.sha = sha;
        this.result = new GHCommitFilesPage(files != null ? files.toArray(GHCommit.File[]::new) : new GHCommit.File[0]);
    }
```

src/main/java/org/kohsuke/github/GHCommitIterable.java
```diff
@@ -0,0 +1,72 @@
...
+    @Nonnull
+    @Override
+    public PagedIterator<GHCommit.File> _iterator(int pageSize) {
+
+        GitHubRequest request = owner.root().createRequest().withUrlPath(owner.getApiTailUrl("commits/" + sha)).build();
+
+        return new PagedIterator<>(
+                adapt(GitHubPageIterator.create(owner.root().getClient(), GHCommitFilesPage.class, request, pageSize)),
+                null);
+    }
```
bitwiseman, line 47: Move the logic that controls which iterating strategy to here. 
I haven't tried to compile this yet, so you may need to make some tweaks, but the basic idea should be there. 

You may or may not need to rerecord your test data. I think not....

```suggestion
    public PagedIterator<GHCommit.File> _iterator(int pageSize) {
        Iterator<GHCommit.File[]> pageIterator;
        if (results.getFiles().length() < GHCommit.GH_FILE_LIMIT_PER_COMMIT_PAGE) {
            // create a page iterator that only provides one page
            pageIterator = Collections.singleton(results.getFiles()).iterator();
        } else {
            // page size is controlled by the server for this iterator, do not allow it to be set by the caller
            pageSize = 0;

            GitHubRequest request = owner.root().createRequest().withUrlPath(owner.getApiTailUrl("commits/" + sha)).build();

           pageIterator = adapt(
                    GitHubPageIterator.create(owner.root().getClient(), GHCommitFilesPage.class, request, pageSize));
        }
        return new PagedIterator<>(pageIterator, null);
    }
```

bitwiseman (changes requested):
This is basically all one big change suggestion to make `listFiles()` return a `PagedIterable`.  
Other than that, this PR looks great!  Well tested.

frink182:
@bitwiseman I've made the suggested changes, let me know how it looks now. Thanks

bitwiseman (approved):

PR #1680, opened 2023-06-27
Re-Prepare request for every retry

emrebasar:
# Description

Currently, a retried request fails if the authentication token expires between retries. This can be a common experience, as GitHub API can throttle a user for up to an hour, which is longer than the lifetime of an authentication token retrieved via the GitHub Application Installation.

This change ensures that a recent token is used in each request by re-creating the connector request for every retry.

# Before submitting a PR:

# When creating a PR: 


bitwiseman:
@emrebasar 
I see what you're saying, this is probably an acceptable change, but you'll at least need to update tests to match the new behavior. 

Also, retry scenarios tend to be quick turnaround.  What you're talking about is generally only going to happen related to rate limiting and long waits. Maybe you should add special error scenario for case you want to address.  Make it similar to https://github.com/hub4j/github-api/blob/c4948917ca9fbeee190bcddf0d12941472dff03d/src/main/java/org/kohsuke/github/GitHubClient.java#L629-L653 where you detect the situation and throw a `RetryRequestException`.  

The detector should check the http error code and prepare a new request.  If the authorization string for new request is the same as the current connectorRequest, then you have a valid failure. Otherwise,  throw a `RetryRequestException` with the updated `GitHubConnectorRequest` that the main loop will use for retry.  

Make sense?

bitwiseman (changes requested):
For most users, preparing the request again is not needed. As noted, we should detect the scenario you describe an update the request only in that case.

emrebasar:
Thank you for the feedback. In our use case, rate limiting is the primary cause of retries, and we observe this specific scenario fairly often, though I agree that it might not be the case for majority of the use cases. I will update the code accordingly, and fix the failing tests.

src/main/java/org/kohsuke/github/GitHubClient.java
```diff
@@ -474,6 +474,22 @@ private void detectKnownErrors(GitHubConnectorResponse connectorResponse,
         }
     }
 
+    private void detectExpiredToken(GitHubConnectorResponse connectorResponse, GitHubRequest request)
```
bitwiseman, line 477: This looks great.

bitwiseman (changes requested):
All we need now is a test covering the newly added code.

emrebasar:
Hi @bitwiseman, I did dive into this, and tried implementing some tests but got stuck at configuring the correct Wiremock behavior. I am planning to continue looking at this, beginning Week 32. How would you like to proceed? Should we close this PR, and open a new one when tests are fixed, or should we keep this one open?

bitwiseman:
@emrebasar 
Take your time, we'll keep this open. Update this PR as needed.  Thanks for working on this.  Feel free to ask questions if you get stuck.

bitwiseman:
Missed javadoc comments on tests.

PR #1692, opened 2023-08-10
Fix visibility when creating a repository

### LINKED ISSUE #1456: Creating a Repo with a Visibility Fails
**Describe the bug**
When using the [GHCreateRepositoryBuilder](https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/GHCreateRepositoryBuilder.java) to create a repo and setting the [visibility](https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/GHRepositoryBuilder.java#L161), the resulting API call fails with the following exception (see below). This happens when any of the Visibility enum values are set (`PUBLIC`, `INTERNAL`, or `PRIVATE`). I'm running this against my company's GitHub instance but I recreated the issue against `github.com` as well.

**To Reproduce**
Code to re-produce. I am using the latest release of this library; version 1.306:
```
GitHub.connect(username, token).getOrganization(someOrg).createRepository("a-new-repo").visibility(Visibility.PRIVATE).create();
```

Error:
```
Exception in thread "main" org.kohsuke.github.HttpException: {"message":"Invalid visibility. Valid visibilities are: public, private, internal","documentation_url":"https://docs.github.com/rest/reference/repos#create-an-organization-repository"}
	at org.kohsuke.github.GitHubConnectorResponseErrorHandler$1.onError(GitHubConnectorResponseErrorHandler.java:56)
	at org.kohsuke.github.GitHubClient.detectKnownErrors(GitHubClient.java:424)
	at org.kohsuke.github.GitHubClient.sendRequest(GitHubClient.java:386)
	at org.kohsuke.github.GitHubClient.sendRequest(GitHubClient.java:355)
	at org.kohsuke.github.Requester.fetch(Requester.java:76)
	at org.kohsuke.github.AbstractBuilder.done(AbstractBuilder.java:109)
	at org.kohsuke.github.GHRepositoryBuilder.done(GHRepositoryBuilder.java:241)
	at org.kohsuke.github.GHCreateRepositoryBuilder.done(GHCreateRepositoryBuilder.java:12)
	at org.kohsuke.github.GHCreateRepositoryBuilder.create(GHCreateRepositoryBuilder.java:131)
```

This exception *DOES NOT* occur when changing the visibility on an existing repo. For example, this works:
```
someOrganization.getRepository("a-new-repo").setVisibility(Visibility.PRIVATE);
```

I believe it comes down to how the request is being filled out. When using the builder, the enum is set on the request as an object: 
[GHRepositoryBuilder](https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/GHRepositoryBuilder.java#L163) calls [AbstractBuilder](https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/AbstractBuilder.java#L133) calls [GitHubRequest](https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/GitHubRequest.java#L636). 
In contrast, when setting visibility on an existing repo, a different `with` overload is eventually invoked that correctly transforms the `enum` to a lowercase string: [GHRepository](https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/GHRepository.java#L1356) calls [GitHubRequest](https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/GitHubRequest.java#L555).

Locally, I changed the [GHRepositoryBuilder#visibility](https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/GHRepositoryBuilder.java#L163) method to `return with("visibility", visibility.toString());` and the eventual repo creation doesn't fail.

I can try to find some time to work together a patch with the changes but I was having lots of trouble trying to get the project to build on my computer and I'm not quite sure if the above fix is the correct one so I just wanted to open this bug report before doing anything else.

For now, I will just use the "Create" builder to create the repo and then make a separate call afterwards to set the visibility on the newly created Repo with the API that works.

Sorry if this is a bit scatterbrained. Please let me know if there's more info you need from me. Thank you.

dbaur:
# Description

Fixes #1456.

Calls toString method on the visibility enum in GHRepositoryBuilder ensuring the correct value in the request.

# Before submitting a PR:

# When creating a PR: 


dbaur:
One remark:

The issue also mentions that for some requests the repository is created with the wrong visibility. I think that this is due to GitHub API differentiating between the "visibility" field and the "private" field in its API. 

When creating a repository for a user (i.e. not in an organization) and setting the visibility to "Visibility.PRIVATE" it seems that the API ignores the visibility setting if not also the _private method on the builder is set to true.

This can lead to confusing results, since one typically expects the visibility of the repository to be private when setting the visibility setting.

However, I am not sure if it is in the scope of this project to fix this problem of the GitHub API.

bitwiseman (approved):

bitwiseman (approved):

Issue #1724, opened 2023-10-12
Incompatibility with the latest version of jjwt


PierreBtz:
**Describe the bug**

When using the latest version of jjwt (0.12.x) to handle authentication with a JWT token, I observe the following issue:

```
java.lang.NoSuchMethodError: 'io.jsonwebtoken.JwtBuilder io.jsonwebtoken.JwtBuilder.setIssuedAt(java.util.Date)'
```

**To Reproduce**

I created the following reproducer: https://github.com/PierreBtz/jwt-github-bug.
A unit test shows the issue, using an older version of jjwt the test is GREEN, using the latest 0.12.(x), the test is RED (see README for setup details).

**Expected behavior**

Jwt works properly.

**Desktop (please complete the following information):**
 - OS: OsX Ventura 13.6
 - Browser N/A
 - Version N/A

**Additional context**

The root cause seems to be https://github.com/jwtk/jjwt/pull/794 which deprecated a bunch of setters.
An evident fix would be to modify the code to use the new setters instead of the old deprecated one, and *I'm happy to provide a PR for this however*:

1. I don't understand why a deprecated method wouldn't work anymore :/
2. If we start using the new setters, it will be a breaking change forcing folks to update jjwt to 0.12.x.

bitwiseman:
Because in the `JwtBuilder` that method was not deprecated, it was removed.

Compare:
https://github.com/jwtk/jjwt/blob/529f04dd9097331220c3239bdadacee6b1dfd6de/api/src/main/java/io/jsonwebtoken/JwtBuilder.java#L402-L404

https://github.com/jwtk/jjwt/blob/651f2790dbdb9c10e41f11d7ef52c9ea53f136c3/api/src/main/java/io/jsonwebtoken/JwtBuilder.java#L418-L419

bitwiseman:
0.12.0 is barely a week old.  
https://github.com/jwtk/jjwt/releases/tag/0.12.0 

I would suggest you file a bug and/or PRs in the https://github.com/jwtk/jjwt project to return these methods. I understand why they removed them, but I don't see why these particular methods had to be removed before 1.0 release.  

As to how to address this, we could keep the old dependency in this project and add code that tries to call the new method and catches the `java.lang.NoSuchMethodError` and falls back to the old method.

ihrigb:
I had a bit of time, so I quickly contributed a change: https://github.com/hub4j/github-api/pull/1727

PierreBtz:
[...]

PR #1727, opened 2023-10-16
Update to newer version of jjwt library

ihrigb:
Fixes: #1724

# Description

jjwt recently introduced an incompatible change in their API. This addresses and fixes it. (see https://github.com/hub4j/github-api/issues/1724)
This may leed to problems as jjwt is an optional dependency. Users would also need to update the jjwt version once they will update to a new build of the github-api.

# Before submitting a PR:

# When creating a PR: 


src/main/java/org/kohsuke/github/extras/authorization/JWTTokenProvider.java
```diff
@@ -173,16 +172,16 @@ private String refreshJWT() {
...
         // Let's set the JWT Claims
         JwtBuilder builder = Jwts.builder()
-                .setIssuedAt(Date.from(issuedAt))
-                .setExpiration(Date.from(expiration))
-                .setIssuer(this.applicationId)
-                .signWith(privateKey, SignatureAlgorithm.RS256);
+                .issuedAt(Date.from(issuedAt))
+                .expiration(Date.from(expiration))
+                .issuer(this.applicationId)
+                .signWith(privateKey, Jwts.SIG.RS256);
```
bitwiseman, line 178: Suggest something like this to deal with backward compatibility:
```
        JwtBuilder builder = Jwts.builder();

        // Let's set the JWT Claims
        try {
            builder = builder.issuedAt(Date.from(issuedAt))
                    .expiration(Date.from(expiration))
                    .issuer(this.applicationId)
                    .signWith(privateKey, Jwts.SIG.RS256);
        catch (java.lang.NoSuchMethodError e) { 
            builder = builder.setIssuedAt(Date.from(issuedAt))
                    .setExpiration(Date.from(expiration))
                    .setIssuer(this.applicationId)
                    .signWith(privateKey, SignatureAlgorithm.RS256);
        }
```
This will need manual testing with the old version to verify it works.

bitwiseman:
NOTE: I expect this will fail code coverage due the reflective code not being covered.  I can manually override that.

bitwiseman:
@ihrigb 
I hope you don't mind, I went ahead and added a scenario to the build that runs tests against v0.11.x.  When I did so, I found some issues with the class loading timing and figured I would go ahead and address them.  Take a look at the new structure and comments and tell me if they make sense.

src/main/java/org/kohsuke/github/extras/authorization/JwtBuilderUtil.java
```diff
@@ -0,0 +1,151 @@
...
+     * @param privateKey
+     *            private key
+     * @return JWT
+     */
+    static String buildJwt(Instant issuedAt, Instant expiration, String applicationId, PrivateKey privateKey) {
+
+        try {
+            return DefaultBuilderImpl.buildJwt(issuedAt, expiration, applicationId, privateKey);
+        } catch (NoSuchMethodError | NoClassDefFoundError e) {
+            LOGGER.info(
```
bitwiseman, line 47: Changed this from warning to info to keep from spamming logs too heavily.  What we really want is one `warning` when the method first switches to reflective building and then not again.

ihrigb:
@bitwiseman I don't mind at all. Thanks for the testing effort. I marked the PR ready.

PR #1730, opened 2023-10-19
Paginator

anujhydrabadi:
# Description

`paginator()` method in `PagedIterable` to support starting at a particular page, and next, previous, first, last and jumping to any particular page, along with a few other supporting methods. Made in parallel to `iterator()` method so as to keep backward compatibility, but supports all functionality that `iterator()` provides.

# Before submitting a PR:

# When creating a PR: 

### DIFF
--- a/src/main/java/org/kohsuke/github/PagedIterable.java
+++ b/src/main/java/org/kohsuke/github/PagedIterable.java
@@ -26,6 +26,8 @@ public abstract class PagedIterable<T> implements Iterable<T> {
      */
     private int pageSize = 0;
 
+    private int startPage = 0;
+
     /**
      * Sets the pagination size.
      *
@@ -41,6 +43,11 @@ public abstract class PagedIterable<T> implements Iterable<T> {
         return this;
     }
 
+    public PagedIterable<T> withStartPage(int startPage) {
+        this.startPage = startPage;
+        return this;
+    }
+
     /**
      * Returns an iterator over elements of type {@code T}.
      *
@@ -50,6 +57,26 @@ public abstract class PagedIterable<T> implements Iterable<T> {
     public final PagedIterator<T> iterator() {
         return _iterator(pageSize);
     }
+
+    /**
+     * Returns an iterator over elements of type {@code T}.
+     *
+     * @return an Iterator.
+     */
+    @Nonnull
+    public final Paginator<T> paginator() {
+        return _paginator(pageSize, startPage);
+    }
+
+    /**
+     * Iterator over page items.
+     *
+     * @param pageSize
+     *            the page size
+     * @return the paged iterator
+     */
+    @Nonnull
+    public abstract Paginator<T> _paginator(int pageSize, int startPage);
 
     /**
      * Iterator over page items.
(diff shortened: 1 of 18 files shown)


anujhydrabadi:
Pending tasks:

- [ ] Test `refresh()` method by changing underlying data.
- [ ] Make a few hacky `if` conditions in `GitHubPaginator` (w.r.t. the `next` variable) more readable.
- [ ] Revise and fix javadocs.
- [ ] Capturing snapshot test data.

Question for maintainers/reviewers:

- Should I cover every flow that calls `PagedIterable` with tests? Will be a huge task to have adequate tests for all of them. Currently I have picked one flow (workflow runs) where I am testing every possible case, and assuming the rest of the flows will comply, given the common nature of the APIs.
- I have tests that make assertions on the values of the underlying data, and would fail once that data changes. Should I avoid such tests and think of more creative ways to make my assertions independent of the underlying data, or is it fine? Or is there any other way?

Review comments on the overall design/implementation/behaviour of the methods are welcome, even though it is a WIP PR for now.

bitwiseman:
> Question for maintainers/reviewers:
> 
> * Should I cover every flow that calls `PagedIterable` with tests? Will be a huge task to have adequate tests for all of them. Currently I have picked one flow (workflow runs) where I am testing every possible case, and assuming the rest of the flows will comply, given the common nature of the APIs.

I would assume that testing one scenario per implementation should generally be sufficient.  You'll need to test search scenario separately. 

> * I have tests that make assertions on the values of the underlying data, and would fail once that data changes. Should I avoid such tests and think of more creative ways to make my assertions independent of the underlying data, or is it fine? Or is there any other way?

Ideally, you should find scenarios where the underlying data doesn't change.  However, if that isn't possible, once you record the test data you can mark the test as requiring custom data so that someone doesn't try to re-record it later without knowing the issues.  If you go this path, be sure to add comments explaining how to update/re-record test data in future.

src/main/java/org/kohsuke/github/PagedIterable.java
```diff
@@ -41,6 +44,18 @@ public PagedIterable<T> withPageSize(int size) {
...
 
+    /**
+     * Sets the page size. 1 by default.
+     *
+     * @param startPage
+     *            start page to set
+     * @return the paged iterable
+     */
+    public PagedIterable<T> withStartPage(int startPage) {
+        this.startPage = startPage;
```
bitwiseman, line 55: If this is going to be present on this class, we'll need to pass `startPage` to the `Iterator` as well.

bitwiseman:
I'm not sure I understand why we need a separate `Paginator` class. It tries to smash together item iteration with page iteration. I think if people want to navigate by pages we should have a `toPages(int pageSize)` that returns a `GitHubPageIterable` or even a `GitHubPageIterator` (which already exists but is currently internal).  Either way, this should handle iterating only over pages not their contents. 

Overall I woud expect most of this to do via adding methods to existing classes, `PagedIterable`, `PagedIterator`, `GitHubPageIterator`, `GitHubPageContentsIterable` and `GitHubPageContentsIterable.GitHubPageContentsIterator`. 

If you feel extremely strongly about this, I would even discuss pushing all the functionality of `Paginator` into `PagedIterable` and `PagedIterator` since it seems to be reimplementing significant portions of those classes.

Issue #1722, opened 2023-10-04
Unknown value startup_failure


iamironz:
**Describe the bug**

Hey! 
It looks like `startup_failure` conclusion status has not been added yet into the library itself, but it is presented in REST API:

```
 WARNING: Unknown value startup_failure for enum class org.kohsuke.github.GHWorkflowRun$Conclusion, defaulting to UNKNOWN
```

**To Reproduce**
1. Call `listWorkflows` and filter to the one that has "failed to start" runs
2. Call `listRuns` for this instance
3. Call `withPageSize(n)` (to limit the output size) to the point where `startup_failure` is presented and `toList` 
4. Call `getConclusion` method

**Expected behavior**
No warnings on `getConclusion` method call

**Version:**
 - Library version `1.316`

gsmet:
Looks like a good opportunity to contribute :).

yasin-herken:
I want to work on it

bitwiseman:
@yasin-herken  Go ahead.

yasin-herken:
I couldn't figure out how to add workflow according to the Conclusion enum class but I added a test just in case

PR #1737, opened 2023-11-07
Added startup failure to the Conclusion enum class and added test acc…

yasin-herken:
Fixes#Unknown value startup_failure #1722
I added startup_failures to the Conclusion enum class and wrote a test but I couldn't figure out what GitHub workflows have with conclusion startup_failures so I wrote a test which has 0 length but if I can add workflow that has startup failures, I can change the test result.


bitwiseman (changes requested):
@yasin-herken 
Thanks for the PR!

Please review the steps in the contributing for how to record data for a new test.  I've invited you to the `hub4j-test-org` to enable that.  Then you can do whatever you need to create a workflow that fails to start and records a result for that.

yasin-herken:
I have added the startup-failure workflow however, I couldn't get the `startup-failure ` conclusion. I am getting `failure` conclusion, how to create workflow that has `startup-failure` conclusion

bitwiseman:
@yasin-herken 
Record the test with the the failure conclusion. Then manually edit the recorded JSON to have `startup-failure` instead of `failure`.  Then add `snapshotNotAllowed()` to the top of the test method to ensure it isn't accidentally re-recorded later.

yasin-herken:
Fixes#Unknown value startup_failure https://github.com/hub4j/github-api/issues/1722
I added startup_failures to the Conclusion enum class and wrote a test for it.

bitwiseman (approved):

PR #1742, opened 2023-11-15
Publish "nobridge" version of the library

### LINKED ISSUE #1739: Unable to mock GHObject#id
Hi, 

it seems like it's impossible to mock `GHObject` via Mockito, I observe this behavior:

```
try {
    var a = Mockito.mock(GHObject.class);
    when(a.getId()).thenReturn(123L);
} catch (Throwable e) {
    e.printStackTrace();
}
```

The exception:
```
java.lang.NullPointerException: Cannot invoke "java.lang.Long.longValue()" because the return value of "java.util.concurrent.Callable.call()" is null
	at org.kohsuke.github.GHObject.getId(GHObject.java:140)
	at org.kohsuke.github.GHObject.getId(GHObject.java:140)
```

I believe this may be related to bridge methods, but I have no idea how to implement a workaround.

Could you please elaborate if this is possible?

bitwiseman:
# Description

Converts the project to a multi-module structure  and adds a new "No Bridge" build output.  

* Multi-module - allows us to produce more than one flavor of library based on the same code.
* "No Bridge" build flavor - compiles the library without generating bridge methods. This flavor of the library does NOT guarantee binary backward compatibility. However, it does unblock mocking ( fixes #1739 ). 

Multi-module will also allow us to begin producing a v2 beta flavor without breaking the consumers of the v1 line. 

# Before submitting a PR:

# When creating a PR: 


bitwiseman:
@gsmet @Haarolean 
Do you think you could try this out and see if it helps with the mocking issue?

Haarolean:
@bitwiseman sure, thank you! Will take a look

github-api-nobridge/pom.xml
```diff
@@ -0,0 +1,306 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.kohsuke</groupId>
```
Haarolean, line 3: groupId declaration is redundant here due to the module being inherited from a parent

Haarolean:
@bitwiseman I can confirm that mocking objects works now. Thank you!

bitwiseman:
I'm going to make this a `release/v1.x-nobridge` branch instead of multimodule.

PR #1744, opened 2023-11-19
Add sanity caching and retry controls

bitwiseman:
# Description

* Add one query per second limit on `getRateLimit()` and `getMeta()`
* Add configurable and randomization to retries

Mitigate the most severe parts of #1728

@samrocketman @KeepItSimpleStupid 
Please take a look. 

# Before submitting a PR:

# When creating a PR: 


src/main/java/org/kohsuke/github/GitHubClient.java
```diff
@@ -45,11 +46,22 @@
...
+    private static final int DEFAULT_MAXIMUM_RETRY_MILLIS = DEFAULT_MINIMUM_RETRY_MILLIS;
+
+    // WARNING: These are unsupported environment variables. The GitHubClient class is internal and may change at any
+    // time.
+    private static final int retryCount = Math.max(DEFAULT_CONNECTION_ERROR_RETRIES,
+            Integer.getInteger(GitHubClient.class.getName() + ".retryLimit", DEFAULT_CONNECTION_ERROR_RETRIES));
+    private static final int minRetryInterval = Math.max(DEFAULT_MINIMUM_RETRY_MILLIS,
+            Integer.getInteger(GitHubClient.class.getName() + ".minRetryInterval", DEFAULT_MINIMUM_RETRY_MILLIS));
+    private static final int maxRetryInterval = Math.max(DEFAULT_MAXIMUM_RETRY_MILLIS,
+            Integer.getInteger(GitHubClient.class.getName() + ".minRetryInterval", DEFAULT_MINIMUM_RETRY_MILLIS));
```
KeepItSimpleStupid, line 64: I suppose it should be ?
```suggestion
    private static final int maxRetryInterval = Math.max(DEFAULT_MAXIMUM_RETRY_MILLIS,
            Integer.getInteger(GitHubClient.class.getName() + ".maxRetryInterval", DEFAULT_MAXIMUM_RETRY_MILLIS));
```

samrocketman (changes requested):
Can we not use properties in static fields?  They require a JVM restart which can be a challenge on a heavily active system.  Any particular reason you don't want to do a property look up where the integers get used?

Defining startup properties isn't generally bad but if you're trying to do turning for the retry behavior it's a drag to have downtime every tine you tweak it.

src/main/java/org/kohsuke/github/GitHubClient.java
```diff
@@ -632,11 +651,15 @@ private static IOException interpretApiError(IOException e,
 
     private static void logRetryConnectionError(IOException e, URL url, int retries) throws IOException {
         // There are a range of connection errors where we want to wait a moment and just automatically retry
+        long sleepTime = minRetryInterval;
```
samrocketman, line 654: I suggest setting `minRetryInterval` and `maxRetryInterval` via `Integer.getInteger` property here.

src/main/java/org/kohsuke/github/GitHubClient.java
```diff
@@ -632,11 +651,15 @@ private static IOException interpretApiError(IOException e,
...
+        if (maxRetryInterval > minRetryInterval) {
+            sleepTime = ThreadLocalRandom.current().nextLong(minRetryInterval, maxRetryInterval);
+        }
         LOGGER.log(INFO,
-                e.getMessage() + " while connecting to " + url + ". Sleeping " + GitHubClient.retryTimeoutMillis
+                e.getMessage() + " while connecting to " + url + ". Sleeping " + sleepTime
```
samrocketman, line 659: I recommend creating a trace ID here for debug logging.  That way an admin can search debug logs and find related logs for a single retry sequence.  Here's an example

* log https://github.com/jenkinsci/scm-filter-jervis-plugin/blob/80e28289ede66fa18553b4e0ca5f518a4bd782bc/src/main/groovy/net/gleske/scmfilter/impl/trait/JervisFilterTrait.groovy#L144
* creating trace ID via sha256sum https://github.com/jenkinsci/scm-filter-jervis-plugin/blob/80e28289ede66fa18553b4e0ca5f518a4bd782bc/src/main/groovy/net/gleske/scmfilter/impl/trait/JervisFilterTrait.groovy#L245

When I enable debug logging for a class in the mentioned class it is so active in parallel that all of the logs come in out of order.  Because of that, using the `trace-` ID as a prefix to all of the logs enable me to search for a series of logs along with their retries.  It enabled me to find the maximum retry count across logs as well which helps an admin with tuning.

For exmaple, I default to retries of 30 in my class but I found in practice with GitHub it could retry up to 28 times.  Because that was so close to the max retry limit I increased the retry limit to 60 in my particular setup.

I also set the minimum time between retries to be 1000ms and the maximum to be 3000ms.  I've found GitHub requiring me to retry up to 1 minute in these scenarios because of secondary API limits.

The new secondary API limits are very aggressive at the moment.

samrocketman: I realize the logging mechanisms have to change a little bit with my feedback; they're not as straightforward as making the change just in this area of code.

bitwiseman:
@samrocketman 

I was thinking property lookup is slow and I don't want to do it per query. But compared to query times it's probably negligible. I see your point about down time. 

This isn't intended as a long term feature as it stands right now and I want to have the least change to existing behavior possible. 

I'd be okay with saying that if any of the environment variables are set on startup then they will be checked for every query. So, you have to opt in to the behavior on startup.

samrocketman:
> @samrocketman
> 
> I was thinking property lookup is slow and I don't want to do it per query. But compared to query times it's probably negligible. I see your point about down time.
> 
> This isn't intended as a long term feature as it stands right now and I want to have the least change to existing behavior possible.
> 
> I'd be okay with saying that if any of the environment variables are set on startup then they will be checked for every query. So, you have to opt in to the behavior on startup.

being tunable at all is a plus, really.  In the case of pipeline API interactions we're moving to a weird but workable hack where we obtain one of 10 flocks randomly meaning there can be up to 10 clients active with GitHub at a time in pipelines.  That's kind of how bad it is, though, we're at that point.

bitwiseman:
@samrocketman 
Updated with trace id for logging.  

You said:
> Maybe the delay of "1 second" should be configurable as well ?

Perhaps, but why add the complexity? My thought right now is this is purely a *sanity* check. One second is sane and simple. 
If this is enough to bring stability, we're done. If not, we can add configurability.

KeepItSimpleStupid:
> You said:
> 
> > Maybe the delay of "1 second" should be configurable as well ?
> 
> Perhaps, but why add the complexity? My thought right now is this is purely a _sanity_ check. One second is sane and simple. If this is enough to bring stability, we're done. If not, we can add configurability.

@bitwiseman This comment was from me : I was not sure if you wanted to adopt an iterative approach or if you wanted to anticipate all the needs ;) But it's perfectly fine like that !

For my information, once it's merged and released, for a usage in Jenkins, a new version of [this plugin](https://github.com/jenkinsci/github-api-plugin) would be needed, right ? 

Thanks a lot !

bitwiseman:
@KeepItSimpleStupid 

> @bitwiseman This comment was from me : I was not sure if you wanted to adopt an iterative approach or if you wanted to anticipate all the needs ;) But it's perfectly fine like that !

Yes, iterative.  

> For my information, once it's merged and released, for a usage in Jenkins, a new version of [this plugin](https://github.com/jenkinsci/github-api-plugin) would be needed, right ?

Yes.

bitwiseman:
Coverage is lower due restructured logging behavior.  I think it is acceptable.

bitwiseman:
@samrocketman 
Does this look ready to merge?

bitwiseman:
@samrocketman 
Approved?

samrocketman:
> @samrocketman
> Approved?

Away on holiday at the moment so can't easily review from mobile but I'll take a look

samrocketman (commented):
These changes look good to me.  Thanks for implementing and accommodating feedback.  Interesting use of thread local string I learn something every review in Jenkins projects.

bitwiseman:
Released in 1.318.

samrocketman:
I subscribed to github-api plugin repo for releases

PR #1747, opened 2023-11-20
Allow query parameters for listInstallations query

dunningdan:
# Description

Added a change to allow the option to pass in the `since` parameter per https://developer.github.com/v3/apps/#list-installations. 

# Before submitting a PR:

# When creating a PR: 


bitwiseman (changes requested):
Other methods like this with query parameters in their endpoints, either use a query builder or take the specific parameter(s), in this case that would be `Date since`: 

```java
public PagedIterable<GHAppInstallation> listInstallations(final Date since) {
```

dunningdan:
@bitwiseman sorry about the closed PR. I went to sync the PR fork and accidently removed my commits which cause this to close. I made the change you suggested but it's now at: https://github.com/hub4j/github-api/pull/1748

PR #1780, opened 2024-01-15
Add Suspended Installation Properties to GHAppInstallation

gilday:
# Description

A suspended app installation has `suspended_at` and `suspended_by` properties that were missing from the `GHAppInstallation` class.

https://docs.github.com/en/rest/apps/apps?apiVersion=2022-11-28#list-installations-for-the-authenticated-app


src/main/java/org/kohsuke/github/GHAppInstallation.java
gilday, line 1: These properties are represented in the current test data, but they are always `null`. To record examples of suspended installations, it's my understanding that a test would need to suspend an installation on the hub4j-test-org, and doing so requires admin level access to the org. This sounds prohibitively difficult to contribute. Please advise if there's a good way to test these changes.

bitwiseman: You could manually add test datafiles or I can grant you admin access to the org, whichever you would rather do.

src/main/java/org/kohsuke/github/GHAppInstallation.java
```diff
@@ -45,6 +46,10 @@ public class GHAppInstallation extends GHObject {
     @JsonProperty("repository_selection")
     private GHRepositorySelection repositorySelection;
     private String htmlUrl;
+    @JsonProperty("suspended_at")
```
bitwiseman, line 49: This is conversion in done automatically. 

```suggestion
```

src/main/java/org/kohsuke/github/GHAppInstallation.java
```diff
@@ -45,6 +46,10 @@ public class GHAppInstallation extends GHObject {
...
+    private Date suspendedAt;
+    @JsonProperty("suspended_by")
+    private GHUser suspendedBy;
```
bitwiseman, line 52: This is conversion in done automatically. 

```suggestion
```
This is conversion in done automatically. 

```suggestion
```
This conversion is done automatically. 
```suggestion
    private Date suspendedAt;
    private GHUser suspendedBy;
```

bitwiseman (changes requested):
I understand that adding tests for fields like these is difficult, but they need to be covered.  If they aren't a future change could silent break this functionality.

src/main/java/org/kohsuke/github/GHAppInstallation.java
gilday: Added the tests and test files.

src/main/java/org/kohsuke/github/GHAppInstallation.java
```diff
@@ -311,6 +314,26 @@ public void setRepositorySelection(GHRepositorySelection repositorySelection) {
...
 
+    /**
+     * Gets suspended at.
+     *
+     * @return the suspended at
+     */
+    @SuppressFBWarnings(value = { "EI_EXPOSE_REP" }, justification = "Expected behavior")
+    public Date getSuspendedAt() {
+        return suspendedAt;
+    }
```
bitwiseman, line 325:
```suggestion
    public Date getSuspendedAt() {
        return GitHubClient.parseDate(suspendedAt);
    }
```

src/main/java/org/kohsuke/github/GHAppInstallation.java
```diff
@@ -45,6 +46,8 @@ public class GHAppInstallation extends GHObject {
     @JsonProperty("repository_selection")
     private GHRepositorySelection repositorySelection;
     private String htmlUrl;
+    private Date suspendedAt;
```
bitwiseman, line 49:
```suggestion
    private String suspendedAt;
```

PR #1791, opened 2024-02-13
Add support for artifacts uploaded by actions/upload-artifact@v4

### LINKED ISSUE #1790: Unable to download artifacts uploaded with upload-artifact@v4
GitHub published a new version of upload-artifact, v4: https://github.com/actions/upload-artifact .
It is supposed to be a lot better, most notably faster and more flexible.

I tried to use it and unfortunately, we hit an issue:
```
2024-02-09 18:36:28,471 ERROR [io.qua.bot.bui.git.BuildReportsUnarchiver$ArtifactIsDownloaded] (awaitility-thread) Pull request #166 - Unable to download artifact build-reports-JVM Tests - JDK 17 - windows-latest- retry #5: org.kohsuke.github.HttpException: ﻿<?xml version="1.0" encoding="utf-8"?>
<Error><Code>AuthenticationFailed</Code><Message>Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
RequestId:dfb418bf-301e-007c-297e-5b6ca7000000
Time:2024-02-09T17:36:28.4208458Z</Message></Error>
	at org.kohsuke.github.GitHubConnectorResponseErrorHandler$1.onError(GitHubConnectorResponseErrorHandler.java:62)
	at org.kohsuke.github.GitHubClient.detectKnownErrors(GitHubClient.java:504)
	at org.kohsuke.github.GitHubClient.sendRequest(GitHubClient.java:464)
	at org.kohsuke.github.GitHubClient.sendRequest(GitHubClient.java:427)
	at org.kohsuke.github.Requester.fetchStream(Requester.java:131)
	at org.kohsuke.github.GHArtifact.download(GHArtifact.java:125)
```

I stumbled upon https://github.com/arduino/report-size-deltas/pull/83 which seems to indicate that we shouldn't send the `Authorization` header to the download URL.
Unfortunately, my guess is that the Java HTTP Client is not flexible enough for that and is sending the `Authorization` header anyway.
It's not possible to configure the redirect behavior at the request level either so that's a bit problematic.

I'm not completely sure how to solve this (or even if my diagnostic is the correct one...).

gsmet:
The artifacts upload by actions/upload-artifact@v4 are hosted on a new infrastructure which has several constraints:
- we will have an error if we push the Authorization header to it, which was the case when using the Java 11 HttpClient (and is considered a bad practice so it is good to have fixed it anyway)
- the host name is dynamic so our test infrastructure was having problems with proxying the request - I was able to make it work with some Wiremock magic (damn, it was hard!)

All these problems are sorted out by this pull request and we are now testing an artifact uploaded by v3 and the other uploaded by v4.

Fixes #1790

Note that in the end the problem was only present for the Java 11+ HttpClient as both the Okhttp client and the UrlConnection client actually cleans up the `Authorization` header when redirected to another server.
Personally, I think it's border line a CVE in the JDK but I wouldn't hold my breath in getting it fixed so I think it's worth fixing it ourselves.

# Before submitting a PR:

- [X] Add JavaDocs and other comments explaining the behavior. 
- [X] When adding or updating methods that fetch entities, add `@link` JavaDoc entries to the relevant documentation on https://docs.github.com/en/rest . 
- [X] Add tests that cover any added or changed code. This generally requires capturing snapshot test data. See [CONTRIBUTING.md](CONTRIBUTING.md) for details.

# When creating a PR: 

  - [X] Provide links to relevant documentation on https://docs.github.com/en/rest where possible. If not including links, explain why not.
- [X] All lines of new code should be covered by tests as reported by code coverage. Any lines that are not covered must have PR comments explaining why they cannot be covered. For example, "Reaching this particular exception is hard and is not a particular common scenario."


gsmet:
I added a commit to for Git longpaths support on CI as the CI failures were due to paths that were too long.

gsmet:
CI is all green, now, ready to get merged!

bitwiseman:
@gsmet 
> I added a commit to for Git longpaths support on CI as the CI failures were due to paths that were too long.

Oof.  Hm.  I didn't know about `core.longpaths true` and did a bunch of work recently to keep the recorded file names shorter.  See https://github.com/hub4j/github-api/pull/1743/commits/f0a3695dd4285dc875a4df958868e824f15da723 .

I'd rather not require the long file name flag.  If we go that route, we also need to add instructions to the `CONTRIBUTING.md` explaining what windows user need to do on their systems. 


```
    // old
    "bodyFileName": "u72ug1ib1zbctek798hyrdyou28rbk6ssrokf37zxrpgubk95i__apis_pipelines_1_runs_120_signedartifactscontent-1.txt",
    // new
    "bodyFileName": "u72ug1ib1zbctek798hyrdyou28rbk6ssrokf37zxrpgubk95i__apis_pipelines_1_runs_75_signedartifactscontent-446b2495-1bf6-4d5e-acde-daff02e51161.txt",
```

That string on the front is ridiculous and then the file name GUID isn't shortened by the Wiremock rule.  This is a bug in the test framework.  I'd rather fix it that require an additional setting for Windows.

For this PR (for expediency), would you be willing to revert the long file name commit and manually edit the file names to shorten them? Then file an issue, so I can update the Wiremock rule further to shorten those file names that are close to the limit?

src/main/java11/org/kohsuke/github/extras/HttpClientGitHubConnector.java
```diff
@@ -22,16 +26,30 @@
...
+        // even when redirecting to a different host which is problematic as we don't want
+        // to push the Authorization header when redirected to a different host.
+        // This problem was discovered when upload-artifact@v4 was released as the new
+        // service we are redirected to for downloading the artifacts doesn't support
+        // having the Authorization header set.
+        // The new implementation does not push the Authorization header when redirected
+        // to a different host, which is similar to what Okhttp is doing:
+        // https://github.com/square/okhttp/blob/f9dfd4e8cc070ca2875a67d8f7ad939d95e7e296/okhttp/src/main/kotlin/okhttp3/internal/http/RetryAndFollowUpInterceptor.kt#L313-L318
+        // See also https://github.com/arduino/report-size-deltas/pull/83 for more context
+        this(HttpClient.newBuilder().followRedirects(HttpClient.Redirect.NEVER).build());
```
bitwiseman, line 52: What about for private repos with uploaded artifacts?  Do they require the Authorization header after redirect there?

src/main/java11/org/kohsuke/github/extras/HttpClientGitHubConnector.java
```diff
@@ -69,13 +87,50 @@ public GitHubConnectorResponse send(GitHubConnectorRequest connectorRequest) thr
...
         try {
-            HttpResponse<InputStream> httpResponse = client.send(request, HttpResponse.BodyHandlers.ofInputStream());
+            HttpResponse<InputStream> httpResponse = send(request, 0);
             return new HttpClientGitHubConnectorResponse(connectorRequest, httpResponse);
         } catch (InterruptedException e) {
             throw (InterruptedIOException) new InterruptedIOException(e.getMessage()).initCause(e);
         }
     }
 
+    private HttpResponse<InputStream> send(HttpRequest request, int redirects)
```
bitwiseman, line 97: Okay, I see what you're trying to do here.  Looks good!    

But I'd like moved to `GitHubClient.detectRedirect`.   

From what you described, authorization stripping is the behavior GitHub expects. That means, regardless of what `GitHubConnector` implementation is used, when the connector returns a redirect response this project must strip the Authorization header as appropriate and then follow the redirect.  So, this functionality should go in `GitHubClient` instead of `HttpClientGitHubConnector`. 

On the plus side, `GitHubClient` already has functionality for retries. 

https://github.com/hub4j/github-api/blob/51b3a06fe364a9667d8c4e0b64abc780cff8d371/src/main/java/org/kohsuke/github/GitHubClient.java#L461-L471

A retry exception can be thrown during error detection and the updated request contained in the exception will be used for the next attempt. We use this for token refresh, now we can use it for redirects.   And we already detect redirects.  

https://github.com/hub4j/github-api/blob/51b3a06fe364a9667d8c4e0b64abc780cff8d371/src/main/java/org/kohsuke/github/GitHubClient.java#L508-L535

Use `detectExpiredToken` as the template and replace  `detectRedirect` with the logic you've implemented here.

src/main/java11/org/kohsuke/github/extras/HttpClientGitHubConnector.java
```diff
@@ -104,4 +159,49 @@ public void close() throws IOException {
...
+                .map(URI::create)
+                .orElseThrow(() -> new IOException("Invalid redirection"));
+
+        // redirect could be relative to original URL, but if not
+        // then redirect is used.
+        redirectedURI = originalUri.resolve(redirectedURI);
+        return redirectedURI;
+    }
+
+    // This implements the exact same rules as the ones applied in RedirectFilter
```
bitwiseman, line 175: `RedirectFilter`?  What project is this from?

src/test/java/org/kohsuke/github/junit/GitHubWireMockRule.java
```diff
@@ -45,6 +49,12 @@ public class GitHubWireMockRule extends WireMockMultiServerRule {
     private final static boolean useProxy = takeSnapshot
             || System.getProperty("test.github.useProxy", "false") != "false";
 
+    private final static Pattern ACTIONS_USER_CONTENT_PATTERN = Pattern
+            .compile("https://pipelines[a-z0-9]*\\.actions\\.githubusercontent\\.com", Pattern.CASE_INSENSITIVE);
+    private final static Pattern BLOB_CORE_WINDOWS_PATTERN = Pattern
+            .compile("https://([a-z0-9]*\\.blob\\.core\\.windows\\.net)", Pattern.CASE_INSENSITIVE);
+    private final static String ORIGINAL_HOST = "originalHost";
+
```
bitwiseman, line 57: Sorry this is such a mess.  Well done updating it.

src/test/java/org/kohsuke/github/junit/GitHubWireMockRule.java
```diff
@@ -235,6 +264,8 @@ protected void after() {
         recordSnapshot(this.codeloadServer(), "https://codeload.github.com", true);
 
         recordSnapshot(this.actionsUserContentServer(), "https://pipelines.actions.githubusercontent.com", true);
+
+        recordSnapshot(this.blobCoreWindowsNetServer(), "https://productionresults.blob.core.windows.net", true);
     }
 
     private void recordSnapshot(WireMockServer server, String target, boolean isRawServer) {
```
bitwiseman, line 271: If you look at the `recordSnapshot`, we add headers to the recording that are needed by tests.  

To test this feature, we need to verify that the Authorization header is present on the regular requests and that it is stripped from the redirected request.

To do that I think we need to an option to use/record/verify a placeholder `Authorization` header as part of a test.  You've done a ton of work here already.  If you'd like, I can do make this update.

bitwiseman (changes requested):
This is great work! I'm happy to pitch in if you want a hand with the remaining changes.

src/main/java11/org/kohsuke/github/extras/HttpClientGitHubConnector.java
gsmet: Ah. That's an interesting question. I will test that. It would be a nightmare to handle so I hope not...

The URL is supposed to be extremely short lived so I think they suppose it is safe.

src/main/java11/org/kohsuke/github/extras/HttpClientGitHubConnector.java
gsmet: I dunno both if it's worth it and if it's safe enough.
As for the former, as I said, I did my homework and all the other clients are having this behavior of stripping the `Authorization` header. So sure, we could add another client that wouldn't have this behavior at some point in the future but I also think it's probably better to let the client handle the redirects if it can do it properly.

I also wanted to be very safe and carefully implement the rule of the Java HttpClient to make sure I wouldn't break anything.

Anyway, maybe you're right, I'm just not sure it's a decision that is so obvious.

src/main/java11/org/kohsuke/github/extras/HttpClientGitHubConnector.java
gsmet: Ah yes, sorry, it's the JDK class. I should have added the package.

src/test/java/org/kohsuke/github/junit/GitHubWireMockRule.java
gsmet: Yeah, I struggled a lot with it. Note that I don't think it's a limitation of what you did, I was surprised I didn't have some sort of context attached to requests by Wiremock so that I could add the host transparently there.
If at some point, they add a signature to check the URL is not tainted, we will be screwed :/

src/main/java11/org/kohsuke/github/extras/HttpClientGitHubConnector.java
gsmet: It's `jdk.internal.net.http.RedirectFilter`.

gsmet:
Hey @bitwiseman , thanks for having a look. TBH, I spent too much time on it already as I spent quite a lot of time researching the issue and then experiment to find a solution (I initially hoped I could tweak the Java HttpClient behavior without implementing the redirect logic myself, to no avail...). So if you have some cycles I would appreciate you having a look.

I'm around if you have any questions.

src/main/java11/org/kohsuke/github/extras/HttpClientGitHubConnector.java
bitwiseman:
```suggestion
    // This implements the exact same rules as the ones applied in jdk.internal.net.http.RedirectFilter
```

bitwiseman:
@gsmet  
I re-wrote the test resource renaming functionality (again) to be way more robust.  Ran the new behavior on tests that showed up with the longest files paths.  

Code coverage showed the authorization removal functionality was not being triggered during local testing due to all the wiremock servers running on various ports of `localhost`.  So now host and port must be the same in the redirect to be considered the same host. 

Still don't have a test for authorization removal, but I might file that as an issue for a later PR.

gastaldi:
The changes look good. Is this now good to be merged?

gsmet:
@bitwiseman can't say I'm fully convinced your version is better but if you prefer this version, go for it :).

I suppose you have tested locally that the test I added still works fine? I see the files renamed but the content wasn't updated thus why I ask.

gsmet:
@bitwiseman given https://github.com/hub4j/github-api/issues/1790#issuecomment-1981991420 , we probably need to accelerate a bit on this as it's going to be in the way of more and more users.

src/test/resources/org/kohsuke/github/GHWorkflowRunTest/wiremock/testArtifacts_blob-core-windows-net/mappings/1-a_9_w_artifacts_41e13e58.json
```diff
@@ -7,6 +7,9 @@
     "headers": {
       "Accept": {
         "equalTo": "application/vnd.github.v3+json"
+      },
+      "Authorization": {
+        "absent" : true
```
bitwiseman, line 12: @gsmet Verified that the header is not present for the redirected requests!

gsmet:
Awesome, thanks for driving it home and merging!

bitwiseman:
@gsmet  I'll get the release out with this change this weekend.

gsmet:
Awesome. `upload-artifact@v4` is so much better. Looking forward to being able to move to it!

gsmet:
@bitwiseman I'm working on a pull request to fix https://github.com/quarkiverse/quarkus-github-app/issues/580 and add the necessary events to GitHub API.

It would be nice if you could wait for it to be merged before releasing. It should arrive soon.

PR #1793, opened 2024-02-14
Support state reason for issues

### LINKED ISSUE #1792: State reason for issues not supported
**Describe the bug**
GitHub supports closing an issue with `state_reason`, but this is not supported in this library at the moment. This is not really a bug, but the issue templates for this repository only allows creating bugs.

**To Reproduce**

**Expected behavior**

**Additional context**

stianst:
Hi, first of thanks for providing this library. I'm using it quite a lot, but finding a few things that are missing for my needs, and trying to contribute those, with the first one being supporting closing issues as not planned.

Trying to figure out how to work with the testsuite, but having a few issues. 

When running `mvn install -Dtest.github.org=false -Dtest=GHIssueTest` I'm getting the following exception:
```
Caused by: java.lang.reflect.InaccessibleObjectException: Unable to make field protected java.lang.String java.net.HttpURLConnection.method accessible: module java.base does not "opens java.net" to unnamed module @551bca18
```

Another thing is the test is hardwired to `hub4j-test-org` which I don't have access to, so guess I need to use my own repository somehow.

Closes #1792


gsmet:
Hey @stianst ,

Yeah, you usually need access to the test org when recording testing as it's easier for them to be reproduced by others when something needs a change.

You need @bitwiseman for that.

As for the error you have when running the tests, the test runs by default with an HTTP Client that is not compatible with Java 17+. Run them with Java 11 and things should work better.

src/main/java/org/kohsuke/github/GHIssue.java
```diff
@@ -67,6 +69,9 @@ public class GHIssue extends GHObject implements Reactable {
     /** The state. */
     protected String state;
 
+    /** The state. */
```
gsmet, line 72: This needs an update

src/main/java/org/kohsuke/github/GHIssueStateReason.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2011, Eric Maupin
```
gsmet, line 4: This is off. But I'm not sure if we are encouraged to put our name here, these days. @bitwiseman would know.

src/main/java/org/kohsuke/github/GHIssueStateReason.java
```diff
@@ -0,0 +1,30 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.kohsuke.github;
+
+public enum GHIssueStateReason {
+
+    COMPLETED, NOT_PLANNED
```
gsmet, line 29: As previously stated, you can add `UNKNOWN` here. This avoids errors when GitHub add a new value.

gsmet:
Ah, actually, good news, I forgot I had the permission to do that.

I added you to the test org. You should have received an invite.

@bitwiseman Stian is a colleague from Red Hat leading the Keycloak project.

gsmet (changes requested):
I spotted a few minor issues.

gsmet:
Ah also, do not record things for all the tests, doing:

```
rm -rf src/test/resources/org/kohsuke/github/GHIssueTest/wiremock/closeIssue/
mvn install -Dtest.github.takeSnapshot -Dtest=GHIssueTest#closeIssue
rm -rf src/test/resources/org/kohsuke/github/GHIssueTest/wiremock/closeIssueNotPlanned/
mvn install -Dtest.github.takeSnapshot -Dtest=GHIssueTest#closeIssueNotPlanned
```

will be better.

Then you should add the new snapshots to your commit.

src/main/java/org/kohsuke/github/GHIssueStateReason.java
stianst: Looked at a couple recent commits and they don't have any license headers so removed this.

src/main/java/org/kohsuke/github/GHIssue.java
```diff
@@ -198,6 +203,15 @@ public GHIssueState getState() {
         return Enum.valueOf(GHIssueState.class, state.toUpperCase(Locale.ENGLISH));
     }
 
+    /**
+     * Gets state reason.
+     *
+     * @return the state reason
+     */
+    public GHIssueStateReason getStateReason() {
+        return Enum.valueOf(GHIssueStateReason.class, state_reason.toUpperCase(Locale.ENGLISH));
```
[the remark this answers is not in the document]
stianst: Nice, that's much better. Updated to use that.

stianst:
Got the test working eventually (well, okay I only spent 30 min, so this wasn't too bad ;)), but used an org I created for the purpose. Will try with the official org and update the data.

stianst:
Should be ready to go now

src/main/java/org/kohsuke/github/GHIssue.java
```diff
@@ -198,6 +204,15 @@ public GHIssueState getState() {
         return Enum.valueOf(GHIssueState.class, state.toUpperCase(Locale.ENGLISH));
     }
 
+    /**
+     * Gets state reason.
+     *
+     * @return the state reason
+     */
+    public GHIssueStateReason getStateReason() {
+        return EnumUtils.getEnumOrDefault(GHIssueStateReason.class, state_reason, GHIssueStateReason.UNKNOWN);
```
gsmet, line 213: Are you sure this can't be null? My understanding is that when the issue is not closed, it will be null but I might be mistaken.
If so, please use `EnumUtils.getNullableEnumOrDefault()`. I pointed you to it in my previous comment but it might have been a bit confusing as I pointed to this one first.

gsmet (commented):
I added a very small comment for something that we need to check but otherwise it looks very good.

Note that we will have to wait for @bitwiseman to merge it as I'm not a committer here.

bitwiseman:
> @gsmet @bitwiseman would it be okay to enable fine-grained personal access tokens for the org (https://github.com/organizations/hub4j-test-org/settings/personal-access-tokens-onboarding). They are a bit nicer than the legacy tokens as you can limit to a specific org/repo.

Enabled the fine-grained PATs.

src/main/java/org/kohsuke/github/GHIssue.java
stianst: Updated, I was kinda thinking it should be set to UNKNOWN when null, but obviously null makes more sense ;)

stianst:
Fixed the last issue with using `EnumUtils.getNullableEnumOrDefault` and added Javadocs to `GHIssueStateReason` since the `site` job complained about that.

gsmet (approved):

src/main/java/org/kohsuke/github/GHIssueStateReason.java
```diff
@@ -0,0 +1,13 @@
+package org.kohsuke.github;
+
+public enum GHIssueStateReason {
```
bitwiseman, line 3:
```suggestion

/**
 * The enum GHIssueStateReason.
 */
public enum GHIssueStateReason {
```

src/main/java/org/kohsuke/github/GHIssue.java
```diff
@@ -294,6 +313,19 @@ public void close() throws IOException {
         edit("state", "closed");
     }
 
+    /**
+     * Closes this issue.
+     *
+     * @throws IOException
+     *             the io exception
+     */
+    public void close(GHIssueStateReason reason) throws IOException {
```
bitwiseman, line 322:
```suggestion
    /**
     * Closes this issue.
     *
     * @param reason
     *            the reason the issue was closed
     * @throws IOException
     *             the io exception
     */
    public void close(GHIssueStateReason reason) throws IOException {
```

src/main/java/org/kohsuke/github/GHIssueStateReason.java
```diff
@@ -0,0 +1,17 @@
+package org.kohsuke.github;
+
+
```
bitwiseman, line 3:
```suggestion
```

bitwiseman (approved):

PR #1799, opened 2024-02-22
Only try to strip upload url 'helpful garbage' if it is present

SuperKael:
This library is mostly compatible with the Gitea API, which although somewhat incidental, is nonetheless wonderful. However, it fails to upload release assets due to this one small and easily-fixed issue. Adding this check does not in any way obstruct the normal use of the GitHub API.

# Description

The GHRelease.uploadAsset method substrings the upload URL to grab the URL before the 'helpful garbage' provided by GitHub, which begins with a '{' character. However, if the upload URL does not have a '{' character in it for some reason (such as the API actually being provided by Gitea, not GitHub), this method throws an exception. This change simply adds a check to avoid this exception.

# Before submitting a PR:

# When creating a PR: 

### DIFF
--- a/src/main/java/org/kohsuke/github/GHRelease.java
+++ b/src/main/java/org/kohsuke/github/GHRelease.java
@@ -275,7 +275,9 @@ public class GHRelease extends GHObject {
         String url = getUploadUrl();
         // strip the helpful garbage from the url
         int endIndex = url.indexOf('{');
-        if (endIndex != -1) url = url.substring(0, endIndex);
+        if (endIndex != -1) {
+            url = url.substring(0, endIndex);
+        }
         url += "?name=" + URLEncoder.encode(filename, "UTF-8");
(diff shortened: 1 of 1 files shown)


src/main/java/org/kohsuke/github/GHRelease.java
```diff
@@ -274,7 +274,8 @@ public GHAsset uploadAsset(String filename, InputStream stream, String contentTy
         Requester builder = owner.root().createRequest().method("POST");
         String url = getUploadUrl();
         // strip the helpful garbage from the url
-        url = url.substring(0, url.indexOf('{'));
+        int endIndex = url.indexOf('{');
+        if (endIndex != -1) url = url.substring(0, endIndex);
```
bitwiseman, line 278: We use curly braces in all cases.
```suggestion
        int endIndex = url.indexOf('{');
        if (endIndex != -1) {
            url = url.substring(0, endIndex);
        }
```

PR #1779, opened 2024-01-10
Fix `GHFileNotFoundException` when getting commits from PR found in search

NlightNFotis:
# Description

Fixes #1778.

The issue summarily was that a compensation mechanism for sourcing a URL for a query was generating the wrong URL, that was then being used by other parts of the library to create further queries that the Github API would then reject.

This fixes the issue by doing a string search and replace after the string has been first sourced. It has been observed locally that this is an effective fix for the issue described.

Please consider the above a proposed change. If the project maintainers think that this is not an appropriate fix, I'm happy to rework this (if minor changes are needed) or close the PR (if major changes are needed, that are better to be done by someone more familiar with the codebase than myself).

Thank you,

# Before submitting a PR:

- [x] Changes must not break binary backwards compatibility. If you are unclear on how to make the change you think is needed while maintaining backward compatibility, [CONTRIBUTING.md](CONTRIBUTING.md) for details.
- [x] Add JavaDocs and other comments explaining the behavior. 
- [ ] When adding or updating methods that fetch entities, add `@link` JavaDoc entries to the relevant documentation on https://docs.github.com/en/rest . 
- [x] Add tests that cover any added or changed code. This generally requires capturing snapshot test data. See [CONTRIBUTING.md](CONTRIBUTING.md) for details.
- [x] Run `mvn -D enable-ci clean install site` locally. If this command doesn't succeed, your change will not pass CI.
- [x] Push your changes to a branch other than `main`. You will create your PR from that branch.

# When creating a PR: 

- [ ] Fill in the "Description" above with clear summary of the changes. This includes:
  - [ ] If this PR fixes one or more issues, include "Fixes #<issue number>" lines for each issue. 
  - [ ] Provide links to relevant documentation on https://docs.github.com/en/rest where possible. If not including links, explain why not.
- [ ] All lines of new code should be covered by tests as reported by code coverage. Any lines that are not covered must have PR comments explaining why they cannot be covered. For example, "Reaching this particular exception is hard and is not a particular common scenario."
- [x] Enable "Allow edits from maintainers".

PR #1810, opened 2024-03-09
Make sure we refresh the PRs with the PR API (and not the issues API)

gsmet:
When a PR was loaded from a search, the refresh() method was reloading information from the issues API, which would lead to some information not being refreshed properly, typically the mergeable state.

Related to https://github.com/hub4j/github-api/pull/1779#issuecomment-1966392546


gsmet:
@bitwiseman I created 3 PRs to fix some issues with `GHPullRequest` API behavior.

There might be some conflicts given I added some tests to the same place but they should be easy to fix.
If you prefer, I can bundle the commits in one single PR but I thought it would be better for release notes to keep them separated.

src/test/java/org/kohsuke/github/GHPullRequestTest.java
```diff
@@ -883,6 +883,41 @@ public void checkPullRequestReviewer() throws IOException {
...
+                .isOpen()
+                .titleLike(prName)
+                .list()
+                .toList()
+                .get(0);
+
+        pullRequestFromSearchResults.getMergeableState();
+
+        // wait a bit for the mergeable state to get populated
+        Thread.sleep(5000);
```
bitwiseman, line 912: Please implement `await()` in some form instead of hard sleeps.  Waiting is only needed when recording, not during CI tests. 

https://github.com/hub4j/github-api/blob/main/src/test/java/org/kohsuke/github/GHWorkflowRunTest.java#L574-L584

bitwiseman: This is broader problem.  Filed #1815.

bitwiseman (approved):
The thread sleeps are a broader problem.  If you want to switch to await, cool. If not, merge freely.

src/test/java/org/kohsuke/github/GHPullRequestTest.java
gsmet: Yeah I agree on that. I actually asked myself that and saw that there was some `Thread.sleep`  calls above so I supposed we decided to live with them. I'll try to find some cycles to clean this up.

PR #1821, opened 2024-03-20
Be a lot more flexible as for GHMemberChanges values

gsmet:
Unfortunately, things are not consistent between added and edited for the `member` event.

added provide a permission of some sort and a role name (sometimes...) edited only provides permission and it's not exactly the same ones

For now, I think we need to return strings to avoid issues until GitHub adjusts the API.

@bitwiseman you were entirely right, it's a big mess. My tests didn't show it, I don't know exactly why, maybe it depends on the roles/repositories or if you are using GH or GH Enterprise.

I kept the backwards compatibility and did a translation that is as good as possible.

Unfortunately, I have a user who is blocked by this so if it makes sense to you and you think it goes in the right direction, I would appreciate it being merged and if you could release a new release (sorry I know it's annoying but this inconsistency is really problematic and it was hard to figure it out :/).


gsmet:
I think the bridge methods are causing the coverage issue. I added some tests in the bridge methods test.

bitwiseman (approved):

PR #1824, opened 2024-03-21
Fix visibility of getters in GHRequestedAction

yrodiere:
# Description

`GHRequestedAction` has getters, but those getters have package-private (default) visibility, and thus are unusable.

This changes the visibility of those getters to `public`.

Looks like this part of the API is untested, and sorry, I didn't really have time to add new tests, but I figured this at least nudges things the right way :shrug: 

# Before submitting a PR:

# When creating a PR: 


src/main/java/org/kohsuke/github/GHRequestedAction.java
```diff
@@ -42,7 +42,7 @@ String getIdentifier() {
      *
      * @return the label
      */
-    String getLabel() {
+    public String getLabel() {
```
bitwiseman, line 45:
```suggestion
    String getLabel() {
```

src/main/java/org/kohsuke/github/GHRequestedAction.java
```diff
@@ -51,7 +51,7 @@ String getLabel() {
      *
      * @return the description
      */
-    String getDescription() {
+    public String getDescription() {
```
bitwiseman, line 54:
```suggestion
    String getDescription() {
```

bitwiseman (changes requested):
Related: 
* https://github.com/hub4j/github-api/pull/622
* https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=requested_action#check_run

The docs indicate that the `requested_action` field only contains `identifier` in the current release.  I guess we can make `getIdentifier()`  public, but it seems like making the other two public would be counterproductive.

yrodiere:
I applied your suggestions, thanks.

bitwiseman (approved):

PR #1844, opened 2024-05-16
feat: add ability to add checks with app ids

### LINKED ISSUE #1467: GHBranchProtection using deprecated field in required_status_checks
**Describe the bug**

[GHBranchProtection](https://github.com/hub4j/github-api/blob/4e103a6b192ca88c7948457e2b21ea3e6be9de89/src/main/java/org/kohsuke/github/GHBranchProtection.java#L245) is using the `contexts` field in RequiredStatusChecks. [This field](https://docs.github.com/en/enterprise-server@2.22/rest/reference/repos#update-branch-protection--parameters) was [deprecated in GitHub 3.0](https://docs.github.com/en/enterprise-server@3.0/rest/reference/branches#update-branch-protection--parameters).

> contexts(array of strings) | Required. Deprecated: The list of status checks to require in order to merge into this branch. If any of these checks have recently been set by a particular GitHub App, they will be required to come from that app in future for the branch to merge. Use checks instead of contexts for more fine-grained control.

As a result, repositories using App-authorized checks may experience surprising behaviour if two or more GitHub Apps are posting to the same check and a branch protection rule is (re)created.

**To Reproduce**

Create a branch protection rule with the API which requires one or more checks. Submit the required check from multiple GitHub Apps. Merging is blocked for one of the Apps.

image

**Expected behavior**

Checks can come from any PAT or GitHub App by default.

The new `checks` array should be passed to allow configuration of the required GitHub App for a check.

tginiotis-at-work:
# Description

Fixes #1467

Add ability to receive and specify App IDs for required checks in branch protection rules.

# Before submitting a PR:

# When creating a PR: 


src/main/java/org/kohsuke/github/GHBranchProtection.java
```diff
@@ -221,6 +221,54 @@ public boolean isEnabled() {
...
 
+    /**
+     * The type Check.
+     */
+    public static class Check {
+        @JsonProperty
+        private String context;
+
+        @JsonProperty
+        private Integer app_id;
```
bitwiseman, line 232:
```suggestion
        private String context;
        private Integer appId;
```

src/main/java/org/kohsuke/github/GHBranchProtection.java
```diff
@@ -221,6 +221,54 @@ public boolean isEnabled() {
...
+         * @param context
+         *            the context string of the check
+         * @param appId
+         *            the application ID the check is supposed to come from
+         */
+        public Check(String context, Integer appId) {
+            this.context = context;
+            this.app_id = appId;
+        }
+
```
bitwiseman, line 252:
```suggestion
```

src/main/java/org/kohsuke/github/GHBranchProtectionBuilder.java
```diff
@@ -62,11 +81,24 @@ public GHBranchProtectionBuilder addRequiredChecks(Collection<String> checks) {
...
     }
 
+    /**
+     * Add required checks gh branch protection builder.
+     *
+     * @param checks
+     *            the checks
+     * @return the gh branch protection builder
+     */
+    public GHBranchProtectionBuilder addRequiredChecksWithAppIds(GHBranchProtection.Check... checks) {
```
bitwiseman, line 97:
```suggestion
    public GHBranchProtectionBuilder addRequiredChecks(GHBranchProtection.Check... checks) {
```

src/main/java/org/kohsuke/github/GHBranchProtection.java
tginiotis-at-work: If this has no constructor, how can the user use the `GHBranchProtectionBuilder.addRequiredChecksWithAppIds` method, which accepts this object?

bitwiseman: Jackson is magic.  It will do the binding for you.

tginiotis-at-work: I don't think Jackson is related to the use case I am referring to. Here is an example of the use-case:

https://github.com/hub4j/github-api/blob/3882581deaedb92366c5a7b62b69a26c41f1113e/src/test/java/org/kohsuke/github/GHBranchProtectionTest.java#L203-L204

Or do you mean the user should use json strings and then Jackson to instantiate objects to pass in as arguments to that method?

bitwiseman: Ah, I see.  My mistake.

src/main/java/org/kohsuke/github/GHBranchProtectionBuilder.java
```diff
@@ -50,8 +50,27 @@ public class GHBranchProtectionBuilder {
...
+     *            the checks
+     * @return the gh branch protection builder
+     */
+    @Deprecated
     public GHBranchProtectionBuilder addRequiredChecks(Collection<String> checks) {
-        getStatusChecks().contexts.addAll(checks);
+        if (!(getStatusChecks() instanceof StatusChecksDeprecated)) {
+            statusChecks = new StatusChecksDeprecated();
+        }
+        ((StatusChecksDeprecated) getStatusChecks()).contexts.addAll(checks);
```
bitwiseman, line 73: If someone calls both `addRequiredChecks` and `addRequiredStatusChecks`, only the last call will win.  
How about something like this:
```suggestion
        getStatusChecks(StatusChecksDeprecated.class).contexts.addAll(checks);
```

src/main/java/org/kohsuke/github/GHBranchProtectionBuilder.java
```diff
@@ -532,7 +564,7 @@ private Restrictions getRestrictions() {
 
     private StatusChecks getStatusChecks() {
         if (statusChecks == null) {
-            statusChecks = new StatusChecks();
+            statusChecks = new StatusChecksWithAppId();
         }
         return statusChecks;
     }
```
bitwiseman, line 570: Suggest something like this:
```
    <T extends StatusCheck> T getStatusChecks(Class<T> clazz) {
        if (statusChecks == null) {
            try {
                statusChecks = clazz.getDeclaredConstructor().newInstance();
            } catch (Exception ignored) {
            }
        }
        if (!clazz.isInstance(statusChecks) {
            throw new GHException("Cannot use checks and context status checks");
        } 
        return (T) statusChecks;
    }
```

src/main/java/org/kohsuke/github/GHBranchProtectionBuilder.java
```diff
@@ -50,8 +50,27 @@ public class GHBranchProtectionBuilder {
      *            the checks
      * @return the gh branch protection builder
      */
+    public GHBranchProtectionBuilder addRequiredChecksWithAppIds(Collection<GHBranchProtection.Check> checks) {
+        if (!(getStatusChecks() instanceof StatusChecksWithAppId)) {
+            statusChecks = new StatusChecksWithAppId();
+        }
+        ((StatusChecksWithAppId) getStatusChecks()).checks.addAll(checks);
```
bitwiseman, line 57:
```suggestion
         getStatusChecks(StatusChecksWithAppId.class).checks.addAll(checks);
```

bitwiseman (changes requested):
Minor tweaks.  Thanks!

src/main/java/org/kohsuke/github/GHBranchProtectionBuilder.java
```diff
@@ -50,8 +50,27 @@ public class GHBranchProtectionBuilder {
      *            the checks
      * @return the gh branch protection builder
      */
+    public GHBranchProtectionBuilder addRequiredChecksWithAppIds(Collection<GHBranchProtection.Check> checks) {
```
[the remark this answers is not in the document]
tginiotis-at-work: changed

src/main/java/org/kohsuke/github/GHBranchProtectionBuilder.java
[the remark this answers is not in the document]
tginiotis-at-work: changed

src/main/java/org/kohsuke/github/GHBranchProtectionBuilder.java
tginiotis-at-work: https://github.com/hub4j/github-api/pull/1844#discussion_r1633525350

src/main/java/org/kohsuke/github/GHBranchProtectionBuilder.java
tginiotis-at-work: Apparently the deprecated API can be implemented in the new API. From [the docs](https://docs.github.com/en/enterprise-server@3.0/rest/reference/branches#update-branch-protection--parameters):
image

I underlined the same logic between the old and new API.

I.e. you need to omit the `app_id` field to do it the old way. I've added the `@JsonInclude(JsonInclude.Include.NON_NULL)` to instruct jackson to omit it in that case, and a comment explaining the special "app_id" values.

So I committed a change which allows the use of both the old and new user methods at the same time, preserving both their effects.

How does that look to you?

bitwiseman (approved):

Issue #1805, opened 2024-03-01
Handle rate limit errors appropriately :: GitHub Doesn't always send `Retry-After` header for secondary rate limits


ranma2913:
**Describe the bug**
The [GitHub Docs](https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api?apiVersion=2022-11-28) now say:

> Otherwise, wait for at least one minute before retrying. If your request continues to fail due to a secondary rate limit, wait for an exponentially increasing amount of time between retries, and throw an error after a specific number of retries.

There's also a whole complex list [About secondary rate limits](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#about-secondary-rate-limits). 

**To Reproduce**
Reach Secondary Quota Limits:
* At least 5 GET rps
* At least 1.4 POST rps
* Run until you get Secondary Quota Error: 

> "You have exceeded a secondary rate limit and have been temporarily blocked from content creation. Please retry your request again later"

**Expected behavior**
GitHub SHOULD return a retry-after header but it does not. So I've thought of a couple ideas:
1. Make the GitHubConnectorResponseErrorHandler.isError() method public down the stack so users (me) can customize the behavior. 
2. Update the private GitHubAbuseLimitHandler.isError() logic to handle no `Retry-After` header as described in the [GitHub Docs](https://docs.github.com/en/rest/using-the-rest-api/best-practices-for-using-the-rest-api?apiVersion=2022-11-28). This may not be ideal because note everyone is running a long-running project like I am so they may not be OK with a default 1-minute sleep before throwing a RetryRequestException.  

**Desktop (please complete the following information):**
 - OS: MacOS
 - JVM: Eclipse Temurin 21
 - Spring-Boot 3.2.3

**Additional context**
BTW I'm using a GitHub Enterprise User Account accessing GH Enterprise resources so maybe that's why there's no `Retry-After` header?

Another Idea is related to `rateLimitChecker.checkRateLimit(this, request.rateLimitTarget());` inside the GitHubClient.sendRequest method. If it could also pass the GitHub request instead of just the rateLimitTarget then I could configure more smart ratelimit checker which tracks api calls of differing "points" to reduce throughput of the 'create content' requests which GitHub limits much more heavily than Read requests.

bitwiseman:
@ranma2913

Thanks for filing this.

[...]

PR #1853, opened 2024-06-14
Fix API calls failing because of secondary api limits

jeetchoudhary:
**Issue**: After upgrading GitHub Enterprise Server (GHES) from version 3.9 to 3.11 and 3.12, we encountered issues with failing API calls during Git search (and other)operations. Further investigation revealed that the newer GHES versions consume more CPU resources, leading to secondary rate limits being reached more frequently. Below is a quote from GitHub Enterprise Support:
"Customers planning to upgrade to GitHub Enterprise Server (GHES) version 3.11 or higher should be aware of the potential for increased CPU utilization. If a system's peak CPU usage is already high, typically above 65-70%, it is strongly advisable to scale up the server resources before upgrading. This is to accommodate the additional load from new features and improvements that are part of the latest releases."

Moreover, we realized that GitHub does not guarantee the presence of the `Retry-After` header. However, the `gh-limited-by` header is included in the response when the error is due to rate-limiting

We got ourselves unblocked after making this change.   I see the issue is reported by another user here as well 
https://github.com/hub4j/github-api/issues/1805

**Fix:**  The intention is not to change any existing functionality. So ended up adding an additional optional check, so that the API client can wait for another minute before retrying the API call if either of the headers are present. With the new change the client will execute this block https://github.com/hub4j/github-api/blob/main/src/main/java/org/kohsuke/github/AbuseLimitHandler.java#L82-L93 which is not happening right now


bitwiseman (changes requested):
Needs test to cover the new functionality.  Otherwise looks good.

jeetchoudhary:
image

bitwiseman (approved):

PR #1860, opened 2024-06-22
Begin 2.x release train

bitwiseman:
# Description

I would love for `github-api` v2.x to be a larger change that has major new features.  

However, the most pressing need is clearing technical debt for better stability and ease of adding new features.  These changes break  binary compatibility requiring a major version bump.

v2.0 includes: 
* Drop support for Java 8
* Remove functional dependencies on `HttpURLConnection`
* Remove **most** deprecated classes and methods 
* Remove all "bridge methods"

# Before submitting a PR:

# When creating a PR: 


src/main/java/org/kohsuke/github/GHObject.java
```diff
@@ -79,36 +78,19 @@ public Map<String, List<String>> getResponseHeaderFields() {
...
 
-    /**
-     * Gets html url.
-     *
-     * @return URL of this object for humans, which renders some HTML.
-     * @throws IOException
-     *             on error
-     */
-    @WithBridgeMethods(value = String.class, adapterMethod = "urlToString")
-    public abstract URL getHtmlUrl() throws IOException;
```
bitwiseman, line 110: TODO: verify this is the right thing to do.   Not deprecated... 

Maybe just add the field and implement this.  Unclear why this was abstract.

src/main/java/org/kohsuke/github/GHMyself.java
```diff
@@ -65,7 +47,7 @@ public List<String> getEmails() throws IOException {
      * @throws IOException
      *             the io exception
      */
-    public List<GHEmail> getEmails2() throws IOException {
+    public List<GHEmail> getEmails() throws IOException {
```
bitwiseman, line 50: TODO: something other than this. 
Need to mark this deprecated before making this change.
```suggestion
    @Deprecated
    public List<GHEmail> getEmails2() throws IOException {
```

src/main/java/org/kohsuke/github/GHOrganization.java
```diff
@@ -768,17 +597,14 @@ public PagedIterable<GHEventInfo> listEvents() throws IOException {
...
      * Lists up all the repositories using the specified page size.
      *
-     * @param pageSize
-     *            size for each page of items returned by GitHub. Maximum page size is 100. Unlike
-     *            {@link #getRepositories()}, this does not wait until all the repositories are returned.
      * @return the paged iterable
      */
     @Override
-    public PagedIterable<GHRepository> listRepositories(final int pageSize) {
+    public PagedIterable<GHRepository> listRepositories() {
```
bitwiseman, line 603: TODO: Deprecate this before removing the parameter without replacement.

src/main/java/org/kohsuke/github/GHObject.java
bitwiseman: Decided to go ahead with this removal. If it breaks people this can be rolled back as a non-abstract method that returns `null`.

src/main/java/org/kohsuke/github/GHOrganization.java
bitwiseman: Keeping the existing methods. Refactored for future removal.

PR #1885, opened 2024-06-27
Added all possible status values from GH api spec.

mpetris:
Added all possible github workflow run status values specified in https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28

# Description

The Github REST API allows more values than available in the Status enum for GitWorkflowRuns.

# Before submitting a PR:

- [x ] Changes must not break binary backwards compatibility. If you are unclear on how to make the change you think is needed while maintaining backward compatibility, [CONTRIBUTING.md](CONTRIBUTING.md) for details.
- [ x] Add JavaDocs and other comments explaining the behavior. 
- [ x] When adding or updating methods that fetch entities, add `@link` JavaDoc entries to the relevant documentation on https://docs.github.com/en/rest . 
- [x ] Run `mvn -D enable-ci clean install site` locally. If this command doesn't succeed, your change will not pass CI.

# When creating a PR: 

- [ x] Fill in the "Description" above with clear summary of the changes. This includes:
- [x ] Enable "Allow edits from maintainers".

### DIFF
--- a/src/main/java/org/kohsuke/github/GHWorkflowRun.java
+++ b/src/main/java/org/kohsuke/github/GHWorkflowRun.java
@@ -502,6 +502,28 @@ public class GHWorkflowRun extends GHObject {
         IN_PROGRESS,
         /** The completed. */
         COMPLETED,
+        /** The action required. */
+        ACTION_REQUIRED,
+        /** The cancelled. */
+        CANCELLED,
+        /** The failure. */
+        FAILURE,
+        /** The neutral. */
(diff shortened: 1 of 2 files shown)


bitwiseman (changes requested):
This PR exposes a hole in our Enum tests - which should require that at least we have a known count of enum values. 

:sob:  It would be great to have tests that actually use any of these values.

bitwiseman (approved):

PR #1895, opened 2024-07-04
Accept 429 response codes as an indication of the secondary rate limit being exceeded

### LINKED ISSUE #1805: Handle rate limit errors appropriately :: GitHub Doesn't always send `Retry-After` header for secondary rate limits

### LINKED ISSUE #1842: Github Repository Search resulting in Secondary Rate Limit
**Describe the bug**
Getting the below error when searching for repositories with a keyword that returns a lot of repositories in the result. My usecase can be fulfilled by just getting say top 50 results, but I do not see any way to do that on the server side but only post getting the results.
```
org.kohsuke.github.HttpException: {
  "documentation_url": "https://docs.github.com/free-pro-team@latest/rest/overview/rate-limits-for-the-rest-api#about-secondary-rate-limits",
  "message": "You have exceeded a secondary rate limit. Please wait a few minutes before you try again. If you reach out to GitHub Support for help, please include the request ID ...."
}
```

I was trying the search for repositories within my org and keywords that matched even around 850 repositories were hitting this rate limit. The github documentation quotes a much larger number.

I even tried the `in:name` and `in:displayName` within the query but even that did not seem to help.

**To Reproduce**
```
new GitHubBuilder().withAppInstallationToken(token).build()
                .searchRepositories()
                .org(orgName)
                .q(repoName)
                .list();
```
Just pass in a repo name like "test" or something that results in a lot of repository results in the above code.

**Expected Behaviour**
It should return the repositories or else provide a clear reason of why the error. Or else there should be a way to limit the search results on the GitHub side itself. 

**Additional Info**
Version: org.kohsuke:github-api:1.316

holly-cummins:
Resolves #1842.
Resolves #1805. 

# Description

If a user uses this library to drive the GitHub API hard, there will quite often be failures caused by secondary rate limits being breached (#1842 and #1805). There’s been some discussion of headers on #1842, and also of pluggable 'handle the rate limit' modules, but I think there’s a more fundamental problem with the handling of rate limits. (The good news is this means there's a simpler solution, too.)

The `AbuseLimitHandler` checks for a rate limit violation by looking for a 403 header and some headers:  

```
@Override
boolean isError(@Nonnull GitHubConnectorResponse connectorResponse) {
    return isForbidden(connectorResponse) && hasRetryOrLimitHeader(connectorResponse);
}
```

`HTTP_FORBIDDEN` is a 403. The [GitHub docs](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit) indicate that limits from GitHub are signalled with either 403 or `TOO_MANY_REQUESTS` responses (429). In my experience, I've only seen 429s.

### What about headers? 

The GitHub docs docs suggest that any headers are optional, and that in the absence of headers the code should just wait a minute (which, luckily, is what the current implementation already does):

>  If the retry-after response header is present, you should not retry your request until after that many seconds has elapsed. If the x-ratelimit-remaining header is 0, you should not retry your request until after the time, in UTC epoch seconds, specified by the x-ratelimit-reset header. Otherwise, wait for at least one minute before retrying. If your request continues to fail due to a secondary rate limit, wait for an exponentially increasing amount of time between retries, and throw an error after a specific number of retries.

Based on these docs, I think it's sufficient to count a 429 as a rate limit response, even if headers are missing. Semantically, a 429 is an indication of an abuse detection/rate limit scenario, and should be handled as such.

### What about the rate limit handler?

`GitHubRateLimitHandler` also requires a 403 to 'count' a response as rate limited:

```
@Override
boolean isError(@NotNull GitHubConnectorResponse connectorResponse) throws IOException {
connectorResponse.statusCode() == HttpURLConnection.HTTP_FORBIDDEN
            && "0".equals(connectorResponse.header("X-RateLimit-Remaining"));
```

I didn't update this one, since the secondary rate limit checks are now correctly being handled on the abuse path with my changes. It might be worth updating it, or also even considering whether the distinction between the abuse and rate limit paths is so fuzzy that having both is just maintenance overhead. But that’s a different discussion. 

# Before submitting a PR:

- [ X] Changes must not break binary backwards compatibility. If you are unclear on how to make the change you think is needed while maintaining backward compatibility, [CONTRIBUTING.md](CONTRIBUTING.md) for details.
- [ X] Add JavaDocs and other comments explaining the behavior. 
- [ X] Push your changes to a branch other than `main`. You will create your PR from that branch.

# When creating a PR: 

  - [X ] If this PR fixes one or more issues, include "Fixes #<issue number>" lines for each issue. 
  - [ X] Provide links to relevant documentation on https://docs.github.com/en/rest where possible. If not including links, explain why not.
- [X ] All lines of new code should be covered by tests as reported by code coverage. Any lines that are not covered must have PR comments explaining why they cannot be covered. For example, "Reaching this particular exception is hard and is not a particular common scenario."


src/main/java/org/kohsuke/github/GitHubAbuseLimitHandler.java
```diff
@@ -29,7 +29,19 @@ public abstract class GitHubAbuseLimitHandler extends GitHubConnectorResponseErr
      */
     @Override
     boolean isError(@Nonnull GitHubConnectorResponse connectorResponse) {
-        return isForbidden(connectorResponse) && hasRetryOrLimitHeader(connectorResponse);
+        return isTooManyRequests(connectorResponse)
+                || (isForbidden(connectorResponse) && hasRetryOrLimitHeader(connectorResponse));
+    }
+
+    /**
+     * Checks if the response status code is TOO_MANY_REQUESTS (403).
```
bitwiseman, line 37:
```suggestion
     * Checks if the response status code is TOO_MANY_REQUESTS (429).
```

src/main/java/org/kohsuke/github/AbuseLimitHandler.java
```diff
@@ -89,7 +89,7 @@ public void onError(IOException e, HttpURLConnection uc) throws IOException {
         private long parseWaitTime(HttpURLConnection uc) {
             String v = uc.getHeaderField("Retry-After");
             if (v == null)
-                return 60 * 1000; // can't tell, return 1 min
+                return 61 * 1000; // can't tell, return 1 min plus 1 second for luck
```
bitwiseman, line 92: Is there a functional reason for this?

bitwiseman (changes requested):
This is a definite improvement.  

It still needs tests. Look at the existing abuse limit tests, which use manually constructed test data. Maybe just add one more step to the existing test rather than adding a whole new test. Up to you.

src/main/java/org/kohsuke/github/AbuseLimitHandler.java
holly-cummins: The [GitHub docs](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#exceeding-the-rate-limit) say "Otherwise, wait for at least one minute before retrying." I never know if the check is `< 60s` or `<=60s`, but it seems like waiting something that is unambiguously more than a minute is safest.

holly-cummins:
Oh good, I'm glad it looks sensible. I've added some tests and squashed.

src/main/java/org/kohsuke/github/AbuseLimitHandler.java
bitwiseman, line 92:
```suggestion
                // can't tell, wait for unambiguously over one minute per GitHub guidance
                return 61 * 1000;
```

bitwiseman (approved):

bitwiseman:
Please don't squash/force-push.  We can squash when merging the PR.

bitwiseman (changes requested):
Test failure needs fixing.

holly-cummins:
> Test failure needs fixing.

Oops, sorry about that. And noted about the squashing - different orgs have different requirements, so I have a squash habit, but I'll keep the commits separate. I agree it does make subsequent reviews easier!

src/test/java/org/kohsuke/github/AbuseLimitHandlerTest.java
```diff
@@ -412,7 +412,6 @@ public void onError(IOException e, HttpURLConnection uc) throws IOException {
                         assertThat(uc.getContentType(), equalTo("application/json; charset=utf-8"));
                         assertThat(uc.getContentLength(), equalTo(-1));
                         assertThat(uc.getHeaderFields(), instanceOf(Map.class));
-                        assertThat(uc.getHeaderFields().size(), Matchers.greaterThan(25));
```
bitwiseman: I wonder why I even had this in there...

holly-cummins: Yeah, I think there's an opportunity to make some of the tests a bit more concise and flexible without losing the important sense checks about the mocks' behaviour or the validation of the actual code ... but I didn't want to change too much existing code in an unrelated PR. But I perhaps should have been a bit more conservative about how much I copy-pasted.

holly-cummins:
The CI's failing because I didn't have enough coverage, even with the new test. To get the `parseWaitTime` method to be testable, I had to do a bit of refactoring to make it visible to the test code. I also added a new test to cover both paths, and then I realised that the original code could fail with a `NumberFormatException`, because `Retry-After` can [be a date](https://www.rfc-editor.org/rfc/rfc9110#field.retry-after). So I added new code and tests to handle that case. I also did a bit of refactoring of the test code and removing assertions about the mock that I was pretty sure we definitely didn't care about.

bitwiseman (approved):
Sorry for the slow response cycle.

src/main/java/org/kohsuke/github/AbuseLimitHandler.java
```diff
@@ -105,4 +101,26 @@ public void onError(IOException e, HttpURLConnection uc) throws IOException {
             throw e;
         }
     };
+
+    /*
+     * Exposed for testability. Given an http response, find the retry-after header field
+     * and parse it as either a number or a date (the spec allows both).
+     * If no header is found, wait for a reasonably amount of time.
+     */
```
bitwiseman, line 109:
```suggestion
    /*
     * Exposed for testability. Given an http response, find the retry-after header field and parse it as either a
     * number or a date (the spec allows both). If no header is found, wait for a reasonably amount of time.
     */
```

src/test/java/org/kohsuke/github/AbuseLimitHandlerTest.java
```diff
@@ -349,16 +342,191 @@ public void onError(IOException e, HttpURLConnection uc) throws IOException {
...
+            getTempRepository();
+            fail();
+        } catch (Exception e) {
+            assertThat(e, instanceOf(HttpException.class));
+            assertThat(e.getMessage(), equalTo("Abuse limit reached"));
+        }
+        assertThat(mockGitHub.getRequestCount(), equalTo(2));
+    }
+
+    @Test
```
bitwiseman, line 441:
```suggestion
    /**
     * Tests the behavior of the GitHub API client when the abuse limit handler with a date retry.
     *
     * @throws Exception
     *             if any error occurs during the test execution.
     */
    @Test
```

src/test/java/org/kohsuke/github/AbuseLimitHandlerTest.java
```diff
@@ -349,16 +342,191 @@ public void onError(IOException e, HttpURLConnection uc) throws IOException {
...
```
bitwiseman, line 493:
```suggestion
    /**
     * Tests the behavior of the GitHub API client when the abuse limit handler with a no retry after header.
     *
     * @throws Exception
     *             if any error occurs during the test execution.
     */
    @Test
```

src/main/java/org/kohsuke/github/AbuseLimitHandler.java
```diff
@@ -105,4 +101,22 @@ public void onError(IOException e, HttpURLConnection uc) throws IOException {
...
+            // can't tell, wait for unambiguously over one minute per GitHub guidance
+            return 61 * 1000;
+        }
+
+        try {
+            return Math.max(1000, Long.parseLong(v) * 1000);
+        } catch (NumberFormatException nfe) {
+            // The retry-after header could be a number in seconds, or an http-date
+            ZonedDateTime zdt = ZonedDateTime.parse(v, DateTimeFormatter.RFC_1123_DATE_TIME);
+            return ChronoUnit.MILLIS.between(ZonedDateTime.now(), zdt);
```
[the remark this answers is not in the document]
holly-cummins: Thanks! (And thanks for explaining about the diff to the response time, that makes a lot of sense.)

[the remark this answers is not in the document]
bitwiseman: Sure thing.  If you have time to circle back and address #1909 , that would be awesome.  Thanks for the great contribution!

PR #1897, opened 2024-07-15
Add functionality to get active rules for a branch

ihrigb:
# Description

Add methods to get all active rules for a branch.

# Before submitting a PR:

# When creating a PR: 


src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -620,9 +620,12 @@ public Map<String, Long> listLanguages() throws IOException {
...
     public String getOwnerName() {
-        // consistency of the GitHub API is super... some serialized forms of GHRepository populate
-        // a full GHUser while others populate only the owner and email. This later form is super helpful
-        // in putting the login in owner.name not owner.login... thankfully we can easily identify this
+        // consistency of the GitHub API is super... some serialized forms of
+        // GHRepository populate
+        // a full GHUser while others populate only the owner and email. This later form
+        // is super helpful
+        // in putting the login in owner.name not owner.login... thankfully we can
+        // easily identify this
```
bitwiseman, line 628: Could you revert these extraneous formatting changes?

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -3649,6 +3653,23 @@ public List<GHRepositoryTrafficTopReferralSources> getTopReferralSources() throw
...
+     * Get all active rules that apply to the specified branch
+     * (https://docs.github.com/en/rest/repos/rules?apiVersion=2022-11-28#get-rules-for-a-branch).
+     *
+     * @param branch
+     *            the branch
+     * @return the rules for branch
+     * @throws IOException
+     *             the io exception
+     */
+    public List<GHRepositoryRule> getRulesForBranch(String branch) throws IOException {
```
bitwiseman, line 3666: This endpoint uses pagination.  
```suggestion
    public PagedIterable<GHRepositoryRule> listRulesForBranch(String branch) throws IOException {
```

src/main/java/org/kohsuke/github/GHRepositoryRule.java
```diff
@@ -0,0 +1,692 @@
...
+            RulesetSourceType ruleset_type_source,
+            String ruleset_source,
+            long ruleset_id,
+            Map<String, JsonNode> parameters) {
+        this.type = type;
+        this.ruleset_source_type = ruleset_type_source;
+        this.ruleset_source = ruleset_source;
+        this.ruleset_id = ruleset_id;
+        this.parameters = new HashMap<>(parameters);
+    }
```
bitwiseman, line 53: Jackson doesn't require a constructor to populate the fields in a record. The lack of code coverage proves Jackson truly not using this. Remove it.

src/main/java/org/kohsuke/github/GHRepositoryRule.java
```diff
@@ -0,0 +1,692 @@
...
+
+    /**
+     * The type of the ruleset.
+     */
+    public static enum Type {
+        /**
+         * creation
+         */
+        creation,
```
bitwiseman, line 111: In Java `enum` values use ALLCAPS. 
So, for example: 
```suggestion
        CREATION,
```

src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -3649,6 +3653,23 @@ public List<GHRepositoryTrafficTopReferralSources> getTopReferralSources() throw
...
+        return Arrays.asList(root().createRequest()
+                .method("GET")
+                .withUrlPath(getApiTailUrl("/rules/branches/" + branch))
+                .fetch(GHRepositoryRule[].class));
```
bitwiseman, line 3670: Are these repository rules or branch rules?  The rulesets are definitely repository rulesets, but I think these are the resulting rules for a specific branch.   For example you might have two different repository rulesets that apply to branch. I think the result of the call would be the union of those two rulesets, but you're going to new bunch of tests to show that.

src/main/java/org/kohsuke/github/GHRepositoryRule.java
```diff
@@ -0,0 +1,692 @@
...
+import java.util.function.Function;
+
+/**
+ * Represents a repository rule.
+ */
+public class GHRepositoryRule {
+    private Type type;
+    private RulesetSourceType ruleset_source_type;
+    private String ruleset_source;
+    private long ruleset_id;
```
bitwiseman, line 20: Java uses camelCase, Jackson automatically translates from the snake_case JSON names.
```suggestion
    private Type type;
    private RulesetSourceType rulesetSourceType;
    private String rulesetSource;
    private long rulesetId;
```

src/main/java/org/kohsuke/github/GHRepositoryRule.java
```diff
@@ -0,0 +1,692 @@
...
+    private Map<String, JsonNode> parameters;
```
bitwiseman, line 21: Ouch. I kind of hate the Response schema under https://docs.github.com/en/rest/repos/rules?apiVersion=2022-11-28#get-rules-for-a-branch . 

For Java, I prefer the API we expose for 
https://github.com/gheine/github-api/blob/main/src/main/java/org/kohsuke/github/GHBranchProtection.java but I'm not sure if it possible to translate. 

What we want is a way for people to easily walk the list of rules in a meaningful way.

src/main/java/org/kohsuke/github/GHRepositoryRule.java
```diff
@@ -0,0 +1,692 @@
...
+
+    /**
+     * Basic parameter for a ruleset.
+     *
+     * @param <T>
+     *            the type of the parameter
+     */
+    public abstract static class Parameter<T> implements Function<JsonNode, T> {
+
+        private final static ObjectMapper objectMapper = new ObjectMapper();
```
bitwiseman, line 319: Definitely do not create a new object mapper for this class.  Reuse the one in GitHubClient, it is configured to correctly handle json returned from GitHub.

bitwiseman (commented):
I like the idea you have going with `Parameter<T>` and the instances defined in the `Parameters` interface, but I'm not sure it's the right way to go.  

I noted a bunch of style and detail changes.  You can choose to do them or hold off until I wrap my head around the `Parameter` design and figure out how I feel about it.

ihrigb:
@bitwiseman I will take care of your suggestions.

As of the Parameter vs. Polymorphism topic I had to thoughts about it as well. The parameter way turned out to be more flexible, if GitHub does any change to the API (not sure, if that would happen with the current version being almost two years old). Using polymorphism would result in weird Java code with `instanceof` stuff and casting types. Happy to discuss this either way.

src/main/java/org/kohsuke/github/GHRepository.java
ihrigb: The endpoint provides the union of applicable org and repo rulesets for a branch. What I am not sure about is, if "normal" repository rules will be taken into account as well.

src/main/java/org/kohsuke/github/GHRepositoryRule.java
```diff
@@ -0,0 +1,656 @@
...
+     * @return the parameters
+     */
+    public <T> Optional<T> getParameter(Parameter<T> parameter) {
+        return Optional.ofNullable(this.parameters).map(p -> p.get(parameter.getKey())).map(parameter);
+    }
+
+    /**
+     * The type of the ruleset.
+     */
+    public static enum Type {
```
bitwiseman, line 73: Add `UNKNOWN`, make the `type` field a `String`, and use the EnumUtils to convert in the getter.  See other enums with `UNKNOWN` for examples.

src/main/java/org/kohsuke/github/GHRepositoryRule.java
```diff
@@ -0,0 +1,656 @@
...
+        /**
+         * code_scanning
+         */
+        CODE_SCANNING
+    }
+
+    /**
+     * The source of the ruleset type.
+     */
+    public enum RulesetSourceType {
```
bitwiseman, line 158: I'm not sure if you need `UNKNOWN` for this one, but better safe than breaking.

src/main/java/org/kohsuke/github/GHRepositoryRule.java
ihrigb: Not sure, if I get your idea, but do you want me to add `UNKNOWN` in case of the API returning something the client does not understand?

bitwiseman: Yes, like this.  We've seen multiple cases of GitHub adding new types and we don't want this library to throw when it encounters them. 
https://github.com/hub4j/github-api/blob/cab9bbb8cb37599362cb37f0c278ab262298c59b/src/main/java/org/kohsuke/github/GHCheckRun.java#L102-L122

bitwiseman: NOTE: we don't do this for all enums but this one seems like on a case where there are likely to be new parameter types sooner or later.

ihrigb: got it.

src/main/java/org/kohsuke/github/GHRepository.java
bitwiseman: In the GraphQL Schema they are defined as `type RepositoryRule implements Node {` so I guess this name is correct.

src/main/java/org/kohsuke/github/GHRepositoryRule.java
bitwiseman: I don't have a better solution for this at this time.  Until someone comes up with a better solution, I'm comfortable with taking what we have.

src/main/java/org/kohsuke/github/GHRepositoryRule.java
```diff
@@ -0,0 +1,667 @@
...
+
+    /**
+     * Available parameters for a ruleset.
+     */
+    public interface Parameters {
+        /**
+         * update_allows_fetch_and_merge parameter
+         */
+        public static final BooleanParameter UPDATE_ALLOWS_FETCH_AND_MERGE = new BooleanParameter(
+                "update_allows_fetch_and_merge");
```
bitwiseman, line 189: Which value of the `GHRepositoryRule.Type` enum  applies here?  Is it suppose to be `REQUIRED_LINEAR_HISTORY`? 

I'm fine with `Parameter<T>` taking string in the constructor, but  in `Parameters` I want those strings to be generated from `GHRepositoryRule.Type` to reduce the likelihood of the strings here getting out of sync with the enum values.

src/main/java/org/kohsuke/github/GHRepositoryRule.java
```diff
@@ -0,0 +1,667 @@
...
+         * @return the key
+         */
+        String getKey() {
+            return this.key;
+        }
+
+        @Override
+        public T apply(JsonNode jsonNode) {
+            if (jsonNode == null) {
+                return null;
```
bitwiseman, line 325: You might need to add a custom test to cover this line.

src/main/java/org/kohsuke/github/GHRepositoryRule.java
```diff
@@ -0,0 +1,667 @@
...
+         * @param path
+         *            the path
+         * @param ref
+         *            the ref
+         * @param repositoryId
+         *            the repository id
+         * @param sha
+         *            the sha
+         */
+        public WorkflowFileReference(String path, String ref, int repositoryId, String sha) {
```
bitwiseman, line 509: Jackson will do this without a public constructor.  Make it internal  or remove it entirely.

src/main/java/org/kohsuke/github/GHRepositoryRule.java
```diff
@@ -0,0 +1,667 @@
...
+    public static class WorkflowFileReference {
+        private String path;
+        private String ref;
+        private int repositoryId;
+        private String sha;
+
+        /**
+         * Instantiates a new workflow file reference.
+         */
+        public WorkflowFileReference() {
```
bitwiseman, line 494: Jackson will do this without a public constructor.  Make it internal  or remove it entirely.

src/main/java/org/kohsuke/github/GHRepositoryRule.java
```diff
@@ -0,0 +1,667 @@
...
+     */
+    public static class CodeScanningTool {
+        private AlertsThreshold alertsThreshold;
+        private SecurityAlertsThreshold securityAlertsThreshold;
+        private String tool;
+
+        /**
+         * Instantiates a new code scanning tool.
+         */
+        public CodeScanningTool() {
```
bitwiseman, line 564: Jackson will do this without a public constructor.  Make it internal  or remove it entirely.

src/main/java/org/kohsuke/github/GHRepositoryRule.java
```diff
@@ -0,0 +1,667 @@
...
+         * Instantiates a new code scanning tool.
+         *
+         * @param alertsThreshold
+         *            the alerts threshold
+         * @param securityAlertsThreshold
+         *            the security alerts threshold
+         * @param tool
+         *            the tool
+         */
+        public CodeScanningTool(AlertsThreshold alertsThreshold,
```
bitwiseman, line 577: Jackson will do this without a public constructor.  Make it internal  or remove it entirely.

src/main/java/org/kohsuke/github/GHRepositoryRule.java
```diff
@@ -0,0 +1,667 @@
...
+        String getKey() {
+            return this.key;
+        }
+
+        @Override
+        public T apply(JsonNode jsonNode) {
+            if (jsonNode == null) {
+                return null;
+            }
+            return GitHubClient.getObjectMapper().convertValue(jsonNode, getType());
```
bitwiseman, line 327: Is there a way to do this with the `ObjectReader` returned from [getMappingObjectReader](https://github.com/hub4j/github-api/blob/574a388e336efbed25c071077e9e5e45e741a93c/src/main/java/org/kohsuke/github/GitHubClient.java#L981) instead of adding a new method?

bitwiseman (changes requested):
Thanks for sticking with this through multiple rounds of feedback.  This is looking great, I think this is the last cycle and then we'll be able to merge.

src/main/java/org/kohsuke/github/GHRepositoryRule.java
ihrigb: I am not sure, what do do... Enum values cannot have static fields if you think about sth like `Type.REQUIRED_LINEAR_HISTORY.UPDATE_ALLOWS_FETCH_AND_MERGE`.

ihrigb: @bitwiseman what should I do here?

bitwiseman (approved):

PR #1912, opened 2024-08-19
Populate GHUser before access a suspendedAt field value

styner9:
# Description
`GHUser#getSuspendedAt()` for a suspended user returns null if the object is not populated.
```kotlin
// for filter out suspended user

// not working: user object is not populated -> field value is always null
val users = github.listUsers()
     .filter { it.suspendedAt == null }
     .toList()

// working but tricky: call getCreatedAt() first -> populated -> getSuspendedAt() is not null
val users2 = github.listUsers()
    .filter { it.createdAt != null && it.suspendedAt == null }
    .toList()
```

This PR fixes
- Populate before access field value.
- ~Confirm field naming convention. (camelCase -> snake_case)~

FYI. This bug was introduced by PR #1906. (Released v1.324)

# Before submitting a PR:

# When creating a PR: 

### DIFF
--- a/src/main/java/org/kohsuke/github/GHUser.java
+++ b/src/main/java/org/kohsuke/github/GHUser.java
@@ -40,7 +40,7 @@ public class GHUser extends GHPerson {
     protected String ldap_dn;
 
     /** The suspended_at */
-    private String suspendedAt;
+    private String suspended_at;
 
     /**
      * Gets keys.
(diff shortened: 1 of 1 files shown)


src/main/java/org/kohsuke/github/GHUser.java
```diff
@@ -40,7 +40,7 @@ public class GHUser extends GHPerson {
     protected String ldap_dn;
 
     /** The suspended_at */
-    private String suspendedAt;
+    private String suspended_at;
```
bitwiseman, line 43: Don't need this. Jackson does this for you.

bitwiseman (changes requested):

src/main/java/org/kohsuke/github/GHUser.java
styner9: 👌  Rollbacked to `suspendedAt`.

bitwiseman (approved):

Issue #1908, opened 2024-08-15
Enable github-api to support GraalVM native images


klopfdreh:
**Describe the bug**
Due to some reflections you encounter errors during the runtime when `github-api` is used in a native image.

Example
```plain
at java.base@22.0.1/java.lang.invoke.LambdaForm$DMH/sa346b79c.invokeStaticInit(LambdaForm$DMH)\nCaused by: com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `org.kohsuke.github.GHRepository`: cannot deserialize from Object value (no delegate- or property-based Creator): this appears to be a native image, in which case you may need to configure reflection for the class that is to be deserialized\n at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 2]
```

**To Reproduce**
1. Build an application with Spring Boot Native and `github-api`
2. Perform a `native-image` build
3. Run the native application

**Expected behavior**
`github-api` should be used in a native image without any issues

**Desktop (please complete the following information):**
 N/A

**Additional context**
You could add `META-INF/native-image/<groupid>/<artifactid>/reflect-config.json` and describe the reflection usage:

Example:
```json
[
  {
    "name": "org.kohsuke.github.GHRepository",
    <settings for reflections>
  }
]
```

klopfdreh:
I saw a lot of other issues which are caused by the heavy usage of reflections. When I add all classes which are instantiated with reflections, like explained in the issue, I received the following error

```plain
java.lang.NullPointerException: The root GitHub reference for this instance is null. Probably caused by deserializing this class without using a GitHub instance. If you must do this, use the MappingObjectReader from GitHub.getMappingObjectReader().
```

bitwiseman:
@klopfdreh 
I don't have a lot of time to commit to this project currently. 

I've created a PR to create a v2.x That removes Java 8 and most of the reflection from this project. But it is already out of date and needs more attention to complete.

PRs welcome,  but sounds like the changes you're asking if I would be breaking changes. So we might want to include the 2.x effort.

klopfdreh:
Hey @bitwiseman - yes it seems to be better in 2.x then.

I think it all narrows down to the way objects are read and the root value is passed into the newly created objects with InjectableValues.

Sadly I just came across this issue as we are moving all of our projects to Spring Boot Native and one of it is using this GitHub-API.

[...]

PR #1914, opened 2024-08-23
feat: GraalVM support

klopfdreh:
# Description

This PR will enable the `github-api` to be supported by GraalVM. There have been two options discussed. 

1. Use Spring Boot RuntimeHintsRegistrar / Quarkus @BuildStep to generate the reflect-config.json / serialization-config.json in separate artifacts. 
2. Add the low level GraalVM metadata directly to the project. 
 
@bitwiseman preferred option 2 as it reduces the dependencies and lower the costs.

The See https://github.com/hub4j/github-api/issues/1908 for more information

# Before submitting a PR:

# When creating a PR: 


klopfdreh:
What I did:
1. I ran my application `aot-process` with `1.323` of `github-api` as dependency.
2. Because of the `RuntimeHintsRegistrar` mentioned here https://github.com/hub4j/github-api/issues/1908#issuecomment-2296488149 all reflection-config and serialization-config hints are created at ./target/spring-aot/main/resources/META-INF/<groupid>/<artifactid>/*.json
3. I copied all information which belongs to `org.kohsuke.github` packages into those configs provided by this PR.

Hey @gsmet - it would be very nice if you could check if I missed something in the configs. 😃

klopfdreh:
**Important:** If there are any new classes that requires reflections those need to be added to the configs as well as they are not auto generated via Spring / Quarkus

klopfdreh:
I had to force push as I forgot the `native-image` folder inside of `META-INF`

bitwiseman (changes requested):
@klopfdreh  
Thanks for giving this a try.  This PR avoids the dependency, but it introduces a manual step for any time a class is added or removed.  I'm not opposed this outright, but if you're going to go this route, there needs to be a test that checks that all expected classes are present in the file.

klopfdreh:
For me this sounds a bit odd that you want to handle both classes with/without reflections, as you already know from the reflect-config.json which classes use them, meaning all others don't use reflection.

You can't get a list with classes that require reflection as this is something you need to know and define via spring-aot.

I agree that it is much nicer to find out if you missed a class in the reflect-config.json during build time, but the only way to see if it is working is to start a native image in your pipeline (which is expensive).

Note: It is not that big of a deal to add all classes to reflect and serialization as it is like in the JVM. The performance impact is also very low (I tested it already).

What I suggest would be to add a test that uses spring-aot and checks via Registrar if all classes are mentioned in the reflect-config.json

gsmet:
In the Quarkus extension, we have a test that approximates things a bit, meaning we have a list of excluded classes and any time a class gets added, we have to decide if we exclude it or add it: https://github.com/quarkiverse/quarkus-github-api/blob/main/deployment/src/test/java/io/quarkiverse/githubapi/deployment/GitHubApiDotNamesTest.java#L28 .

It has worked well enough for us even if not ideal.

gsmet:
BTW, I still think that having an annotation and an annotation processor would make things a bit easier.

People wouldn't have to adjust the JSON file, which can be error prone, they would just have to add an annotation.

bitwiseman:
@klopfdreh 

> What I suggest would be to add a test that uses spring-aot and checks via Registrar if all classes are mentioned in the reflect-config.json

I agree to this suggestion. It is a bit on the bare minimum side, but I would accept this. 

Can the test also update the json file (and then report failure)? That would address @gsmet's concern about maintaining the json file.  

>In the Quarkus extension, we have a test that approximates things a bit, meaning we have a list of excluded classes and any time a class gets added, we have to decide if we exclude it or add it: https://github.com/quarkiverse/quarkus-github-api/blob/main/deployment/src/test/java/io/quarkiverse/githubapi/deployment/GitHubApiDotNamesTest.java#L28 .

This is basically lines up with what I was thinking around not registering classes for reflection which don't need it.  

What this test shows is that having an exclude list is not expensive and it lets us check "all classes _except a specific list_ are mentioned in the reflect-config.json".  If you're willing to do this @klopfdreh, that would be even better but not absolute required.

klopfdreh:
@bitwiseman - I am going to add the test, tomorrow. 

Even if this is a manual effort to add the entry into the reflect-config.json I would not let a test change the source code. (It is still only a test)

bitwiseman:
Then If possible make the output when the test fails clear instructions about what to change.

klopfdreh:
Hey @bitwiseman - I added the test now.

Two things which are important:

1. The test build requires java 17 as Spring Boot 3.x is used
2. ensure-java-1.8-class-library can't be used for animal-sniffer-maven-plugin anymore as Spring Boot uses Java 17

I have Java > 17 installed locally and already tested if you run `mvn clean verify` that if you not add a class to `reflect-config.json` / `serialization-config.json` or no-reflect-and-serialization-list the following error occurs

```plain
The class "org.kohsuke.github.AbstractBuilder" needs to be configured or excluded for reflection / serialization and was not mentioned in one of the following resources:

src/main/resources/META-INF/reflect-config.json - example:

  {
    "name": "org.kohsuke.github.AbstractBuilder",
    "allPublicFields": true,
    "allDeclaredFields": true,
    "queryAllPublicConstructors": true,
    "queryAllDeclaredConstructors": true,
    "allPublicConstructors": true,
    "allDeclaredConstructors": true,
    "queryAllPublicMethods": true,
    "queryAllDeclaredMethods": true,
    "allPublicMethods": true,
    "allDeclaredMethods": true,
    "allPublicClasses": true,
    "allDeclaredClasses": true
  }

src/main/resources/META-INF/serialization.json - example:

  {
    "name": "org.kohsuke.github.AbstractBuilder"
  }

src/test/resources/no-reflect-and-serialization-list - example:

  org.kohsuke.github.AbstractBuilder

Please add it to either no-reflect-and-serialization-list or to serialization.json and / or reflect-config.json
```

pom.xml
```diff
@@ -254,19 +255,32 @@
               java.net.http.*
             </ignores>
           </configuration>
-          <executions>
+          <!--<executions>
```
bitwiseman, line 258: Until we officially drop Java 8 support, I think this is required, but it might not be since we have a Java 8 test run. 
But if you're going to remove the only execution, remove the whole `<plugin>` rather than commenting out.

klopfdreh: I am going to remove it, now.

pom.xml
```diff
@@ -254,19 +255,32 @@
...
                 <goal>check</goal>
               </goals>
             </execution>
-          </executions>
+          </executions>-->
         </plugin>
       </plugins>
     </pluginManagement>
     <plugins>
+      <plugin>
```
bitwiseman, line 271: Add a comment explaining the presence of this plugin and other spring dependency. It is a non-standard use. 

In fact, please put it in a profile with activation like https://github.com/hub4j/github-api/blob/6cdb1cac5aa29d67b002e7796ae8a4bae2c487c5/pom.xml#L698 and comment the profile.  That will make it on by default but will group these two items together.

src/test/resources/no-reflect-and-serialization-list
```diff
@@ -0,0 +1,29 @@
+org.kohsuke.github.GHBranchSync
```
bitwiseman, line 1: Is this the full list of classes that don't get reflected or serialized? Or is this just an example?

klopfdreh: Yes it is. When I created the aot metadata during the test those classes were mentioned not to use serialization / reflection. (Good test 👍 😄)

We have to decide if we also put them into the serialization / reflection files like explained or leave them out.

pom.xml
klopfdreh: Because the `AotIntegrationTest` is executed by default with all other tests, the test would fail if you don't execute the plugin by default. Do you think it is really required to put it into a separate profile?

bitwiseman: No it's fine. I was just thinking of our to encapsulate it. No big deal.

src/test/resources/no-reflect-and-serialization-list
bitwiseman: I would expect this list to look more like https://github.com/quarkiverse/quarkus-github-api/blob/main/deployment/src/test/java/io/quarkiverse/githubapi/deployment/GitHubApiDotNamesTest.java#L28 . 

The ones listed here look more like ones that don't have coverage maybe? I would include all of these in json files.

klopfdreh: All right - I am going to adjust the lists accordingly.

klopfdreh: @bitwiseman / @gsmet just a small shout out. The packages `org.kohsuke.github.extras.authorization` and `org.kohsuke.github.extras.okhttp3` are not present anymore.

I adjusted the list according to the mentioned list here: https://github.com/hub4j/github-api/pull/1914#discussion_r1740202891

src/test/resources/no-reflect-and-serialization-list
```diff
@@ -0,0 +1,91 @@
+org.kohsuke.github.AbuseLimitHandler
```
bitwiseman, line 1: I'm still not sure this list is correct, but issues here can be addressed as bug fixes.

bitwiseman (approved):
@klopfdreh 
Great work!  

@gsmet  
This looks good to me.  Please review and approve if you agree.

bitwiseman:
I'm going to merge this. @gsmet if you have any changes, please file an issue and tag @klopfdreh .

PR #1935, opened 2024-09-15
Begin 2.x release train

bitwiseman:
# Description

I would love for github-api v2.x to be a larger change that has major new features.

However, the most pressing need is clearing technical debt for better stability and ease of adding new features. These changes break binary compatibility requiring a major version bump.

v2.0 includes:

* Drop support for Java 8
* Remove functional dependencies on HttpURLConnection
* Remove most deprecated classes and methods
* Remove all "bridge methods"

# Before submitting a PR:

# When creating a PR: 


src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -1676,33 +1533,6 @@ public GHPullRequest getPullRequest(int i) throws IOException {
...
-     * @param state
-     *            the state
-     * @return the paged iterable
-     * @deprecated Use {@link #queryPullRequests()}
-     */
-    @Deprecated
-    public PagedIterable<GHPullRequest> listPullRequests(GHIssueState state) {
-        return queryPullRequests().state(state).list();
-    }
-
```
bitwiseman, line 1705: @ihrigb 
Noted that this was not marked as Deprecated.  

Did a search on usages:
https://github.com/search?q=org%3Ajenkinsci+getPullRequests+path%3A%2F%5Esrc%5C%2Fmain%5C%2Fjava%5C%2F%2F+org.kohsuke.github&type=code

It looks like there's at least one usage in the wild, and since it is in Jenkins it will cause breaks for some time to come. 
https://github.com/jenkinsci/ghprb-plugin/blob/master/src/main/java/org/jenkinsci/plugins/ghprb/GhprbRepository.java#L145

There's a similar one for listPullRequests().  

Might consider bringing these back at least as bridge methods to not break Jenkins plugins.  On the other hand this is a 2.0, accepting some breakages might be acceptable.

PR #1969, opened 2024-10-03
Remove checked exceptions from most external APIs

bitwiseman:
# Description

# Before submitting a PR:

# When creating a PR: 

### DIFF
--- a/src/test/java/org/kohsuke/github/AppTest.java
+++ b/src/test/java/org/kohsuke/github/AppTest.java
@@ -14,6 +14,7 @@ import org.kohsuke.github.GHOrganization.Permission;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.UncheckedIOException;
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.*;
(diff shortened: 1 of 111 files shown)


src/test/java/org/kohsuke/github/AppTest.java
```diff
@@ -1608,8 +1609,8 @@ public void testRepoLabel() throws IOException {
             try {
                 t = r.getLabel("test");
                 fail("Test label should be deleted.");
-            } catch (IOException ex) {
-                assertThat(ex, instanceOf(FileNotFoundException.class));
+            } catch (UncheckedIOException ex) {
+                assertThat(ex.getCause(), instanceOf(FileNotFoundException.class));
```
bitwiseman, line 1613: This going to be a massive issue.  

Consumers of this library that expect to be able to detect missing items by catching `FileNotFoundException` will have to catch `UncheckedIOException` and then check the type of `e.getCause()`.  That's not so bad per se, but it is a significant change in behavior.    

We could handled this be replacing the generic `UncheckedIOException` with an `UncheckedFileNotFoundException`, but it would still be a significant change not something that could be used by existing consumers without updates.

Haarolean: Is it just one more thing people must consider and adjust when switching from 1.x to 2.x? As a maintainer, I would put a bold notice about all these things in a changelog.

Haarolean:
@bitwiseman have you decided not to do this or?

bitwiseman:
@Haarolean 
It has been three months. I have very limited time to devote to this project. Whatever direction the project goes in V2, even if it is this level of breaking change, it will not be done by me.  The branch is still available.

Haarolean:
> @Haarolean It has been three months. I have very limited time to devote to this project. Whatever direction the project goes in V2, even if it is this level of breaking change, it will not be done by me. The branch is still available.

I understand, thanks for a clarification and your effort!

PR #1972, opened 2024-10-09
Do not assume server time is in sync with local machine time on rate limit path

holly-cummins:
See discussion in #1909 and companion PR in #1971. On this test, I did try and cover the guard for missing dates with tests.

# Description

# Before submitting a PR:

- [X] Add JavaDocs and other comments explaining the behavior. 
- [X] When adding or updating methods that fetch entities, add `@link` JavaDoc entries to the relevant documentation on https://docs.github.com/en/rest . 
- [X] Add tests that cover any added or changed code. This generally requires capturing snapshot test data. See [CONTRIBUTING.md](CONTRIBUTING.md) for details.

# When creating a PR: 

  - [X] Provide links to relevant documentation on https://docs.github.com/en/rest where possible. If not including links, explain why not.
- [X] All lines of new code should be covered by tests as reported by code coverage. Any lines that are not covered must have PR comments explaining why they cannot be covered. For example, "Reaching this particular exception is hard and is not a particular common scenario."


src/main/java/org/kohsuke/github/GitHubRateLimitHandler.java
```diff
@@ -71,15 +79,29 @@ public void onError(GitHubConnectorResponse connectorResponse) throws IOExceptio
...
+        // thinks it is
+        String dateField = connectorResponse.header("Date");
+        ZonedDateTime now;
+        if (dateField != null) {
+            now = ZonedDateTime.parse(dateField, DateTimeFormatter.RFC_1123_DATE_TIME);
+        } else {
+            now = ZonedDateTime.now();
         }
-    };
+        return Math.max(MINIMUM_RATE_LIMIT_RETRY_MILLIS, (Long.parseLong(v) - now.toInstant().getEpochSecond()) * 1000);
```
bitwiseman, line 103: This is now mostly the same as the abuse limit wait parse code you submitted.  Would it make sense to extract this into a shared method?  Perhaps even make it public so others could use it?  Opened #1973 to track this.

bitwiseman (approved):

PR #1986, opened 2024-11-27
initial feature add

gitPushPuppets:
# Description

# Before submitting a PR:

# When creating a PR: 

### DIFF
--- a/src/main/java/org/kohsuke/github/GHRepository.java
+++ b/src/main/java/org/kohsuke/github/GHRepository.java
@@ -2898,7 +2898,8 @@ public class GHRepository extends GHObject {
 
     /**
      * Gets all custom property values that are set for a repository. Users with read access to the repository can use
-     * this endpoint. <a href="https://docs.github.com/en/rest/repos/custom-properties?apiVersion=2022-11-28#get-all-custom-property-values-for-a-repository">
+     * this endpoint. <a href=
+     * "https://docs.github.com/en/rest/repos/custom-properties?apiVersion=2022-11-28#get-all-custom-property-values-for-a-repository">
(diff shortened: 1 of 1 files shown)


src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -125,8 +125,11 @@ public GHRepository() {
     private GHRepository source, parent;
 
     private Boolean isTemplate;
+
     private boolean compareUsePaginatedCommits;
 
+    private Set<GHRepositoryCustomProperty> custom_properties;
```
bitwiseman, line 131: Thoughts:  I'm not sure that `Set` helps you here.  You might want to represent this as
* `List` - getting all of them
* `Map<String, String>` - the docs seem to indicate that you can't have multiple properties with the same `name` value on a repository or organization.  And the properties are just name/value pairs...
* `Map<String,CustomProperty>` - This might be a good choice for future proofing, in case GitHub decides to make CustomProperty records more complex in future.

For either of the `Map` options you need to check whether `key` value uniqueness case-insensitive, case-sensitive but restricted to ASCII, or literary any Unicode/UTF-8 String.

bitwiseman:
Looks reasonable. Only one comment so far.

src/main/java/org/kohsuke/github/GHRepository.java
gitPushPuppets: I got my custom properties returned to me using this code base referenced in a console app.

I was discussing map options with a friend of mine but haven't been able to get it to work yet. I am not familiar with how this repo handles object mapping and kept running into null errors.

PR #2015, opened 2025-01-21
GHPerson#getType() should call populate() only if type is null

### LINKED ISSUE #1816: 401 when trying to extract GHUser#type
Hi, a code like this

```
new GitHubBuilder()
        .withJwtToken("token")
        .build()
        .getApp()
        .getInstallationById(githubInstallationId)
        .getAccount()
        .getType()
```

leads to 401:
```
Caused by: org.kohsuke.github.HttpException: {"message":"Bad credentials","documentation_url":"https://docs.github.com/rest"}
	at org.kohsuke.github.GitHubConnectorResponseErrorHandler$1.onError(GitHubConnectorResponseErrorHandler.java:62) ~[github-api-unbridged-1.318.jar:na]
```

There are a few issues here:
1. The field is already initialized as "Organization" after calling `getAccount()`, so populating here is not necessary, perhaps.
2. `installation.getAccount().getUrl()` returns `https://api.github.com/users/<ORGANIZATION_LOGIN>`, which perhaps is invalid, because it's an organization, not a user.

anujhydrabadi:
# Description
While calling the app or app installations endpoints, `type` is returned in the response from GitHub for users and organizations. Do not call `populate()` if `type` is already set. Perhaps we can fix this at the design level to prevent these kinds of bugs (caused because we do not know what fields have been populated already) from occurring altogether, but making a quick fix for this in the meantime.
Fixes #1816 

# Before submitting a PR:

# When creating a PR: 


anujhydrabadi:
Please review. The codecov check is failing as it expects me to test the following code block too, which I feel will not be required. Please let me know if I need to add anything here.
```java
if (isOffline()) {
            return; // cannot populate, will have to live with what we have
        }
``` 
Thanks!

bitwiseman:
That code coverage hole is odd, but obviously not due to this change.

bitwiseman (approved):

PR #2036, opened 2025-02-17
Removed unused checked exceptions

rnveach:
# Description

Removes exceptions not thrown by the methods.


bitwiseman:
Removing exceptions... I thought this was a binary breaking change, but the API checks didn't fail for this change... 

I need to look at this further.

rnveach:
Do you want me to try and split this between main and test?

rnveach:
Documentation:
https://docs.oracle.com/javase/specs/jls/se11/html/jls-13.html#jls-13.4.21 (Java 11 JLS)
> Changes to the throws clause of methods or constructors do not break compatibility with pre-existing binaries; these clauses are checked only at compile time.

This says it is not a breaking binary change.

PR #2049, opened 2025-02-28
Revert removal of getPullRequests

### LINKED ISSUE #1957: GHRepository.getPullRequests(GHIssueState) not deprecated but removed from 2.x
https://github.com/hub4j/github-api/pull/1935/files#r1778850947 - Anchor

@ihrigb 
Noted that `GHRepository.getPullRequests(GHIssueState)` was not marked as Deprecated but was removed in `2.0-alpha-1`. 

Did a search on usages:
https://github.com/search?q=org%3Ajenkinsci+getPullRequests+path%3A%2F%5Esrc%5C%2Fmain%5C%2Fjava%5C%2F%2F+org.kohsuke.github&type=code

It looks like there's at least one usage in the wild, and since it is in Jenkins it will cause breaks for some time to come. 
https://github.com/jenkinsci/ghprb-plugin/blob/master/src/main/java/org/jenkinsci/plugins/ghprb/GhprbRepository.java#L145

There's a similar one for listPullRequests().  

Options:
* Bring some methods back as bridge methods in 2.0 to not break Jenkins plugins.  
* Change to 1.x that converts methods removed in 2.x to bridge methods - this would force projects to change their code in their next release while maintaining binary compatibility.  Maybe only convert some methods.  This might be a way to push some additional changes into 2.x.

On the other hand this is a 2.0 release, some incompatibility is to be expected.

bitwiseman:
# Description

Method was not marked as deprecated, but was removed. Adding it back to prevent breaks.

Fixes #1957

# Before submitting a PR:

# When creating a PR: 


src/main/java/org/kohsuke/github/GHRepository.java
```diff
@@ -1503,6 +1503,22 @@ public GHPullRequest getPullRequest(int number) throws IOException {
...
 
+    /**
+     * Retrieves all the pull requests of a particular state.
+     *
+     * @param state
+     *            the state
+     * @return the pull requests
+     * @throws IOException
+     *             the io exception
+     * @see #listPullRequests(GHIssueState) #listPullRequests(GHIssueState)
```
bitwiseman, line 1514:
```suggestion
```

Issue #1405, opened 2022-03-15
OutOfMemoryError on GHRepository.readZip due to lack of streaming 


blacelle:
**Describe the bug**
I encounter OutOfMemoryError over GHRepository.readZip while having reasonable free heap. 

**To Reproduce**
1. Consider a large Github repository (e.g. 50MB)
2. Rely on GHRepository.readZip, while being a bit memory constrained (e.g. with Xmx128M)
3. Encounter a OutOfMemoryError

**Expected behavior**
I would expect the whole InputStream not to be materialized in-memory by org.kohsuke.github.connector.GitHubConnectorResponse.ByteArrayResponse.bodyStream()

**Additional context**
A typical stack looks like:

```
Java heap space: java.lang.OutOfMemoryError
java.lang.OutOfMemoryError: Java heap space
	at org.apache.commons.io.output.AbstractByteArrayOutputStream.toByteArrayImpl(AbstractByteArrayOutputStream.java:366)
	at org.apache.commons.io.output.ByteArrayOutputStream.toByteArray(ByteArrayOutputStream.java:163)
	at org.apache.commons.io.IOUtils.toByteArray(IOUtils.java:2241)
	at org.kohsuke.github.connector.GitHubConnectorResponse$ByteArrayResponse.bodyStream(GitHubConnectorResponse.java:187)
	at org.kohsuke.github.Requester.lambda$fetchStream$3(Requester.java:117)
	at org.kohsuke.github.Requester$$Lambda$568/0x00000008404da440.apply(Unknown Source)
	at org.kohsuke.github.GitHubClient.createResponse(GitHubClient.java:485)
	at org.kohsuke.github.GitHubClient.sendRequest(GitHubClient.java:387)
	at org.kohsuke.github.GitHubClient.sendRequest(GitHubClient.java:355)
	at org.kohsuke.github.Requester.fetchStream(Requester.java:117)
	at org.kohsuke.github.GHRepository.downloadArchive(GHRepository.java:3246)
	at org.kohsuke.github.GHRepository.readZip(GHRepository.java:3195)
```

bitwiseman:
Agreed, the current implementation has this issue. We chose to do it that way to address other problems in more common scenarios. Allocating some memory that will be quickly returned is okay in most cases. 

[...]

PR #2059, opened 2025-03-13
Add an option to avoid buffered response stream.

atsushieno:
context: #1405

GHArtifact.download() now explicitly sets this option to make response stream non-buffered.

# Description

# Before submitting a PR:

# When creating a PR: 

### DIFF
--- a/src/main/java/org/kohsuke/github/GHArtifact.java
+++ b/src/main/java/org/kohsuke/github/GHArtifact.java
@@ -115,7 +115,11 @@ public class GHArtifact extends GHObject {
     public <T> T download(InputStreamFunction<T> streamFunction) throws IOException {
         requireNonNull(streamFunction, "Stream function must not be null");
 
-        return root().createRequest().method("GET").withUrlPath(getApiRoute(), "zip").fetchStream(streamFunction);
+        return root().createRequest()
+                .method("GET")
+                .withUrlPath(getApiRoute(), "zip")
+                .avoidBufferedResponseStream()
+                .fetchStream(streamFunction);
     }
 
     private String getApiRoute() {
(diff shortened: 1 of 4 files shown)


bitwiseman:
@atsushieno 
Add added pom config to ignore the new `default` method not getting covered.

bitwiseman:
@atsushieno  
Thanks for this PR. 

I made some tweaks to the code to get coverage and handle some edge cases. 

Then moved the unbuffered setting down to `fetchStream()`, so now all stream requesting calls avoid buffering.  

Then added a fallback so buffering is avoided only when status code is  HTTP_OK (200). 

Looking at this, it seems like the only case where we need buffering is for multiple reads of the body stream, which only occur in error, finest logging, and debugging scenarios. 

 I think it could be possible to moved to having unbuffered response streaming be the default for all calls when status == 200 and specifically `preventBufferedResponseBodyStream()` only when calling to `fetchStream()`.  Basically, `response.useBufferedBodyStream()` would be: 

```
public boolean useBufferedBodyStream() {
    return statusCode() != HTTP_OK && !request().preventBufferedResponseBodyStream();
}
```

Maybe with a static variable that allows forcing buffering for debugging.

atsushieno:
@bitwiseman Thanks a lot to make this changeset way more solid. I saw a bunch of changes and the set of further changes seems in the right direction to me.

bitwiseman:
@atsushieno  
PR now switches all successful request calls to use stream instead of byte arrays, automatically switching back to byte array buffering when needed.

atsushieno:
I noticed that, that would be even better if everything works!

bitwiseman (approved):

PR #2074, opened 2025-03-27
Migrate legacy date-time api to new date-time api

### LINKED ISSUE #2073: Replace methods which return `Date` with `Instant` or `ZonedDateTime` for 2.x
There are multiple methods which return `Date` such as `GHObject#getUpdatedAt()`.
the old java `Date` api should really be replaced with one of the following:
- `Instant` (imo this would be the best pick)
- `LocalDateTime`
- `ZonedDateTime`

since a 2.x version is currently being made, now would be the best time to replace it.

it is recommended to avoid the old date-time api and instead use the newer api. many of the methods on `Date` are even marked as deprecated.

solonovamax:
# Description

Migrates the legacy date-time api to newer date-time api introduced in JSR 310.

Fixes #2073

Did not introduce any new tests for this, as no new functionality was introduced, only updated the previous tests to use `Instant` instead of `Date`.

Several methods with `date` in the name were renamed to instead have `instant` in the name. All of these methods are package-private or only accessible to the tests.

# Before submitting a PR:

- [x] ~~Changes must not break binary backwards compatibility. If you are unclear on how to make the change you think is needed while maintaining backward compatibility, [CONTRIBUTING.md](CONTRIBUTING.md) for details.~~ These changes break binary backwards compatibility, however this is meant for new 2.x release.

# When creating a PR: 

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -142,6 +142,10 @@
       <artifactId>jackson-databind</artifactId>
     </dependency>
     <dependency>
+      <groupId>com.fasterxml.jackson.datatype</groupId>
+      <artifactId>jackson-datatype-jsr310</artifactId>
+    </dependency>
+    <dependency>
       <groupId>com.infradna.tool</groupId>
       <artifactId>bridge-method-annotation</artifactId>
       <version>1.30</version>
@@ -636,7 +640,7 @@
       <!--
       This plugin is used to generate AOT metadata during tests so that it can be
       compared against those in META-INF/native-image/org.kohsuke/github-api/*.
-      The tests are start with the name "Aot..." 
+      The tests are start with the name "Aot..."
       -->
       <plugin>
         <groupId>org.springframework.boot</groupId>
--- a/src/main/java/org/kohsuke/github/GHApp.java
+++ b/src/main/java/org/kohsuke/github/GHApp.java
@@ -5,8 +5,8 @@ import org.kohsuke.github.internal.EnumUtils;
 
 import java.io.IOException;
 import java.net.URL;
+import java.time.Instant;
 import java.util.Collections;
-import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
@@ -158,10 +158,10 @@ public class GHApp extends GHObject {
      * @return a list of App installations since a given time.
      * @see <a href="https://developer.github.com/v3/apps/#list-installations">List installations</a>
      */
-    public PagedIterable<GHAppInstallation> listInstallations(final Date since) {
+    public PagedIterable<GHAppInstallation> listInstallations(final Instant since) {
         Requester requester = root().createRequest().withUrlPath("/app/installations");
         if (since != null) {
-            requester.with("since", GitHubClient.printDate(since));
+            requester.with("since", GitHubClient.printInstant(since));
         }
         return requester.toIterable(GHAppInstallation[].class, null);
     }
--- a/src/main/java/org/kohsuke/github/GHAppInstallation.java
+++ b/src/main/java/org/kohsuke/github/GHAppInstallation.java
@@ -6,8 +6,8 @@ import org.kohsuke.github.internal.EnumUtils;
 
 import java.io.IOException;
 import java.net.URL;
+import java.time.Instant;
 import java.util.Collections;
-import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
@@ -189,8 +189,8 @@ public class GHAppInstallation extends GHObject {
      *
      * @return the suspended at
      */
-    public Date getSuspendedAt() {
-        return GitHubClient.parseDate(suspendedAt);
+    public Instant getSuspendedAt() {
+        return GitHubClient.parseInstant(suspendedAt);
     }
 
     /**
--- a/src/main/java/org/kohsuke/github/GHAppInstallationToken.java
+++ b/src/main/java/org/kohsuke/github/GHAppInstallationToken.java
@@ -1,5 +1,6 @@
 package org.kohsuke.github;
 
+import java.time.Instant;
 import java.util.*;
 
 // TODO: Auto-generated Javadoc
@@ -66,7 +67,7 @@ public class GHAppInstallationToken extends GitHubInteractiveObject {
      *
      * @return date when this token expires
      */
-    public Date getExpiresAt() {
-        return GitHubClient.parseDate(expires_at);
+    public Instant getExpiresAt() {
+        return GitHubClient.parseInstant(expires_at);
     }
 }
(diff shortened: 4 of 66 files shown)


solonovamax:
Also a few things I want to mention:
- when trying to create the PR through IntelliJ, it did not let me due to a configuration setting on your end. I was trying to create a PR from my fork to this repo, so it's not an issue with not having access rights to the repository, that is a separate message. The error message was
   > Failed to create a pull request
   > Although you appear to have the correct authorization credentials, the `hub4j` organization has enabled OAuth App access restrictions, meaning that the data access to third-parties is limited. For more information on these restrictions, including how to enable this app, visit https://docs.github.com/articles/resstricting-access-to-your-organization-s-data/
- I noticed a few inconsistencies in the source code:
  - some methods have underscores in their names (this is something that would be best to fix right now before the 2.x release)
    - `GHRelease.getPublished_at()`
    - `GHDeployKey.getAdded_by()`
    - `GHDeployKey.isRead_only()`
    - `GHDiscussion.private_()`
    - `GHGistBuilder.public_()`
    - `GHPullRequestCommitDetail.getComment_count()`
    - `GHPullRequestCommitDetail.getHtml_url()`
    - `GHRepositoryBuilder.private_()`
  - some classes have a license header at the top, others do not
  - some classes use glob imports, others do not

bitwiseman:
- The intelliJ problem I'm open to 
- The license problem has an open issue
- The underscore problem - please file an issue to add an ArchUnit test that enforces no underscores (with the ability to add exceptions for cases we choose to allow) 
- I also need to set naming guidance for future additions to the library.

bitwiseman:
As for your code changes, the build failed because you introduced binary breaking changes.  That is not a surprise.  

We could use depreciation annotations or bridge methods to address this. I know we're moving to 2.0, but I'd still like to minimize breaking charges. 

If we use depreciation, I'd say we name the new  `Instsnt` methods without the `get` on front. Have  the `Date` methods call the `Instsnt` methods, and then we can leave the tests unchanged and still get full coverage. 


We previously handled these kind of API breaking changes using bridge methods. One advantage of using bridge methods over depreciation annotations is it forces people to switch to using the new API the next time they compile their library (not at some future point).  The disadvantage is they make mocking tests fail in odd ways. Still, we just removed them, I was kind of hoping to avoid bringing them back.  

Sigh.

solonovamax:
> As for your code changes, thanks for the fast turnaround. The build failed because you introduced binary breaking changes. That is not a surprise. Other than that, it looks good.

yeah, I gathered that much.
I was going to comment about it, but forgot to

all the tests I was able to run locally all passed (a few were skipped, most likely bc they're windows or macos specific tests & I'm on linux)

> We previously handled these kind of API breaking changes using bridge methods. One advantage of using bridge methods over depreciation annotations is it forces people to switch to using the new API the next time they compile their library (not at some future point). The disadvantage is they make mocking tests fail in odd ways. Still, we just removed them, I was kind of hoping to avoid bringing them back.

wdym by "bridge methods"?

> The intelliJ problem I'm open to changing settings if it is needed. File a separate issue.

sure, I can do that.

> The license problem has an open issue. Not sure how to enforce it yet, but we should.

looking at spotless, it seems that it has a check for the license header. so, if you wanted to enforce it you could use that.
see: https://github.com/diffplug/spotless/tree/main/plugin-maven#license-header

> The underscore problem - please file an issue to add an ArchUnit test that enforces no underscores (with the ability to add exceptions for cases we choose to allow)

sure, I can look into that in a bit.
 
> We use spotless plugin to enforce formatting and depend on it to keep imports clean. There's probably a setting we could add to enforce usage either for or against globs.  Again please file an issue.

sure, will do.
it seems that there is currently an open issue on removing wildcard imports: https://github.com/diffplug/spotless/issues/649
it does propose a solution of
```xml
<replaceRegex>
    <name>Remove wildcard imports</name>
    <searchRegex>import\s+(?:static\s+)?[^\*\s]+\*;(\r\n|\r|\n)</searchRegex>
    <replacement>$1</replacement>
</replaceRegex>
```
though, simply running this will cause the compile to fail because it doesn't replace the glob import with imports for the used classes/methods.

solonovamax:
also something else I just noticed, is that it seems in some places all local variables (which are not later modified) are declared as `final`, whereas in other places they are not.

bitwiseman:
> also something else I just noticed, is that it seems in some places all local variables (which are not later modified) are declared as `final`, whereas in other places they are not.

Yeah, there isn't really one right answer and it can be confusing for new or occasional contributors. 

See this example for pros/cons of different options:
https://github.com/hub4j/github-api/blob/main/src%2Fmain%2Fjava%2Forg%2Fkohsuke%2Fgithub%2Fexample%2Fdataobject%2FReadOnlyObjects.java

bitwiseman:
> wdym by "bridge methods"?

Bridge methods are binary method overloads for methods with the same arguments (or no arguments) and different return types. 

Look at https://github.com/jenkinsci/bridge-method-injector

You can see examples of how we used them by looking at the `main-1.x` branch.

solonovamax:
> > wdym by "bridge methods"?
> 
> Bridge methods are binary method overloads for methods with the same arguments (or no arguments) and different return types.
> 
> Look at https://github.com/jenkinsci/bridge-method-injector
> 
> You can see examples of how we used them by looking at the `main-1.x` branch.

honestly, I feel like the breaking changes are rather small. if anyone is still heavily relying on the old date api, then they can either just wrap it in `Date.from(...)` or they just don't upgrade to the newer api.

I guess the one complication of this would be jenkins plugins, as I believe that this api is used in several different locations, and many jenkins plugins are not exactly being maintained (eg. I have https://github.com/jenkinsci/discord-notifier-plugin/pull/138 and https://github.com/jenkinsci/github-branch-source-plugin/pull/790 which have just been sitting there for nearly a year lol, one of which has even *forced* me to use my own build of the plugin otherwise rendering jenkins entirely unusable)

solonovamax:
> Maybe the right thing to do here is to have the primary artifact be unbridged (with no added descriptor) and have a secondary "-bridged" artifact?

yeah, this might be the best option

I feel like the majority of consumers of the library will have little to no use for the bridged methods and will just be able to migrate without significant issue.
then again, for those who don't have any use for the bridged methods, having them there causes no harm, correct?

bitwiseman:
Actually, we created the `unbridged` because there were issues with consumers of this library mocking classes that had bridge methods. :sob:

solonovamax (commented):
Noticed you forgot to mark a few things as `@Deprecated`

unsure if this is intentional or not, so I'm not going to push a commit that changes it, but if it wasn't then that's why I'm commenting on it

bitwiseman:
@solonovamax 
Do you have the cycles to add `@Deprecated` including the Javadoc everywhere it is needed?

bitwiseman:
@solonovamax  
I'm done making changes.

PR #2101, opened 2025-05-08
Add workflow run query builder for specific workflow

HerrDerb:
# Description

Currently the `GHWorkflowRunQueryBuilder` is only available in `GHRepository` which does not allow to use the query builder for a specific workflow.

By adding it to `GHWorkflow` and extend the `GHWorkflowRunQueryBuilder` to also be usable for a specific workflow id one can query a specific workflow, being more efficient


HerrDerb:
@bitwiseman I need to fix/add the tests in a quite minute :)

HerrDerb:
I am going to close this as not needed anymore

Issue #2009, opened 2025-01-15
`GitHubAbuseLimitHandler#isError` fails to detect some secondary rate limit errors


yrodiere:
**Describe the bug**

#1895 attempted to detect more errors related to secondary rate limits, but it seems there are more.

It appears GitHub APIs -- https://api.github.com/search/issues in particular, maybe more -- can hit a secondary limit and just return a 403 error, with no particular header indicating that a limit was reached. No `Retry-After`, no `gh-limited-by`, nothing. Only the body response explains "You have exceeded a secondary rate limit. Please wait a few minutes before you try again". (and in fact waiting 30s/1min is enough).

This, understandably, doesn't get caught by `GitHubAbuseLimitHandler#isError`, and results in the request simply failing.

**To Reproduce**

Send the following request a few dozen times (or just a dozen when unauthenticated):

https://api.github.com/search/issues?sort=updated&order=desc&q=repo%3Ayrodiere%2Fquarkus-github-playground+is%3Aissue+is%3Aopen+label%3Atriage%2Fneeds-feedback%2Ctriage%2Fneeds-reproducer+label%3Aarea%2Fhibernate-search+updated%3A%3C2025-01-15T12%3A03%3A47.243460581

Obviously my actual use case is not to send the _same_ request over and over, but it's enough to reproduce the problem.

Interestingly, you can just click that URL in your browser, refresh quickly a couple times, and you'll be able to observe the exact error I'm running into: HTTP 403 with no particular retry header.

See bottom of this messsage for more logs.

**Expected behavior**

Ideally, I'd expect such requests to be detected as secondary rate limit errors, and the client to just wait and retry. For my use case it's fine to wait a minute.

Failing that, I'd settle for a way to override `GitHubAbuseLimitHandler#isError` and implement some dirty logic based on the response body in there. But this method is currently package-protected.

**Desktop (please complete the following information):**
 - OS: Fedora 41
 - Browser [e.g. chrome, safari]: no browser involved
 - Version [e.g. 22]: no browser involved

**Additional context**

Here is a failing request/response with all details/headers. I got it using `-Djdk.httpclient.HttpClient.log=all`.

```
[...]

PR #2127, opened 2025-08-21
Make `isError` on GitHubConnectorResponseErrorHandler public.

rozza-sb:
# Description

Changes the access modifier on GithubConnectorResponseErrorHandler and subclasses to public. This allows library users to overwrite what constitutes an error for the response, particularly for (secondary) rate limiting.

Relates to #2009, where GithubAbuseLimitHandler is not properly detecting secondary rate limits but users cannot manually override it while the method is package-private. Additionally, the definition of `isError` under `STATUS_HTTP_BAD_REQUEST_OR_GREATER` is public, which could imply that this should be public from the start. A future PR will do a fix to the underlying issue, but in the mean time this would allow users to implement their own fix.

# Before submitting a PR:

# When creating a PR: 

### DIFF
--- a/src/main/java/org/kohsuke/github/GitHubAbuseLimitHandler.java
+++ b/src/main/java/org/kohsuke/github/GitHubAbuseLimitHandler.java
@@ -107,7 +107,7 @@ public abstract class GitHubAbuseLimitHandler extends GitHubConnectorResponseErr
      *             Signals that an I/O exception has occurred.
      */
     @Override
-    public boolean isError(@Nonnull GitHubConnectorResponse connectorResponse) {
+    public boolean isError(@Nonnull GitHubConnectorResponse connectorResponse) throws IOException {
         return isTooManyRequests(connectorResponse)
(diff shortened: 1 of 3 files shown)


rozza-sb:
Regarding said better implementation, as mentioned in my comment under #2009 there are two potential messages that can be returned that we can check for:
- "You have exceeded a secondary rate limit"
- "You have triggered an abuse detection mechanism"

On a local branch that I had started working on, I have a method that we could extend `isError` with:
```java
private boolean hasRelevantErrorMessages(GitHubConnectorResponse connectorResponse) throws IOException {
        String content = new String(connectorResponse.bodyStream().readAllBytes(), StandardCharsets.UTF_8).toLowerCase();
        return content.contains("you have exceeded a secondary rate limit") || content.contains("you have triggered an abuse detection mechanism");
    }
```

If the general concept of this seems good as the fix, I'd go on ahead with this and open a PR once I get some time 👍

bitwiseman:
@rozza-sb 
I'm going to close this PR.  Please open a new one for your other proposed change.

PR #2186, opened 2026-01-25
feat: limit searching issues based on type

### LINKED ISSUE #2032: Add more methods to QueryBuilder
1)
`GHPullRequestQueryBuilder` doesn't seem to support `pageSize(int)` but `GHIssueQueryBuilder` does.

I don't know if there is a technical reason, but it would help to support repos with a large number of PRs. I didn't realize this was a possibility at first and took a long time to start iterating through issues.

2)
`GHIssueQueryBuilder` seems to still pull in PRs. Is it possible to add another method to the query to drop PRs from Issue queries and the same for Issues from PR queries. It would help the amount of data to transfer from the server if the server supported it.

Anonycoders:
# Description

* Fixes #2032 - ( Is it possible to add another method to the query to drop PRs from Issue queries ... )

## Changes

* `isPullRequest` => filter issues down to `pull_requests` only.
* `isIssue` => filter issues down to `issue` only.
* `repo` => allow limit the search to a specific repository of given organization.

https://docs.github.com/en/search-github/searching-on-github/searching-issues-and-pull-requests#search-only-issues-or-pull-requests

# Before submitting a PR:

# When creating a PR: 


src/test/java/org/kohsuke/github/AppTest.java
```diff
@@ -855,6 +855,44 @@ public void testIssueSearch() {
...
+     * Test issue search with isPullRequest filter to only return pull requests.
+     */
+    @Test
+    public void testIssueSearchPullRequestsOnly() {
+        PagedSearchIterable<GHIssue> r = gitHub.searchIssues()
+                .repo("hub4j", "github-api")
+                .isPullRequest()
+                .isOpen()
+                .sort(GHIssueSearchBuilder.Sort.UPDATED)
+                .list();
```
bitwiseman: Anyway to shorten the return list?  If we can test this without storing several pages of json that we don't really care about. 

Maybe `isClosed()` and `sort.CREATED`, then read the first few records?

PR #2219, opened 2026-04-08
Add missing setter methods to GHTeam

mvanhorn:
GHTeam currently only has `setDescription` and `setPrivacy`. The [GitHub API](https://docs.github.com/en/rest/teams/teams#update-a-team) supports updating several more properties.

This adds 4 missing setters, all following the same PATCH pattern as the existing ones:

- `setName(String name)`
- `setNotificationSetting(String notificationSetting)`
- `setPermission(String permission)`
- `setParentTeamId(Long parentTeamId)` (nullable, to support removing parent)

Each is a single PATCH call to the team endpoint. An atomic multi-property update builder could follow as a separate enhancement if maintainers want one.

Fixes #2218

This contribution was developed with AI assistance (Codex).

### DIFF
--- a/src/main/java/org/kohsuke/github/GHTeam.java
+++ b/src/main/java/org/kohsuke/github/GHTeam.java
@@ -511,6 +511,58 @@ public class GHTeam extends GHObject implements Refreshable {
         root().createRequest().method("PATCH").with("privacy", privacy).withUrlPath(api("")).send();
     }
 
+    /**
+     * Sets the team's name.
+     *
+     * @param name
+     *            the new name
+     * @throws IOException
+     *             the io exception
+     */
+    public void setName(String name) throws IOException {
+        root().createRequest().method("PATCH").with("name", name).withUrlPath(api("")).send();
+    }
+
+    /**
+     * Sets the team's notification setting.
+     *
+     * @param notificationSetting
+     *            the notification setting (e.g. "notifications_enabled" or "notifications_disabled")
+     * @throws IOException
+     *             the io exception
+     */
+    public void setNotificationSetting(String notificationSetting) throws IOException {
+        root().createRequest()
+                .method("PATCH")
+                .with("notification_setting", notificationSetting)
+                .withUrlPath(api(""))
+                .send();
+    }
+
+    /**
+     * Sets the team's permission.
+     *
+     * @param permission
+     *            the permission (e.g. "pull", "push", or "admin")
+     * @throws IOException
+     *             the io exception
+     */
+    public void setPermission(String permission) throws IOException {
+        root().createRequest().method("PATCH").with("permission", permission).withUrlPath(api("")).send();
+    }
+
+    /**
+     * Sets the team's parent team by ID.
+     *
+     * @param parentTeamId
+     *            the ID of the parent team, or {@code null} to remove the parent
+     * @throws IOException
+     *             the io exception
+     */
+    public void setParentTeamId(Long parentTeamId) throws IOException {
+        root().createRequest().method("PATCH").with("parent_team_id", parentTeamId).withUrlPath(api("")).send();
+    }
+
     private String api(String tail) {
         if (organization == null) {
             // Teams returned from pull requests to do not have an organization. Attempt to use url.
--- a/src/main/java/org/kohsuke/github/GHTeam.java
+++ b/src/main/java/org/kohsuke/github/GHTeam.java
@@ -35,6 +35,28 @@ public class GHTeam extends GHObject implements Refreshable {
     }
 
     /**
+     * Notification setting for a team. Controls whether members receive notifications when the team is mentioned.
+     */
+    public enum NotificationSetting {
+
+        /** Notifications enabled: members receive notifications when the team is @mentioned. */
+        NOTIFICATIONS_ENABLED("notifications_enabled"),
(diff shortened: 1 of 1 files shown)


src/main/java/org/kohsuke/github/GHTeam.java
```diff
@@ -511,6 +511,58 @@ public void setPrivacy(Privacy privacy) throws IOException {
...
+
+    /**
+     * Sets the team's notification setting.
+     *
+     * @param notificationSetting
+     *            the notification setting (e.g. "notifications_enabled" or "notifications_disabled")
+     * @throws IOException
+     *             the io exception
+     */
+    public void setNotificationSetting(String notificationSetting) throws IOException {
```
rozza-sb: Looking at `GHTeamBuilder`, this doesn't already exist in this repo as a concept under a team. However, I think this would work better as an enum so that we can only provide valid values to the request

src/main/java/org/kohsuke/github/GHTeam.java
```diff
@@ -511,6 +511,58 @@ public void setPrivacy(Privacy privacy) throws IOException {
...
+
+    /**
+     * Sets the team's permission.
+     *
+     * @param permission
+     *            the permission (e.g. "pull", "push", or "admin")
+     * @throws IOException
+     *             the io exception
+     */
+    public void setPermission(String permission) throws IOException {
```
rozza-sb: `GHOrganization.Permission` exists for this already, so should be reused here

rozza-sb (commented):
Some notes on notifications and permissions

mvanhorn:
Thanks for the feedback @rozza-sb - addressed in 8dadafe:

- `setPermission` now takes `GHOrganization.Permission` so callers can only pass valid values (ADMIN / MAINTAIN / PULL / PUSH / TRIAGE). The enum names lowercase directly into the API values the Teams endpoint expects.
- `setNotificationSetting` now takes a new `GHTeam.NotificationSetting` enum (`NOTIFICATIONS_ENABLED` / `NOTIFICATIONS_DISABLED`). The enum overrides `toString()` to return the exact API strings (with underscores), and the setter passes the string to `.with()` so `transformEnum()` doesn't rewrite the underscores as hyphens.

No existing callers of either setter outside `GHTeam.java`, so no downstream breakage from the signature changes.

src/main/java/org/kohsuke/github/GHTeam.java
bitwiseman: Feel free to add enum methods and have the string methods call those to validate the string input.  Still need the string methods though.

src/main/java/org/kohsuke/github/GHTeam.java
```diff
@@ -34,6 +34,28 @@ public enum Privacy {
         UNKNOWN
     }
 
+    /**
+     * Notification setting for a team. Controls whether members receive notifications when the team is mentioned.
+     */
+    public enum NotificationSetting {
+
+        /** Notifications enabled: members receive notifications when the team is @mentioned. */
+        NOTIFICATIONS_ENABLED("notifications_enabled"),
```
bitwiseman, line 43: Enums must have an UNKNOWN. See other enums such as PRIVACY and their usage. 

You probably don't need the string either as Jackson will automatically convert for you.

src/main/java/org/kohsuke/github/GHTeam.java
```diff
@@ -511,6 +533,58 @@ public void setPrivacy(Privacy privacy) throws IOException {
...
 
+    /**
+     * Sets the team's name.
+     *
+     * @param name
+     *            the new name
+     * @throws IOException
+     *             the io exception
+     */
+    public void setName(String name) throws IOException {
```
bitwiseman, line 544: We're moving to not have set methods on objects. I don't see any on here yet...

bitwiseman (changes requested):
> Each is a single PATCH call to the team endpoint. An atomic multi-property update builder could follow as a separate enhancement if maintainers want one.

Let's just go straight to that multi-property update builder. So, GHLabel for an example of the behavior 

Also, of course testing needed .

rozza-sb:
I have a local branch for the multi property updater - I'm finishing the wiremocks for this on Monday but I can push what I have first thing

Would these single property updaters not be separately useful in terms of completing the properties that can do this, given there are some already under GHTeam?

mvanhorn:
Good to know about the multi-property updater - makes sense as a companion. The single-property setters are still useful on their own since they match the existing pattern in GHTeam (e.g. setDescription, setPrivacy). Happy to coordinate if there's overlap.

rozza-sb:
Based on [this](https://github.com/hub4j/github-api/pull/2219#discussion_r3068476002) comment, if the general trend is to move away from single method updates then I think just the multi-update builder by itself would be acceptable

I have a separate comment re: current GHTeamBuilder and GHLabel behaviour but I'll ask in a separate PR

mvanhorn:
Thanks for the clarity @bitwiseman @rozza-sb. Makes sense to move toward the builder pattern instead of individual setters. Happy to close this in favor of @rozza-sb's multi-property updater - the enum additions (NotificationSetting, UNKNOWN on Privacy) and wiremock recordings from this branch are available if any of it is useful for the builder PR.

PR #2249, opened 2026-04-13
Add builder for updating multiple GHTeam properties.

rozza-sb:
# Description

Currently GHTeam has some single methods on them to update description and privacy properties. There should be a mechanism for not only all properties being updatable (as supported per the [GitHub API](https://docs.github.com/en/rest/teams/teams?apiVersion=2026-03-10#update-a-team)), but also for multiple properties to be updated atomically.

This change adds support for a new builder that allows for properties to be updated.

Fixes #2218.

# Before submitting a PR:

- [ ] Run `mvn -D enable-ci clean install site "-Dsurefire.argLine=--add-opens java.base/java.net=ALL-UNNAMED"` locally. If this command doesn't succeed, your change will not pass CI.

# When creating a PR: 

### DIFF
--- a/src/main/java/org/kohsuke/github/GHTeam.java
+++ b/src/main/java/org/kohsuke/github/GHTeam.java
@@ -45,7 +45,9 @@ public class GHTeam extends GHObject implements Refreshable {
          */
         MAINTAINER,
         /** A normal member of the team. */
-        MEMBER
+        MEMBER,
+        /** Unknown role. */
+        UNKNOWN
     }
 
     /**
@@ -59,7 +61,11 @@ public class GHTeam extends GHObject implements Refreshable {
         /**
          * No one receives notifications.
          */
-        NOTIFICATIONS_DISABLED
+        NOTIFICATIONS_DISABLED,
+        /**
+         * Unknown notification setting.
+         */
+        UNKNOWN
     }
 
     /**
--- a/src/main/java/org/kohsuke/github/GHTeam.java
+++ b/src/main/java/org/kohsuke/github/GHTeam.java
@@ -49,6 +49,20 @@ public class GHTeam extends GHObject implements Refreshable {
     }
 
     /**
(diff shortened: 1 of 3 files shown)


rozza-sb:
@bitwiseman re: using GHLabel as a base, I agree that this approach would be the more maintainable one going forward. However, we already have `GHTeamBuilder` using an old approach. Happy to include the new `AbstractBuilder` as a separate class, but what should it be named in the mean time? Long term there is probably a migration to be had for that type but it will break compatibility as is

As it is this branch has a separate builder for updates, with only wiremocks missing. I'll be working on a commit that does the GHLabelBuilder-like pattern instead.

rozza-sb:
I've taken a look at the GHLabel pattern, and I think this works quite well as both the [create](https://docs.github.com/en/rest/issues/labels?apiVersion=2026-03-10#create-a-label) and [update](https://docs.github.com/en/rest/issues/labels?apiVersion=2026-03-10#update-a-label) endpoints support the same properties. For GHTeam, there are differences between the [create](https://docs.github.com/en/rest/teams/teams?apiVersion=2026-03-10#create-a-team) and the [update](https://docs.github.com/en/rest/teams/teams?apiVersion=2026-03-10#update-a-team) endpoints:
1. Create can take `maintainers` and `repo_names` whereas update can not
2. Update can take `permissions` whereas create can not
  a. Admittedly this field is marked as deprecated so we could not support this)
  b. If we do support `permissions`, different values are supported depending on whether we are creating or updating (`admin` is supported during update only)

Based on the difference between the create and update it does feel like these should be separate. My current plan is to get the existing change with wiremocks setup, partially due to the separate behaviour, and partially due to not having time before I head abroad to get this new setup in a satisfied state. If we're really against the existing approach, even as a stopgap, we could look at some migration strategy to support the new pattern with differing Creator and Updater requirements without breaking the existing `GHTeamBuilder`

Issue #2035, opened 2025-02-17
GHCommitPointer throws NPE on null repository


rnveach:
**Describe the bug**

I receive an NPE on `GHCommitPointer#getCommit` because `getRepository()` returns `null`.

I assume `getRepository()` should always be populated as I need a `GHRepository` to even get the initial PR/issue in the first place.

**To Reproduce**

While scanning https://github.com/checkstyle/checkstyle/pulls , I pick up PR 1.
I call `GHPullRequest#getHead()` and get the `GHCommitPointer`.
I call `GHCommitPointer#getCommit()` and receive an NPE.
`getRepository()` returns `null`.

**Expected behavior**

No NPE.

**Desktop (please complete the following information):**

**Additional context**

rnveach:
@bitwiseman Can you explain how `repo` is normally populated? I assume there is reflection going on here as there is no setter or assignment.

https://github.com/hub4j/github-api/blob/e14ec3b3677760714cd096ad8157a3e6a6dded65/src/main/java/org/kohsuke/github/GHCommitPointer.java#L46

bitwiseman:
@rnveach  
https://github.com/FasterXML/jackson populating from JSON to objects.

n09tal:
Hi, I submitted a fix proposal for this issue here:
https://github.com/hub4j/github-api/pull/2265

Summary:
- Added null check in GHCommitPointer#getCommit
- Added regression test GHCommitPointerTest#nullRepoNpe
- Verified targeted test passes

PR #2265, opened 2026-05-17
Fix GHCommitPointer null repository path and add regression test

n09tal:
Fix issue #2035: handle null repository in GHCommitPointer#getCommit and add regression test.

### What changed

1) `src/main/java/org/kohsuke/github/GHCommitPointer.java`
- Updated `getCommit()` to avoid an uncontrolled NullPointerException path.
- Added a local `repository` variable from `getRepository()`.
- Added a null guard:
  - If repository is null, throw `IOException("Cannot commit because repository is null.")`.
  - Otherwise continue with `repository.getCommit(getSha())`.

2) `src/test/java/org/kohsuke/github/GHCommitPointerTest.java` (new file)
- Added regression test `nullRepoNpe`.
- Test creates a `GHCommitPointer` with default null repository state.
- Verifies `getCommit()` now throws `IOException` (not NPE).
- Verifies exception message contains `"repository is null"`.

3) `src/main/java/org/kohsuke/github/GHPullRequest.java`
- Minor inline comment update in `getHead()` for debugging context.
- No functional logic change.

### Why
Issue #2035 reports an NPE when `GHCommitPointer#getCommit()` is called and repository data is missing.  
This change makes behavior explicit and controlled, and adds test coverage for the null-repository path.

### Validation
- Targeted test command:
  - `mvn -Dtest=GHCommitPointerTest#nullRepoNpe test`
- Result:
  - `Tests run: 1, Failures: 0, Errors: 0`
  - `BUILD SUCCESS`

### DIFF
--- a/src/main/java/org/kohsuke/github/GHCommitPointer.java
+++ b/src/main/java/org/kohsuke/github/GHCommitPointer.java
@@ -52,9 +52,21 @@ public class GHCommitPointer {
      * @throws IOException
      *             the io exception
      */
+
+    //original code
+    // public GHCommit getCommit() throws IOException {
+    //     return getRepository().getCommit(getSha()); //when repo is null, so this will throw NPE
+    // }
+
+    //fixed code
     public GHCommit getCommit() throws IOException {
-        return getRepository().getCommit(getSha());
+    GHRepository repository = getRepository();
+    if (repository == null) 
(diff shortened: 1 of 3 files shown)


rnveach:
From my perspective, this doesn't look like a fix. It just replaces NPE with IOE.

PR #2269, opened 2026-05-28
Add verified_at to GHVerification

finlaysawyer:
# Description

Adds support for `verified_at`, which is included in the `verification` response for verified commits.

https://docs.github.com/en/rest/commits/commits?apiVersion=2026-03-10#get-a-commit


andjelalabovic (approved):

src/main/java/org/kohsuke/github/GHVerification.java
```diff
@@ -119,6 +124,16 @@ public String getSignature() {
         return signature;
     }
 
+    /**
+     * Gets the time when the signature was verified by GitHub.
+     *
+     * @return the instant the signature was verified, or null if not available.
+     */
+    @WithBridgeMethods(value = Date.class, adapterMethod = "instantToDate")
```
bitwiseman: This doesn't need bridging. It is a new getter..

finlaysawyer: Updated

bitwiseman (approved):

